PR #9, opened 2023-03-08
Package skeleton with minimal configuration

YooSunYoung:


setup.cfg
```diff
@@ -0,0 +1,36 @@
...
+include_package_data = True
+
+[options.packages.find]
+where = src
+
+[flake8]
+# See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length
+max-line-length = 88
+extend-ignore = E203
+exclude = .*,__pycache__,dist
```
jl-wynen, line 36: Please change to the more modern version and move everything you can into `pyproject.toml`. See ScippNeutron.

jl-wynen (changes requested):

setup.cfg
jl-wynen: Though ScippNeutron might also be out of date. Setuptools is in the process of migrating everything to pyproject.toml.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+# Excluding tests because bandit doesn't like `assert`.
+exclude_dirs = ["tests"]
+
+[tool.codespell]
+skip = "./.git,./.tox,*/*_cache"
+
+[tool.isort]
+skip_gitignore = true
+line_length = 88
+multi_line_output = 2  # for compatibility with yapf
```
SimonHeybrock, line 46: Is this needed with `black`?

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+requires = [
+  "setuptools>=42",
+  "setuptools_scm[toml]>=6.2",
+]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
+
+[project]
+name = "peek_data"
```
SimonHeybrock, line 11: Is this the name we intend to keep?

YooSunYoung: I am not sure... I don't have a better name at the moment though.

pyproject.toml
jl-wynen: No, it should simply be
```toml
[tool.isort]
skip_gitignore = true
profile = "black"
```

YooSunYoung: `isort` sorts the imported libraries in the alphabetical order as well.
`black` and `isort` have some conflicts and it's described here.
https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html

Actually, I don't like that it re-orders the imported calls or classes in alphabetical order.
Can I just drop the `isort`??

SimonHeybrock: We use it in all other projects, I think alphabetical has some utility, so I think we should keep it.

pyproject.toml
nvaytet: `liveview` and `seelive` are available on `pip`.
I had also named an old project `visens` (for VISualization for Ess Neutron Science) which we can reuse (`visens` means 'seeing' in latin).

YooSunYoung: (`visens` also means `aware` and `a queen bee` in Swedish, and `songs` in Danish. Interesting.)
This dashboard will be one type of visualization but I think they should be separated.

I opened a new issue for naming: #12

YooSunYoung:
I will merge this for now and open a new pull request with newer updates.

PR #28, opened 2023-03-23
Fake data generator

### LINKED ISSUE #23: Create dummy test data
Please add ideas of visually satisfying and physically meaningful data that we can use for demo and testing.

I am thinking of simulated 3d scan of lime...?

YooSunYoung:
Fixes #23


src/beamlime/resources/images/generators.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def add_noise(frame: np.ndarray, noise_range=0.2) -> np.ndarray:
+    """
+    Add uniform random noise within the range [0, noise_range) on the image.
+    """
+    return frame + np.random.uniform(0, abs(noise_range), frame.shape)
+
+
+def fake_2d_detector_img_generator(
```
SimonHeybrock, line 37: Any reason for not returning `sc.Variable`?

YooSunYoung: Because it is only used to generate `raw-data`.
The fake data feeder that I added in #30 will wrap this into a dictionary.

So the `offline.data_feeder` will generate what `kafka-listener` will eventually return for testing.

SimonHeybrock (approved):

PR #30, opened 2023-03-23
Fake data feeder

YooSunYoung:
It also contains the changes in pr #28 

For now, the `feeding` is done by putting an item in the `Queue`.
The ``Fake2dDetectorImageFeeder`` instance can create a coroutines.
It will be better if I can create an abstract class that you can use as a _little_ `application`


YooSunYoung:
For now, `proto` script also contains a dummy `data-reduction` application.

Just to see what is happening, I added printing lines in the calls.
They should be replaced with `logging`.

Screenshot from 2023-03-23 14-51-22

src/beamlime/offline/fake_2d_image.ipynb
```diff
@@ -0,0 +1,272 @@
+{
```
YooSunYoung, line 1: I... still didn't set up the document pages so here is the exported pdf 

[fake_2d_image.pdf](https://github.com/scipp/beamlime/files/11052385/fake_2d_image.pdf)

YooSunYoung:
I will close this since all changes are also in the pr #32

PR #32, opened 2023-03-28
Feature data reduction handler

### LINKED ISSUE #3: Dummy data offline visualization module

YooSunYoung:
Fixes #30 
Fixes #3 

It contains the changes in pr #28 and #30 

It has a prototype jupyter notebook that creates instances of `data-generator/feeder`, `data-reduction` and `visualization`.


src/beamlime/core/application.py
```diff
@@ -0,0 +1,79 @@
...
+
+from colorama import Style
+from colorama.ansi import AnsiBack, AnsiFore, AnsiStyle
+
+
+class BeamlimeApplicationInterface(ABC):
+    _input_ch = None
+    _output_ch = None
+    verbose = None
+    verbose_option = None
```
jl-wynen, line 17: Why are those class methods?

src/beamlime/offline/data_feeder.py
```diff
@@ -0,0 +1,52 @@
...
+from PIL.ImageOps import flip
+
+from ..core.application import BeamlimeApplicationInterface
+from ..resources.images import load_icon_img
+from ..resources.images.generators import fake_2d_detector_img_generator
+
+
+class Fake2dDetectorImageFeeder(BeamlimeApplicationInterface):
+    detector_size = None
+    num_frame = None
```
jl-wynen, line 15: Again, why class attributes and not instance attributes?

src/beamlime/core/application.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(
+        self,
+        config: dict,
+        verbose: bool = False,
+        verbose_option: Union[AnsiFore, AnsiStyle, AnsiBack, str] = Style.RESET_ALL,
+    ) -> None:
+        self.parse_config(config)
+        self.verbose = verbose
+        self.verbose_option = verbose_option
```
jl-wynen, line 27: Why do application interfaces need to know about output styles?

src/beamlime/offline/data_feeder.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(
+        self, config: dict, verbose: bool = False, verbose_option: str = Fore.BLUE
+    ) -> None:
+        super().__init__(config, verbose, verbose_option)
+
+    def parse_config(self, config: dict) -> None:
+        if "detector-size" not in config:
+            config["detector-size"] = (64, 64)
```
jl-wynen, line 24: Seems odd that this modifies the external config dict. Is the value written here used outside of this class?

src/beamlime/offline/data_reduction.py
jl-wynen, line 1: I'm struggling to understand the various maps in this file. Can you give a brief explanation what they all do and how they store data? (history, working_target_map, etc.)

src/beamlime/offline/proto.ipynb
jl-wynen, line 1: @SimonHeybrock We had problems with asyncio in notebooks. I don't remember the details, do you think the usage here is ok?

src/beamlime/offline/data_feeder.py
YooSunYoung: Yes, I wanted to save the configuration along with the `log`s.
But indeed it might cause unexpected effect on other applications if each one of `workflow`s becomes an application...
I'll update this part.

src/beamlime/offline/data_feeder.py
YooSunYoung: Initially I did it since the `_run` is `staticmethod` and it needed to check the validity of these fields
but now I removed that logic so I can make them as instance attributes...! Thanks

src/beamlime/offline/data_reduction.py
YooSunYoung: A bit of context:
`workflow` configuration describes **which** `function` you use to process on **which** `target`(data) and **how** the previous input/result is applied to your new input/result.
`target` configuration should describe which information you need to pick up from the raw data and how you wrap it before you feed then into the workflows.


###  `workflow_target_map`
is link between `workflow`s and `targets`.
You can access to the `target` configuration through this map by the `name` of the `workflow`.
It was needed because `target` list is independent from `workflow` configurations,
so that multiple `workflow`s may use the same `target`.

### `workflow_map`
is just a dictionary that has the workflow configuration.
I had to make this first at the beginning and I wrote the reason down here: https://github.com/scipp/beamlime/discussions/33 

### `history`
is for saving previous `input` from the data flow or `result` out of the workflow.
So the `process` function is quite messy to deal with different types of `input` or `result` handling...

src/beamlime/core/application.py
YooSunYoung: It was a temporary solution until I implement `logging`...! I will remove this as soon as we have proper loggers.

src/beamlime/offline/proto.ipynb
YooSunYoung: So far yes...?
But it sometimes doesn't show the error messages or just stuck somewhere then I have to restart the kernel...

jl-wynen: > just stuck somewhere then I have to restart the kernel...

This sounds familiar. I think the problems had to do with the event loop not terminating. Our decision at the time as to not use async in notebooks. Might be that we need to revisit this decision if you require notebooks.

src/beamlime/offline/data_feeder.py
YooSunYoung: I opened the issue #34 so it'll be resolved there.

src/beamlime/core/application.py
YooSunYoung: I opened the issue #34 so it'll be resolved there.

PR #36, opened 2023-04-04
Logging feature

### LINKED ISSUE #34: Logging and log parser/monitoring
Currently applications just print to std I/O but they should be logged in the same place and collected.
Also, it should be easy to parse the log and for analysis.

We might also want performance monitoring, and logs should also have the related information. -> I am not quite sure if it should be separate thing or we can just use logs for this... it'll depend how and why we want to monitor the performance...?

Then we should also remove prompt coloring option from application constructors, mentioned here: #32

YooSunYoung:
Fixes #34 

# New features

## Handler
These two handlers are not related to each other.

1. _HeaderMixin: emits header if needed, whenever a formatter is changed.
2. BeamlimeFileHandler: appends logs into the log file.
3. BeamlimeStramHandler: prints logs into the stdout.

## Formatter
It only has "{" style formatting.
``LogColumn`` and ``LogHeader`` are used to describe formatting.

1. _HeaderFormatter
    - creates formatter(string) using header or headers using formatter(string).
    - base class for other formatter classes below.
2. BeamlimeFormatter
    - header formatter using "{asctime:23} | {app_name:15} | {levelname:8} | {message}"
3. BeamlimeAnsiColorFormatter
    - header formatter using "{ansi_color}{asctime:23} | {app_name:15} | {levelname:8} | {message}{reset_color}"

## Record
Log record classes that have ``app_name``, ``ansi_color`` and ``reset_color`` for formatting.

I wanted to keep all logs in one file, so I tried to add ``app_name`` in the ``LogRecord``,
but it was not quite easy to add a custom field without hacking arbitrary arguments in the `_log` method.

I made ``msg`` as a dictionary in the application interfaces,
so that in the handler, the record can split the ``msg`` into ``app_name`` and actual ``msg``.
I could not use ``args`` since ``scipp.WidgetHandler`` was formatting the ``args``.

## Helper Functions
- ``get_logger``: retrieves the logger with name `beamlime`.
- ``initialize_file_handler``:  creates a new log file in `~/.beamlime/logs/`.
  1. Retrieve the ``log-dir`` from the configuration or the default path and create a log file into the directory.
  2. Create a file handler with the new log file.
  3. Add the handler to the ``logger`` or `beamlime` logger.

# Updates
## Application interface
Now it uses logger instead of formatted printing!
I added logging mixin into the application interfaces such as `info` or `debug` so that it easily wraps the `app_name` into the `msg`.

# TODOs
- `logger` setter is missing in the application interface.
- logs still not printed in the jupyter notebook....

# Screen shots
## In the jupyter notebook
image
## In the log file
image


src/beamlime/logging/formatters.py
jl-wynen, line 1: We have been using %-formatting in Scipp so far. Is your setup here compatible with other code that uses %? In particular, did you test the log widget with Scipp objects? https://scipp.github.io/reference/logging.html#Formatting-Scipp-Objects

And what about 3rd party code that might log?

src/beamlime/logging/formatters.py
```diff
@@ -0,0 +1,156 @@
...
+_SepStyle: TypeAlias = Literal["|", ","]
+
+
+class LogColumn:
+    def __init__(
+        self,
+        variable_name: str,
+        title: str = None,
+        min_length: int = None,
+        style: _FormatStyle = "{",
```
jl-wynen, line 19: Why is this an argument? Do you want to support other styles in the future?

src/beamlime/logging/formatters.py
```diff
@@ -0,0 +1,156 @@
...
+from typing import Any, Literal, Mapping, TypeAlias, overload
+
+_FormatStyle: TypeAlias = Literal["{"]
+_SepStyle: TypeAlias = Literal["|", ","]
+
+
+class LogColumn:
+    def __init__(
+        self,
+        variable_name: str,
```
jl-wynen, line 16: In Scipp, we require keyword arguments for functions with easily confused arguments. E.g. multiple str-args. You should probably do this, too.

src/beamlime/logging/handlers.py
```diff
@@ -0,0 +1,75 @@
...
+        mode: str = "a",
+        encoding: str | None = None,
+        delay: bool = False,
+        errors: str | None = None,
+        header: bool = True,
+    ) -> None:
+        import logging
+
+        if not issubclass(logging.getLogRecordFactory(), BeamlimeLogRecord):
+            logging.setLogRecordFactory(BeamlimeLogRecord)
```
jl-wynen, line 39: Does this work with 3rd party code which logs?

src/beamlime/logging/handlers.py
```diff
@@ -0,0 +1,75 @@
...
+        self.header = header
+        self.color_list = [
+            Fore.BLACK,
+            Fore.GREEN,
+            Fore.BLUE,
+            Fore.RED,
+            Fore.MAGENTA,
+            Fore.LIGHTBLUE_EX,
+        ]
+        self.color_map = {"": "\033[0m"}
```
jl-wynen, line 62: Does this work cross-platform?

src/beamlime/logging/handlers.py
```diff
@@ -0,0 +1,75 @@
...
+        super().__init__()
+        self.setFormatter(BeamlimeAnsiColorFormatter())
+
+    def handle(self, record: BeamlimeColorLogRecord) -> bool:
+        if record.app_name not in self.color_map:
+            if self.color_list:
+                self.color_map[record.app_name] = self.color_list.pop()
+            else:
+                # TODO: Update this if we need more colors
+                self.color_map[record.app_name] = self.color_map[""]
```
jl-wynen, line 72: This mechanism seems odd. It can assign different colours to apps depending on what order they first log a message in. Did you consider more deterministic alternatives?

src/beamlime/core/application.py
```diff
@@ -40,19 +40,48 @@ def __del__(self) -> None:
...
+        self._log(level=WARN, msg=msg)
+
+    def exception(self, msg: str) -> None:
+        self._log(level=ERROR, msg=msg)
+
+    def error(self, msg: str) -> None:
+        self._log(level=ERROR, msg=msg)
+
+
+class BeamlimeApplicationInterface(_LogMixin, ABC, BeamLimeApplicationProtocol):
```
jl-wynen, line 63: Didn't notice earlier, but why does this inherit from `BeamlimeApplicationProtocol`?

src/beamlime/offline/data_reduction.py
```diff
@@ -152,5 +149,4 @@ async def _run(self):
             if not send_result:
                 break
             new_data = await self.receive_data(timeout=1)
-            if self.verbose:
-                print(self.verbose_option, f"Sending {result}", Style.RESET_ALL)
+            self.info(f"Sending {result}")
```
jl-wynen, line 152: Given that this message might never be emitted, it is better to use lazy formatting:
```python
self.info("Sending {}", result)
```
(provided that your formatter does {-style formatting.)

You'll need to update the logging mixin accordingly.

src/beamlime/logging/formatters.py
YooSunYoung: Yes...! The screen shot from the jupyter notebook is with the scipp log widget. I didn't change the formatter for the scipp widget though. It'll be broken if I want to change the formatter I guess...
I will add `%` and `$` indeed.

I opened the new issue about this https://github.com/scipp/beamlime/issues/37

src/beamlime/logging/formatters.py
YooSunYoung: Yes...!

src/beamlime/logging/handlers.py
YooSunYoung: It should, if they use `logging.LogRecord`, since the BeamlimeLogRecord is a subclass of LogRecord and it has default values in it.

src/beamlime/logging/handlers.py
YooSunYoung: Right... I'll change it to `colorama.Style`

src/beamlime/logging/handlers.py
YooSunYoung: I wanted to have something similar to what docker-compose does.
I couldn't think of a better solution... do you have any idea...?

src/beamlime/core/application.py
YooSunYoung: I wanted to keep minimum abstrace method in the `Protocol` while we have common methods in the `Interface`.
In principle it doesn't have to inherit but it was easier to check if it to has all methods and properties that the Protocol require...? Should it not inherit from the Protocol...?

src/beamlime/logging/handlers.py
jl-wynen: The applications could define it by setting entries in a global map. Otherwise I don't really see a flexible solution. Maybe the random colours are fine.

src/beamlime/core/application.py
jl-wynen: AFAIK, the purpose of protocols is to define an interface without requiring inheritance. If you want to enforce compliance with the interface, you should just use `abc` and no protocols.

src/beamlime/logging/formatters.py
jl-wynen: What happens with `log.info("a variable: {}", var)`?

YooSunYoung: I found out why...!
The `args` are formatted by `%` by default in the LogRecord and the `BeamlimeLogRecord`` doesn't override `getMessage` so scipp object is formatted by `%` if you pass it as `args`.
And now I remember that it was also why I couldn't use `args` for formatting `app_name`.

```
    def getMessage(self):
        """
        Return the message for this LogRecord.

        Return the message for this LogRecord after merging any user-supplied
        arguments with the message.
        """
        msg = str(self.msg)
        if self.args:
            msg = msg % self.args
        return msg
```

YooSunYoung: I can either make this `@final` so that you can only use `%` in the message,
or update `getMessage` to handle all three format styles....?

src/beamlime/logging/records.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class BeamlimeLogRecord(LogRecord):
+    """
+    Note
+    ----
+    First 8 Arguments cannot be keyword-only arguments
+    because of the way ``logging`` creates an empty record.
+    It uses positional arguments to create a record instance.
+    """
```
jl-wynen, line 27: Not really a docstring. Can you turn this into a comment instead?

src/beamlime/logging/records.py
```diff
@@ -0,0 +1,122 @@
...
+        self.app_name = app_name
+        if record is not None:
+            self.__copy_from__(record)
+        else:
+            super().__init__(
+                name, level, pathname, lineno, msg, args, exc_info, func, sinfo
+            )
+
+    def __copy_from__(self, record: LogRecord) -> None:
+        self.__dict__.update(record.__dict__)
```
jl-wynen, line 76: Interesting solution. Is this a standard way of doing this?

It seems that if `record` is a subclass of `LogRecord`, this could break. Does this matter here?

src/beamlime/logging/records.py
```diff
@@ -0,0 +1,122 @@
...
+        self.msg = self.msg.removeprefix(app_name_col)
+        return super().getMessage()
+
+
+class BeamlimeColorLogRecord(BeamlimeLogRecord):
+    _extra_defaults = {
+        "app_name": "",
+        "ansi_color": "",
+        "reset_color": Style.RESET_ALL,
+    }
```
jl-wynen, line 89: Any reason to have this extra dict here? You could also put the default arguments directly in `__init__`.

src/beamlime/offline/proto.ipynb
```diff
@@ -24,7 +55,10 @@
    },
    "outputs": [],
    "source": [
-    "manager = BeamLimeApplicationManager()"
+    "# get_logger().addHandler(BeamlimeStreamHandler(header=True))\n",
+    "# get_logger().setLevel('INFO')\n",
+    "# manager = BeamLimeApplicationManager(save_log=True)\n",
```
jl-wynen, line 60:
```suggestion
```

src/beamlime/logging/formatters.py
```diff
@@ -0,0 +1,298 @@
...
+                fmt + _reset_color_col.formatter if _color_ends[icol] else fmt
+                for icol, fmt in enumerate(_color_started_fmts)
+            ]
+
+            self._fmt = sep.join(_colored_fmts)
+
+        for column in columns:
+            self.__setitem__(column.variable_name, column)
+
+    def _init_from_literal_fmt(self, *fmts: str) -> None:
```
jl-wynen, line 198: Instead of the custom parsing logic here, you can maybe use https://docs.python.org/3/library/string.html?highlight=string#string.Formatter to make this more robust.

src/beamlime/logging/formatters.py
```diff
@@ -0,0 +1,298 @@
...
+        else:
+            sep = " " * self.padding[0] + self.sep + " " * self.padding[1]
+            _ansi_color_col = LogColumn("ansi_color", title="", ansi_colored=False)
+            _reset_color_col = LogColumn("reset_color", title="", ansi_colored=False)
+
+            _padded_columns = [_ansi_color_col] + list(columns) + [_reset_color_col]
+
+            # Color starting/terminating flags
+            _color_starts = [
+                True if column.ansi_colored and not p_column.ansi_colored else False
```
jl-wynen, line 173:
```suggestion
                column.ansi_colored and not p_column.ansi_colored
```
And similar below

src/beamlime/logging/handlers.py
```diff
@@ -0,0 +1,86 @@
...
+        errors: Union[str, None] = None,
+        header: bool = True,
+    ) -> None:
+        self.header = header
+        super().__init__(filename, mode, encoding, delay, errors)
+        self.setFormatter(BeamlimeHeaderFormatter())
+
+    def handle(self, record: BeamlimeLogRecord) -> bool:
+        if not isinstance(record, BeamlimeLogRecord):
+            return super().handle(BeamlimeLogRecord(record))
```
jl-wynen, line 46: Why `super`?

src/beamlime/logging/handlers.py
```diff
@@ -0,0 +1,86 @@
...
+        ]
+        self.color_list = cycle(self.palette)
+        self.color_map = {"": Style.RESET_ALL}
+        super().__init__()
+        self.setFormatter(BeamlimeAnsiColorFormatter())
+
+    def set_application_color(
+        self, app_name: str, ansi_color: Optional[str] = None
+    ) -> None:
+        self.color_map.update({app_name: ansi_color or next(self.color_list)})
```
jl-wynen, line 72:
```suggestion
        self.color_map[app_name] = ansi_color or next(self.color_list)
```

src/beamlime/logging/loggers.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import Any, Callable, Union
+
+
+def hold_logging(logging_func: Callable) -> Callable:
+    from functools import wraps
+    from inspect import signature
+
+    @wraps(logging_func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        logging._acquireLock()
```
jl-wynen, line 15: Is there are way to do this without using protected attributes of `logging`?

src/beamlime/logging/loggers.py
```diff
@@ -0,0 +1,127 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+
+import logging
+import sys
+from typing import Any, Callable, Union
+
+
+def hold_logging(logging_func: Callable) -> Callable:
```
jl-wynen, line 9: What does this function do?

src/beamlime/logging/loggers.py
```diff
@@ -0,0 +1,127 @@
...
+    def __init__(self, name: str, level: Union[int, str] = 0) -> None:
+        from .records import BeamlimeColorLogRecord
+
+        self._logRecordFactory = BeamlimeColorLogRecord
+        super().__init__(name, level)
+
+    def _wrap_stack_info(
+        self, stack_info: bool = False, stacklevel: int = 1
+    ) -> tuple[str, int, str, Union[str, None]]:
+        if logging._srcfile:
```
jl-wynen, line 44: Another protected attribute. Why do you need this?

src/beamlime/logging/loggers.py
```diff
@@ -0,0 +1,127 @@
...
+                "a matching formatter"
+                f" with the LogRecord type {self._logRecordFactory}.\n",
+                key_err,
+            )
+
+        try:
+            if not (hdlr in self.handlers):
+                self.handlers.append(hdlr)
+        finally:
+            pass
```
jl-wynen, line 127:
```suggestion
        if not (hdlr in self.handlers):
            self.handlers.append(hdlr)
```
?

src/beamlime/logging/loggers.py
jl-wynen, line 1: This is all quite complicated and tied into internals of the logging package. What is the benefit of using this over a default logger and records?

The app name could be encoded in the logger name along the lines of `beamlime.data_feeder` (logging supports 'child' logger like this).

src/beamlime/logging/formatters.py
YooSunYoung: Now it kinda supports `%`...! But there are some TODOs for fancy features : D...
I opened an issue regarding this : #37

src/beamlime/logging/handlers.py
YooSunYoung: I decided not to change global `LogRecordFactory` and use more general form of log-call instead.

src/beamlime/logging/records.py
YooSunYoung: Actually it's not true anymore. I will delete it.

src/beamlime/logging/records.py
YooSunYoung: `LogRecord` doesn't have this function.
I needed to somehow create a new `Beamlime*Record` from existing `LogRecord` when the logger is not a native one.
and standard formatters only use ``LogRecord.__dict__``, so I came up with this...?
It shouldn't break anything since it's just updating ... no...?

src/beamlime/logging/formatters.py
YooSunYoung: I need to update this part anyways. I'll do it when I work on this: #37

src/beamlime/logging/loggers.py
YooSunYoung: I updated not to use them...!

src/beamlime/logging/loggers.py
YooSunYoung: I locks the thread and it's often used in many places in ``logging`` so I thought it'd be better to have it as a decorator since both of them must be called before and after the function call.

src/beamlime/logging/loggers.py
YooSunYoung: > _srcfile is used when walking the stack to check when we've got the first
> caller stack frame, by skipping frames whose filename is that of this
> module's source. It therefore should contain the filename of this module's
> source file.
> Ordinarily we would use __file__ for this, but frozen modules don't always
> have __file__ set, for some reason (see Issue #21736). Thus, we get the
> filename from a handy code object from a function defined in this module.
> (There's no particular reason for picking addLevelName.)

I updated this part, I procrastinated since I had to rewrite ``findCaller``...

src/beamlime/logging/loggers.py
YooSunYoung: Yes... it all started because I didn't know it'd be complicated to add one more keyword argument in the `log` function...

But, isn't then the child logger will be a different instance..? I wanted to make a logger that collects global logs.

src/beamlime/logging/records.py
YooSunYoung: It was originally for other purposes but it's not redundant so I removed these...!

jl-wynen (approved):

PR #40, opened 2023-04-17
Add pytest CI action

YooSunYoung:
Github actions will run based on these conditions below.

## pull-request - pr.yml
 Run `tox -e static` and `tox -e test` only on a single environment (ubuntu environment with python version 3.9)
 Upload coverage-report as an artifact.

## merge to main - main.yml
 Run `tox -e static` on ubuntu environment with python version 3.9
 Run `tox -e test` on 9 different environments (ubuntu/mac-os/windows with python version 3.9/3.10/3.11)

``main.yml`` will also be triggered when the file itself is updated in the pull-request.


SimonHeybrock:
> ## pull-request - pr.yml
> 
> Run `tox -e static` and `tox -e test` only on a single environment (ubuntu environment with python version 3.9) 
> [...]
> ## merge to main - main.yml
> 
> Run `tox -e static` on ubuntu environment with python version 3.9 Run `tox -e test` on 9 different environments (ubuntu/mac-os/windows with python version 3.9/3.10/3.11)


You are not going to be happy with this, since we will have broken `main` builds in many cases (most common candidate: Windows, since its Python `int` is 32 bit by default).

That being said, do we even need Windows? The ESS instruments will definitely run Linux.

YooSunYoung:
> > ## pull-request - pr.yml
> > Run `tox -e static` and `tox -e test` only on a single environment (ubuntu environment with python version 3.9)
> > [...]
> > ## merge to main - main.yml
> > Run `tox -e static` on ubuntu environment with python version 3.9 Run `tox -e test` on 9 different environments (ubuntu/mac-os/windows with python version 3.9/3.10/3.11)
> 
> You are not going to be happy with this, since we will have broken `main` builds in many cases (most common candidate: Windows, since its Python `int` is 32 bit by default).
> 
> That being said, do we even need Windows? The ESS instruments will definitely run Linux.

I am very happy to remove Windows : D....

.github/workflows/pr.yml
```diff
@@ -0,0 +1,24 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+
+name: CI
+
+on:
+  push:
+    branches:
+      - main
```
SimonHeybrock, line 9: So this is running in addition to the `main.yml`?

tox.ini
```diff
@@ -8,12 +8,27 @@ commands = python -m pytest --cov=beamlime --cov-report html:coverage_html/pytho
...
 [testenv:deps]
 description = Update dependencies by running pip-compile-multi
-basepython = python3.9
 skip_install = true
 deps = pip-compile-multi
 changedir = requirements
 commands = pip-compile-multi -d .
 
+[testenv:config-build]
+description = Build new default configuration yaml file
```
SimonHeybrock, line 17: Can you extend this description, it is not clear to me as an outsider when and how this should be used/run.

.github/workflows/reusable-test.yml
```diff
@@ -0,0 +1,51 @@
...
+        shell: bash -l {0}
+    strategy:
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: Setup conda environment
+        uses: mamba-org/provision-with-micromamba@main
```
SimonHeybrock, line 37: Using conda/mamba because...? Do we depend on anything that is conda-only?

.github/workflows/reusable-test.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          micromamba-version: 1.0.0
+          environment-file: docs/environments/conda/${{ inputs.build-config }}.yaml
```
SimonHeybrock, line 40: Are you planning on having multiple different environments? If so, for what purpose?

YooSunYoung: I don't know, I copied and pasted this from other repositories and 
I thought it'd be better to keep it re-usuable for other repositories in case we want to merge everything...?

.github/workflows/reusable-test.yml
YooSunYoung: It was just much faster to build the virtual environment with mamba.
Should this be avoided...?

.github/workflows/pr.yml
YooSunYoung: No it shouldn't be...! I updated.

.github/workflows/reusable-test.yml
SimonHeybrock: I am asking you you are not using `pip` (with `tox` to handle the virtual envs), e.g., like in scippnexus or plopp.

YooSunYoung: Using mamba&tox was sliiightly faster but basically no difference,
and as you said, it'll be better to use ``ci.txt`` instead of ``lime-dev.yml``, so I updated it to use pip&tox.

### Using mamba&tox
image
[link to the action summary](https://github.com/scipp/beamlime/actions/runs/4743039071)

### Using pip&tox
image
[link to the action summary](https://github.com/scipp/beamlime/actions/runs/4733193251)

Issue #38, opened 2023-04-11
Performance tests


SimonHeybrock:
We have the approximate boundary conditions outlined below. We should implement performance tests early on to ensure that the architecture is capable of handling this. The tests should be able to run at least semi-automatically, to ensure that they can easily be re-run with changes, to ensure we keep meeting the metrics.

- Event rates of up to 1e7 events/second. There may be instruments with more later (when the accelerator power is higher), but I think 1e7 is a good initial benchmark.
- Up to 10 million pixels. Typical is closer to 1 million.
- Updates of NXlog info from dozens or even more than 100 time-series logs. We can use an actual NeXus file as a baseline and then figure out how to model updates for all the NXlog groups.
- Live update of visualizations about once per second.

SimonHeybrock:
Here is a notebook that can add a large number of events to our existing DREAM file (which will be an early offender in terms of high event rates):

[large-file.ipynb.zip](https://github.com/scipp/beamlime/files/11286267/large-file.ipynb.zip)

SimonHeybrock:
I suggest as a first step before using actual neutron data we do something simpler:

- Some large NumPy arrays, initialized somehow.
- Should test sizes equivalent to the expected ranges, i.e., up to 1e7 (testing with 5e7 as a bonus) events/second -> 1e6 events per pulse, which would mean about 1e6*2*8 Byte = 16 MByte per pulse.
  - Time for concatenating a few pulses (1-10 seconds worth of pulses).
- Feed this data though the pipeline, with some operations to reduce it to 1-D.

SimonHeybrock:
Here is some code for the simplest somewhat realistic test case, event data similar to what we get at some stage during loading/processing a chunk:

```python
import scipp as sc
from numpy.random import default_rng
rng = default_rng(seed=1234)


def make_event_data(*, npixel, nevent) -> sc.DataArray:
    weights = sc.ones(dims=['event'], shape=[nevent], unit='counts')
    pulse_time = sc.epoch(unit='ns') + sc.arange(
        'event', nevent, unit='ns', dtype='int64'
    )
    tof = sc.array(dims=['event'], values=rng.random((nevent,)), unit='s')
    pixel = sc.array(dims=['event'], values=rng.integers(npixel, size=(nevent,)), unit=None)
    da = sc.DataArray(weights, coords={'pulse_time': pulse_time, 'tof': tof, 'pixel':pixel})
    return da
```

Then we have to:

```python
[...]

PR #45, opened 2023-04-18
Multi-application handling interface part-1

YooSunYoung:
> `part-2` will be the complete update of the `beamlime.core.system` module.

Main changes: 
  - Under `beamlime.applications`, three main types of application,
    `data-feeder`, `data-reduction` and `visualization` were introduced and they are removed from `offline` module.
  - Each application instances can now be controlled by `start`, `stop`, `pause` and `resume` interface, but they only control `_started`, `_paused`, `_stopped` boolean flags. The flags can be used in the `_run`. _I'm not sure if it's the best way of controlling the process... but other than that, maybe the `BeamlimeSystem` should also have functions that abruptly stop application instances, in case `stop` does not kill the process...?_
  - `async_timeout` helper function in `beamlime.core.schedulers` module.
  - More await calls among applications.
  - Reorganized protocols and interfaces, mainly focusing on Application part since the `BeamlimeSystem` and `Communication` part are still WIP
  - Unit test of protocol checks for the application interfaces.
  - `beamlime.core.system` module introduced for multi-application handling, still WIP, blocked by `Communication` interface progress.
    -> Using `beamlime.communication.queue_handlers` module as a temporary solution

TODO: New sample system of three applications `data-feeder`, `data-reduction` and `visualization`
with the new fake data generator @SimonHeybrock provided here: #38


src/beamlime/applications/interfaces.py
```diff
@@ -1,195 +1,35 @@
...
-            except TypeError:
-                self._task = asyncio.create_task(self._run(), name=name)  # py39, py310
-        return self._task
-
-
-class BeamlimeApplicationInterface(_LogMixin, _FlagControlMixin, _DaemonInterface, ABC):
     def __init__(self, config: dict = None, logger=None, **kwargs) -> None:
         self._pause_interval = 0.1
         self._init_logger(logger=logger)
         self.app_name = kwargs.get("name", "")
```
nvaytet, line 26: Why use `get` on `kwargs` here (and see below) and not have function arguments with default values (which would also make generating API docs more easy)? Is there a technical reason I am missing?

src/beamlime/applications/interfaces.py
```diff
@@ -1,195 +1,35 @@
...
         self._input_ch = None
         self._output_ch = None
+        self._update_rate = kwargs.get("update_rate", 1)
+        self._timeout = kwargs.get(
```
nvaytet: Can we use function arguments with default values instead of `kwargs.get()`?

src/beamlime/core/protocols.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
+from asyncio import Task
 from typing import Any, Protocol, runtime_checkable
```
nvaytet, line 7: I cannot comment on L9, but does it make sense to rename `BeamlimeApplicationControlProtocol` to `BeamlimeControlProtocol`? I got confused while reading between `BeamlimeApplicationControlProtocol` and `BeamlimeApplicationProtocol` (I just saw something starting with `BeamlimeApplication` and ending in `Protocol` and started mixing them up).

No worries if you think it's necessary to keep it as is.

src/beamlime/applications/interfaces.py
```diff
@@ -228,185 +80,39 @@ def output_channel(self):
     def output_channel(self, output_channel):
         self._output_ch = output_channel
 
-    @async_timeout(Empty)
-    async def receive_data(self, *args, timeout=10, wait_interval=0.2, **kwargs):
-        # TODO: Move async_timeout(exception=Empty) to communication handler interface
-        # and remove the decorator or use @async_timeout(exception=TimeoutError).
-        return self.input_channel.get(*args, timeout=wait_interval, **kwargs)
+    async def receive_data(self, *args, **kwargs) -> Any:
```
nvaytet, line 83: This interface has `receive_data` and `send_data` methods defined, which are specified in the `BeamlimeDownstreamProtocol`. Should it be both a `BeamlimeApplicationProtocol` and a `BeamlimeDownstreamProtocol`?

src/beamlime/applications/visualization.py
```diff
@@ -66,10 +58,10 @@ async def process(self, new_data):
         return self.info("value updated with frame number %s", frame_number)
 
     async def _run(self):
-        new_data = await self.receive_data(timeout=20, wait_interval=1)
```
nvaytet: :+1: for not needing these values here anymore.

src/beamlime/applications/groups.py
```diff
@@ -0,0 +1,225 @@
...
+    Protocol
+    --------
+    BeamlimeApplicationProtocol
+
+    """
+
+    def __init__(
+        self,
+        constructor: Callable,
+        instance_num: int = 2,
```
nvaytet, line 64: Is 2 a good default? From what I can see below, you currently cannot add and remove instances dynamically (like appending and popping from a list or a dict of instances).
Do you want the `__init__` to actually create the instances, or do you want the group to just take in as input instances that are already created?

My experience with groups structures is that they end up being more flexible if they just take in things that are already created. You can still perform some initialization inside the group `__init__`, but sometimes you may want to take in some instances that you want to initialize slightly differently, and then if you are creating them inside the group `__init__` that becomes more difficult to achieve.

src/beamlime/applications/groups.py
```diff
@@ -0,0 +1,225 @@
...
+        if logger is None:
+            from ..logging import get_logger
+
+            self.logger = get_logger()
+        else:
+            self.logger = logger
+
+    def _init_instances(self, constructor: Callable, instance_num: int) -> None:
+        self.constructor = constructor
+        self._instances = []
```
nvaytet, line 88: I think maybe a `dict` of instances instead of a `list` may allow for easier debugging if you want to inspect a particular instance?

src/beamlime/applications/data_reduction.py
```diff
@@ -33,22 +33,10 @@ def heatmap_2d(data, threshold=0.2, binning_size=(64, 64)):
 method_map = {"heatmap_2d": heatmap_2d, "handover": lambda x: x}
```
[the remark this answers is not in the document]
YooSunYoung: I couldn't really write a general handler that can be used for most of `data-reduction` applications yet.
Since now I have clear idea that there should be `data-feeder(pre-data-reduction)`, `data-reduction` and `visualisation`, I should be able to write a general one, but for now they are more or less an example, how to write a `handler` class with the `BeamlimeApplicationInterface`.

So, the `heatmap_2d` is only for this specific one, and didn't intend to use in other handlers.
This `BeamlimeDataReductinoApplication` is one of the `handler` classes that you can use in the configuration.

Maybe I should move those `handler`s to another sub package and keep `applications` only with general ones...

Related to #48

src/beamlime/applications/mixins.py
```diff
@@ -0,0 +1,180 @@
...
+    Process control interfaces
+
+    Protocol
+    --------
+    BeamlimeApplicationControlProtocol
+    """
+
+    _started = False
+    _paused = True
+    _stopped = False
```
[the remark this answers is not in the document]
YooSunYoung: Yes, the initial idea was,
you might want to know if this application was ever started or not,
even though the coroutine, `_run`, was suspended at some point before.
I'm still not sure about what's the best practice but I agree with you that we don't need these many flags right now.
I'll make it simpler.

src/beamlime/applications/mixins.py
```diff
@@ -0,0 +1,180 @@
...
+    Logging interfaces
+
+    Protocol
+    --------
+    BeamlimeLoggingProtocol
+    """
+
+    def _log(self, level: int, msg: str, args: tuple):
+        if isinstance(self.logger, BeamlimeLogger):
+            self.logger._log(level=level, msg=msg, args=args, app_name=self.app_name)
```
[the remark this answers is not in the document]
YooSunYoung: Yes indeed... I intentionally forgot to add properties in the `Protocol`
since it doesn't allow `issubclass` when it has non-method members... I will add them.

src/beamlime/applications/mixins.py
```diff
@@ -0,0 +1,180 @@
...
+               # receives data, processes and sends the result every 0.1 second.
+    # or if there is another coroutine running,
+    app.create_task()
+    ```
+    """
+
+    async def should_proceed(self):
+        @async_timeout(ApplicationPausedException)
+        async def wait_resumed(timeout: int, wait_interval: int):
+            if not self._stopped and self._paused:
```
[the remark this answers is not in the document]
YooSunYoung: Yes, all three mixins assumes the inheriting class meets the `BeamlimeApplicationProtocol`.
Is it maybe too much assumptions...?

src/beamlime/applications/interfaces.py
```diff
@@ -1,195 +1,35 @@
...
-        return self._task
-
-
-class BeamlimeApplicationInterface(_LogMixin, _FlagControlMixin, _DaemonInterface, ABC):
     def __init__(self, config: dict = None, logger=None, **kwargs) -> None:
         self._pause_interval = 0.1
         self._init_logger(logger=logger)
         self.app_name = kwargs.get("name", "")
         self._input_ch = None
         self._output_ch = None
```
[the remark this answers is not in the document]
YooSunYoung: Yes, they do...! And indeed they are ambiguous, I'll update them.
I am 90% sure that `input_channel` and `output_channel` will be removed once I implement proper communication interfaces...!

src/beamlime/applications/interfaces.py
YooSunYoung: It was only because I was not sure where those variables should be set, either `__init__` or `parse_config` and the `parse_config` ... 
and it was easier to keep this form in case I want them to be parsed from `config` : D.... 
Still not sure how should they be, either
1. `ApplicationGroup` parses the configuration and create instance accordingly,
2. `handler` constructor should have all the key-word arguments with default values that is expecting from the configuration so that `BeamlimeSystem` can simply pass the subset of configuration...? 
I'll figure it out as soon as possible...

src/beamlime/applications/interfaces.py
YooSunYoung: Yes...! They will be done with communication interfaces...!

src/beamlime/applications/groups.py
YooSunYoung: Yes...! I updated the default value to be 1. I forgot to update them after I tested this code : D....

And also I agreed that we shouldn't create the instances at the `__init__` so I removed that part and let the `BeamlimeSystem` do the job. It has the method `init_instances`.
Also, the `populate` method can create more instances.

src/beamlime/applications/interfaces.py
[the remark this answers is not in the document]
YooSunYoung: It's fixed in #49

src/beamlime/applications/interfaces.py
YooSunYoung: Also fixed in #49

tests/applications/coroutine_mixin_test.py
```diff
@@ -0,0 +1,46 @@
...
+    logger.setLevel(DEBUG)
+
+    app_name = "Dummy Daemon Application"
+    app = DummyDaemonApp(config={}, name=app_name)
+    app.start()
+    del app
+
+    log_output = capsys.readouterr()[-1]
+    assert (
+        f"{app_name:15} | {getLevelName(DEBUG):8} | "
```
nvaytet, line 42: Having tests based on hard-coded string lengths feels quite brittle. In addition, the `:15` and `:8` are implementation details that the test should not have to know about.
I would suggest to make the requirements less hard, such as simply checking the `app_name` and `DEBUG` are there in the output?

tests/applications/flag_control_mixin_test.py
```diff
@@ -0,0 +1,114 @@
...
+    app.start()
+    assert app._started and not app._paused
+
+
+def test_start_twice():
+    app = DummyApp()
+    app.start()
+    assert app._started and not app._paused
+    # app.start()
+    # assert app._started and not app._paused
```
nvaytet, line 20: uncomment?

tests/applications/flag_control_mixin_test.py
```diff
@@ -0,0 +1,114 @@
...
+    app.start()
+    app.pause()
+    assert app._started and app._paused
+
+
+def test_pause_twice():
+    app = DummyApp()
+    app.start()
+    app.pause()
+    app.pause()
```
nvaytet, line 42: Some people might expect that calling `pause` twice pauses once and then resumes (like on e.g. a CD player).
We should make sure the behaviour of `pause` is well documented to reflect this.

tests/applications/flag_control_mixin_test.py
```diff
@@ -0,0 +1,114 @@
...
+    app = DummyApp()
+    app.start()
+    app.resume()
+    assert app._started and not app._paused
+
+
+def test_resume_not_started():
+    app = DummyApp()
+    app.resume()
+    assert app._started and not app._paused
```
nvaytet, line 74: I think it's strange than `resume` can start the app if it has not been started yet. I would have expected an error to be raised.

tests/applications/flag_control_mixin_test.py
```diff
@@ -0,0 +1,114 @@
...
+    # app.start()
+    # assert app._started and not app._paused
+
+
+def test_start_twice_paused():
+    app = DummyApp()
+    app.start()
+    app.pause()
+    app.start()
```
nvaytet, line 27: I don't know if calling `start` if the app is already started should do nothing, as now, or if an exception should be raised?

I guess it depends on what we want to have. Either something that is pretty flexible and will mostly keep running even if it receives commands which are incompatible with its current state, or something that is much more rigid but lets us know immediately if it is receiving the wrong kind of commands...

tests/applications/flag_control_mixin_test.py
```diff
@@ -0,0 +1,114 @@
...
+    app.pause()
+    app.resume()
+    assert app._started and not app._paused
+
+
+def test_resume_not_paused():
+    app = DummyApp()
+    app.start()
+    app.resume()
```
nvaytet, line 67: Maybe should raise if app is not paused?

tests/applications/flag_control_mixin_test.py
```diff
@@ -0,0 +1,114 @@
...
+    assert app._started and not app._paused
+
+
+def test_resume_paused_twice():
+    app = DummyApp()
+    app.start()
+    app.pause()
+    app.resume()
+    app.resume()
```
nvaytet, line 82: Should raise?

tests/applications/flag_control_mixin_test.py
YooSunYoung: Yes, that is true.

I think the underlying interface should be as it is, but on the user-interface,
we will want the `pause` button to be grayed out or change the label as `resume` as you said.

I will put this in the `GUI` requirements.

src/beamlime/applications/groups.py
```diff
@@ -197,13 +199,14 @@ def create_task(self) -> list[asyncio.Task]:
...
+        from uuid import uuid4
+
         if len(self._instances) < MAX_INSTANCE_NUMBER:
-            self._instances.append(self.constructor())
-            if hasattr(self, "_output_ch"):
-                self._instances[-1].output_channel = self._output_ch
-            if hasattr(self, "_input_ch"):
-                self._instances[-1].input_channel = self._input_ch
-            return self._instances[-1]
+            inst_name = self.app_name + str(uuid4())
```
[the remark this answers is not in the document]
YooSunYoung: I'd rather keep the same style, thanks!

tests/applications/flag_control_mixin_test.py
YooSunYoung: Yes, I agree. So I updated it to raise an error when `resume` was called before `start`.

src/beamlime/applications/mixins.py
```diff
@@ -2,27 +2,36 @@
...
     BeamlimeLoggingProtocol
     """
 
+    _logger = None
+
+    @property
+    def logger(self) -> Logger:
+        return self._logger
+
+    @logger.setter
```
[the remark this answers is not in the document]
YooSunYoung: But I did update it to be read-only now. We can add setter later if needed...!

tests/applications/flag_control_mixin_test.py
```diff
@@ -16,16 +28,18 @@ def test_start_twice():
     app = DummyApp()
```
nvaytet, line 28: We have adopted the unofficial convention that when we make tests that are supposed to verify an exception is raised, we reflect it in the test name. So in this case, this test could be called something like `test_start_twice_raises()`

tests/applications/flag_control_mixin_test.py
```diff
@@ -16,16 +28,18 @@ def test_start_twice():
...
     app.start()
     assert app._started and not app._paused
-    # app.start()
-    # assert app._started and not app._paused
+    with pytest.raises(ApplicationStartedException):
+        app.start()
+        assert app._started and not app._paused
```
nvaytet, line 33: I think (?) this line is never reached, it exists the context as soon as the error is raised.
The checks on the flags should be outside.

YooSunYoung: Oh.. you are right...! I fixed it...! Thank you

nvaytet (approved):

Issue #43, opened 2023-04-18
Multiple ``BeamlimeApplication`` orchestrating interface.


YooSunYoung:
Currently, the offline/proto.py has some idea of how the applications will be managed, but it's currently just-barely-working and not flexible enough for other environments than 
It supports jupyter notebook visualization and multi-thread Queue communication,
but up/down stream data-flow between multi-process should also be possible.

Therefore we need to define better interface for orchestrating ``BeamlimeApplication``s.
The ``BeamlimeApplication`` itself should also be improved.

YooSunYoung:
This one is no longer needed right now due to dependency injection framework.

Issue #41, opened 2023-04-18
Upstream/Downstream  protocols for applications


YooSunYoung:

YooSunYoung:
It is no longer needed via sharing data buffer between applications instead of streaming data through different instances

PR #49, opened 2023-04-26
Communication broker - part 1 / Multi-application handling interface part-2

YooSunYoung:
I fixed some problems from #45 thanks to the communication protocol updates.

Related to #41, #43

## Main changes
- Communication channel mapping per application is now done by the communication `broker`, `beamlime.communications.broker.CommunicationBroker`.
- `broker` follows the `BeamlimeBrokerProtocol` which inherits `CommunicationProtocol`.
- `broker` will `get`, `put`, `consume` and `produce` data according to the type of channel that application is subscribing to.
- `data-stream` configuration now has all application specs including `data-reduction`.
- `__init__` function of `BeamlimeApplicationInterface` updated so that it does not need to have `config` as a dictionary and parse it. All specs can be dumped by **config when it's instantiated.
- You can test this with the `src/beamlime//offine/offline_system_example.ipynb`. It will be moved to docs later...!

## TODO - Things I will continue after the vacation!
- `put` and `poll` method implementation.
- Add tests for communication interfaces.
- Add communication handles. (Currently single-process queue only. We need a blackboard, multi-processing queue and kafka, kafka interfaces are already tested but still need to implement auto-configuration part.)
- Figure out what issues we will have in multi-processing, since now all application is sharing one broker instance. (I have an idea of instance twins...!)
- Add logging into broker and data flow monitor.


YooSunYoung:
I will merge this one since it was reviewed in other PR and will be reviewed again in the PR #53

PR #57, opened 2023-05-22
Simple configuration

YooSunYoung:
The configuration files had relatively complicated nested structure due to bad initial design.
Based on the current sample applications, configurations could be more simple by extracting general parts, defining minimum configuration and removing unnecessary options.

## Main changes
- Have matching keys as the keyword arguments of a constructor so that the keys don't need any modification.
- Remove unimplemented configurations.

## TODO:
- Configuration validation is still not implemented.
- Exporting the run-time configuration (with real values that were used for the process) should be implemented.


src/beamlime/applications/groups.py
```diff
@@ -16,19 +15,34 @@
...
     but it should pass the protocol test.
     See ``protocol_test.test_instace_group_protocol`` of the unit test package.
 
     Parameters
     ----------
-    app_name: str
+    name: str
         Name of the application.
 
+    constructor: str
```
SimonHeybrock, line 31: We are using https://github.com/tox-dev/sphinx-autodoc-typehints, so there is no need to repeat the type in the docstring. The Sphinx plugin will take care of that. Not sure if this is configured for `beamlime` yet.

src/beamlime/config/builders.py
```diff
@@ -1,145 +1,84 @@
...
+    return config
 
 
-def _channel_config(
-    name: str,
-    channel_type: Literal[
-        "MQUEUE", "SQUEUE", "KAFKA-PRODUCER", "KAFKA-CONSUMER", "BULLETIN-BOARD"
-    ] = "SQUEUE",
-    **other_specs,
+def micro_service_config(
```
SimonHeybrock, line 29: Docstrings would be useful to explain the context, I have no idea what this does, and where it is to be used as part of the larger architecture.

src/beamlime/config/builders.py
```diff
@@ -148,109 +87,116 @@ def build_offline_fake2d_config():
...
-    config["application-subscriptions"] = [
-        _app_subscription("data-generator", "kafka-producer"),
-        _app_subscription("data-feeder", "kafka-log-consumer", "kafka-data-consumer"),
-    ]
-    return config
+    return ServiceRecipe(name, constructor, specs)
 
 
-def build_fake_event_kafka_config():
+def _fake_event_data_stream_specs() -> dict:
```
SimonHeybrock, line 199: It seems odd to have this fake-creation logic (similar above), including the configuration, mixed in with all the other code. Would it make sense to separate this more?

src/beamlime/config/preset_options.py
```diff
@@ -1,39 +1,144 @@
...
-    def DEFAULT(cls) -> Union[str, int]:
-        pass
+    @staticmethod
+    def maximum() -> PresetValueType:
+        ...
+
+
+class SystemClass:
+    """
+    Beamlime system class.
```
SimonHeybrock, line 36: What system? What is a system class?

src/beamlime/core/system.py
```diff
@@ -1,79 +1,140 @@
...
             logger=logger,
-            timeout=self.system_config.get("timeout", 600),
-            wait_interval=self.system_config.get("wait-interval", 1),
+            timeout=timeout,
+            wait_interval=wait_interval,
         )
+
+        self._init_logger(save_log, log_dir)
+        self.broker = self._build_broker(communications, subscriptions)
+        self.applications = self._build_application(applications)
```
SimonHeybrock, line 91: It feels like there may be an advantage in invert the logic here. Right now, we pass things to `BeamlimeSystem`, which then constructs some sub-components. Dependency injection might be a better design choice here, making the entire thing less coupled, and more flexible (also for testing). I don't know if you have already considered that option?

src/beamlime/core/system.py
```diff
@@ -97,18 +158,11 @@ def kill(self) -> None:
             self._tasks.clear()
 
     def __del__(self) -> None:
-        while self.applications:
-            app_name, app_inst_gr = self.applications.popitem()
-            app_inst_gr.kill()
-            self.info("Deleting %s instance ... ", app_name)
-            del app_inst_gr
+        del self.applications
```
SimonHeybrock, line 161: Why is this explicit `del` of `applications` necessary?

src/beamlime/logging/loggers.py
```diff
@@ -3,24 +3,21 @@
...
 from typing import Union
 
 from .sources import find_caller
 
 
-class hold_logging(ContextDecorator):
-    def __enter__(self):
-        from . import _lock
+@contextmanager
+def hold_logging() -> None:
```
SimonHeybrock, line 13: Is this for logging multiple messages that form an atomic unit, i.e., should not get interleaved with other messages?

src/beamlime/logging/logging.py
```diff
@@ -120,5 +134,49 @@ def get_logger(
...
+            return logging.getLogger(name)
+    else:
+        return logging.getLogger(name)
+
+
+def safe_get_logger(
+    logger: Union[logging.Logger, None] = None, verbose: bool = True
+) -> logging.Logger:
+    """
+    Do nothing if the logger si a subclass of ``logging.Logger``.
```
SimonHeybrock, line 148:
```suggestion
    Do nothing if the logger is a subclass of ``logging.Logger``.
```

src/beamlime/resources/exporters.py
```diff
@@ -76,16 +79,17 @@ def export_preset_configs(
             "# Please don't update it manually.\n"
             "# Use `tox -e config-build` to generate a new one.\n\n"
         ),
-        order=["general", "data-stream"],
+        # order=["general", "data-stream"],
```
SimonHeybrock, line 82: Commented code?

src/beamlime/core/system.py
YooSunYoung: Yes, that was ultimately what I wanted to achieve, so I wanted to have an helper to build a pipeline.
And then it turned out to be more like a controller...
I will separate that part and inject them instead, and I thought I should remove that optional logger logic from the constructor as well so it makes sense.

src/beamlime/config/preset_options.py
YooSunYoung: Initially it was a helper class to build a pipe line and then now it's a wrapper of application instances that controls them all at once... I'll add description and think of a better class name....

src/beamlime/core/system.py
SimonHeybrock: 👍 (but maybe do that in a different PR?)

src/beamlime/core/system.py
YooSunYoung: I forgot to delete this after using the debugger ...

YooSunYoung:
Close due to bigger updates

PR #58, opened 2023-06-16
Wipe out source code for easier review on next bigger updates

YooSunYoung:


YooSunYoung:
Agreed to wipe out code base first, but keeping history at the stand-up meeting ;)

PR #59, opened 2023-06-16
Dependency injection framework with tests

YooSunYoung:
Minimum dependency injection framework that assumes all arguments and attributes of provider functions/class need sub-dependencies injected.
Tests have coverage 100% but still might be missing some cases.

TODO: Documentation is not... yet...!


src/beamlime/constructors/providers.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def __call__(self, *args, **kwargs):
+        """Build and return the product with attribute dependencies injected."""
+        # TODO: Split this into two steps, build chain of sub-providers
+        # and call the providers.
+        _constructor: Constructor
+
+        if args or kwargs:
+            _constructor = partial(self.constructor, *args, **kwargs)
+        else:
```
YooSunYoung, line 161: The reason why I tried to inject the arguments/attributes here was because I wanted to use ``Provider`` as a function wrapper, so that it can just be called with the arguments as well, but...

Now I think it's not a very useful interface since it has the function call as a member variable anyways.

So this will be moved to ``Binder`` __getitem__ instead and the Provider class will have much less responsibility.

YooSunYoung:
[How it will be used in applications](https://github.com/scipp/beamlime/blob/blueprint/src/beamlime/applications/daemons.py)
[How to run daemons in jupyter notebook](https://github.com/scipp/beamlime/blob/blueprint/docs/blueprint.ipynb)
[How to run daemons (without plotting)](https://github.com/scipp/beamlime/blob/blueprint/docs/run_example.py)

[Rough Ideas including examples above](https://github.com/scipp/beamlime/pull/68)

src/beamlime/constructors/providers.py
YooSunYoung: This one is updated in the next commits.

YooSunYoung:
Here is much simpler example that can be run only with this changes:

```
from beamlime.constructors import Factory
factory = Factory()
shared_list = []
factory.register(list[int], lambda: shared_list)

@factory.provider
class DataStreamStart:
    raw_data: list[int]
    def run(self):
        for count in range(1, 11):
            self.raw_data.append(count)

@factory.provider
class DataStreamEnd:
    raw_data: list[int]
    def run(self):
        for data in self.raw_data:
            print("Count: ", data)

@factory.provider
class MainDaemon:
    app_start: DataStreamStart
    app_end: DataStreamEnd

    def run(self):
        self.app_start.run()
        self.app_end.run()

factory[MainDaemon].run()

```
---------------------------
```
Count:  1
Count:  2
Count:  3
Count:  4
Count:  5
Count:  6
Count:  7
Count:  8
Count:  9
Count:  10
```

src/beamlime/constructors/injection.py
jl-wynen, line 1: The injection framework uses a relatively deep inheritance hierarchy. Is there a way to make it smaller?
E.g., can the various interfaces be implemented directly in `FactoryBase`?

src/beamlime/constructors/injection.py
```diff
@@ -0,0 +1,171 @@
...
+    """Provider registration interfaces."""
+
+    def register(self, product_type: ProductType, provider_call: Constructor) -> None:
+        """
+        Register a provider function ``provider_call`` of ``product_type``
+        into the ``self._providers`` dictionary.
+
+        If ``product_type`` is ``UnknownType``,
+        it will retrieve the product type from the provider call
+        and use it as an index.
```
jl-wynen, line 29: Can you expand the docs to explain when it is possible to retrieve the type and how to use this? An example may help.

src/beamlime/constructors/injection.py
```diff
@@ -0,0 +1,171 @@
...
+        self._validate_new_provider(_product_type, new_provider_call)
+        self._providers[_product_type] = new_provider_call
+
+    def _validate_new_provider(
+        self, product_type: ProductType, new_provider: Provider
+    ) -> None:
+        """
+        Raises
+        ------
+        ProviderExistsError:
```
jl-wynen, line 52: Remove colon: https://scipp.github.io/reference/developer/coding-conventions.html?highlight=convention#docstrings

src/beamlime/constructors/injection.py
```diff
@@ -0,0 +1,171 @@
...
+            If there is an existing provider of the ``product_type``.
+            However, if the full name of the ``new_provider_call``
```
jl-wynen, line 54: The function doesn't use the name. Is this comment correct?

src/beamlime/constructors/injection.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def build_attr_dependency(self, dep_type: ProductType):
+        try:
+            return self.assemble(dep_type)
+        except ProviderNotFoundError:
+            return Empty
+
+    def inject_dependencies(self, _obj: _Product) -> _Product:
+        """
+        Check if the ``incomplete_obj`` has attribute dependencies to be filled.
```
jl-wynen, line 124: What is `incomplete_obj`? Did you rename the arg and forget to update the comment?

src/beamlime/constructors/injection.py
```diff
@@ -0,0 +1,171 @@
...
+                # by checking the missing attributes above.
+            else:
+                setattr(_obj, _attr_name, _attr)
+        return _obj
+
+    def assemble(self, product_type: Type[_Product]) -> _Product:
+        _provider = self.find_provider(product_type)
+        kwargs = {**self.build_arguments(_provider), **_provider.keywords}
+        _obj = _provider.constructor(*_provider.args, **kwargs)
+        return self.inject_dependencies(_obj)
```
jl-wynen, line 171: I'm confused. It looks like after `_obj = _provider.constructor(*_provider.args, **kwargs)`, `_obj` is a `Constructor`. Then, `self.inject_dependencies(_obj)` sets a number of attributes on that `Constructor` but never calls it.
So when does it make a `_Product`?

src/beamlime/constructors/providers.py
```diff
@@ -0,0 +1,155 @@
...
+
+    """
+
+    _constructor: Constructor
+    args: tuple
+    keywords: dict
+    product_spec: ProductSpec
+    arg_dep_specs: Dict[str, DependencySpec]
+
+    def __new__(
```
jl-wynen, line 48: Why did you need to use `__new__` here? This looks like it can be done in `__init__`.

jl-wynen (commented):
Comments in `base.py` are mostly low level. Please start with the file level comments and in particular in `injection.py` because they are about the larger scale design.

src/beamlime/constructors/injection.py
YooSunYoung: So the ``constructor`` is a property so its return type is ``Constructor`` which is callable, and the return type of the ``constructor`` call is ``_Product``.
I should have called the provider directly instead of calling constructor ... and use type hint in the __call__.
I'm on it thank you...!

YooSunYoung: It is now updated. It might still be confusing that it doesn't allow position-only arguments.
The idea was to annotate all the arguments as non-position-only, but then we might not be able to use external functions from other libraries with this injection framework.
But I don't expect that to happen often anyways so I would rather keep it this way.

src/beamlime/constructors/base.py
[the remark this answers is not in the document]
YooSunYoung: It was just my own preference due to unnecessary concerns what if there is a global variable with the exact same name that are set by other modules... which will not be the case anyways...
Another reason is that I added `_` or `__` or nothing to distinguish different scopes if there are nested scopes within one function.

And often I forgot to remove them after removing or merging those scopes,
so I will read through all files and renamed variables to have normal names...!
Thank you...!

src/beamlime/constructors/base.py
```diff
@@ -0,0 +1,108 @@
...
+    ...
+
+
+def _is_conflicting(tp: ProductType, factories: Tuple[FactoryBase, ...]):
+    _providers = [factory.providers[tp] for factory in factories if tp in factory]
+    if _providers:
+        _standard = _providers.pop()
+        if any([_provider for _provider in _providers if _provider != _standard]):
+            return True
+    return False
```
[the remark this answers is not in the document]
YooSunYoung: No, it is using it's own ``__eq__``, comparing ``provider.constructor`` not the ``provider`` instances itself.
It should actually have better comparison logic in ``__eq__`` though...

src/beamlime/constructors/contexts.py
```diff
@@ -0,0 +1,65 @@
...
+
+        original_provider = self.pop(product_type)
+
+        try:
+            self.register(product_type, temp_provider)
+            yield None
+        finally:
+            self.pop(product_type)
+            if original_provider is not UnknownProvider:
+                self._providers[product_type] = original_provider
```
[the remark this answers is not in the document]
YooSunYoung: Because `register` re-wraps the provider, parses the function annotation again and checks the validity of the new ``Provider`` object, which is unnecessary in this context,
since the original provider must have been registered by either `provider` decorator or the `register` function.
but maybe it is not that safe... let me think...

src/beamlime/constructors/base.py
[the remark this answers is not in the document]
YooSunYoung: Now ``Provider`` has ``__hash__`` so it is using ``set``...!

src/beamlime/constructors/base.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def check_conflicting_providers(*factories: FactoryBase):
+    from functools import reduce
+
+    tp_set_list = [set(tp for tp in factory) for factory in factories]
+    tp_set = reduce(lambda x, y: x.union(y), tp_set_list)
+
+    # If there is any overlapping providers
+    if sum([len(tp_set) for tp_set in tp_set_list]) > len(tp_set):
```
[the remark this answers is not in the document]
YooSunYoung: Now it is not using this logic and using ``set`` instead..!

src/beamlime/constructors/inspectors.py
```diff
@@ -0,0 +1,187 @@
...
+        if isinstance(product_type, ProductSpec):
+            ...
+        else:
+            validate_annotation(product_type)
+            self.product_type = product_type
+            self.returned_type = extract_underlying_product_type(product_type)
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, ProductSpec):
+            raise NotImplementedError(
```
jl-wynen, line 71: Why raise an exception? Typically, Python objects return `False` on type mismatch.

src/beamlime/constructors/inspectors.py
```diff
@@ -0,0 +1,187 @@
...
+class InsufficientAnnotationError(Exception):
+    ...
+
+
+class ProductNotFoundError(Exception):
+    ...
+
+
+class UnknownType:
+    ...
```
jl-wynen, line 23: Are those part of the public API? If so please import them in `__init__`.

src/beamlime/constructors/providers.py
```diff
@@ -0,0 +1,382 @@
...
+    to see if the ``callable_obj`` is a method.
+    """
+    if (
+        not isinstance(callable_obj, partial)
+        and not isinstance(callable_obj, type)
+        and hasattr(callable_obj, "__name__")
+        and not (callable_obj.__name__ == _lambda_name)
+        and hasattr(callable_obj, "__qualname__")
+        and callable_obj.__qualname__ != callable_obj.__name__
+    ):
```
jl-wynen, line 75: Does this not work? https://docs.python.org/3/library/inspect.html#inspect.ismethod

src/beamlime/constructors/providers.py
```diff
@@ -0,0 +1,382 @@
...
+        else:
+            self.args = args
+            self.keywords = kwargs
+
+    def _init_dependencies(self) -> None:
+        self.arg_dep_specs: DependencySpecDict = collect_argument_specs(
+            self._constructor, *self.args, **self.keywords
+        )
+        self.attr_dep_specs: DependencySpecDict = collect_attr_specs(self._constructor)
+        self.product_spec: ProductSpec = get_product_spec(self._constructor)
```
jl-wynen, line 137: This method defines instance attributes that are not listed in `__init__`.

src/beamlime/constructors/providers.py
```diff
@@ -0,0 +1,382 @@
...
+            return self.can_provide(orig_tp)
+        elif isinstance(provided_tp, type) and isinstance(requested_tp, type):
+            # If the requested type is a parent class of the provided type.
+            return issubclass(provided_tp, requested_tp)
+        else:
+            return False
+
+    def __call__(self, *args, **kwargs) -> Product:
+        """Call the constructor with extra arguments."""
+        return self.constructor(*(*self.args, *args), **{**self.keywords, **kwargs})
```
jl-wynen, line 175:
```suggestion
        return self.constructor(*self.args, *args, **self.keywords, **kwargs)
```

src/beamlime/constructors/providers.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        """Compare ``constructor``, ``args`` and ``keywords`` of the provider."""
+        if not isinstance(other, Provider):
+            raise NotImplementedError(
+                "Comparison between Provider" " with other type is not supported."
```
jl-wynen, line 181:
```suggestion
                "Comparison between Provider with other type is not supported."
```

src/beamlime/constructors/providers.py
```diff
@@ -0,0 +1,382 @@
...
+            return False
+
+    def __call__(self, *args, **kwargs) -> Product:
+        """Call the constructor with extra arguments."""
+        return self.constructor(*(*self.args, *args), **{**self.keywords, **kwargs})
+
+    def __eq__(self, other: object) -> bool:
+        """Compare ``constructor``, ``args`` and ``keywords`` of the provider."""
+        if not isinstance(other, Provider):
+            raise NotImplementedError(
```
jl-wynen, line 180: Same as before, why not return `False`?

jl-wynen (commented):
Haven't looked at the tests yet

src/beamlime/constructors/factories.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        Build and inject attribute dependencies of the ``product``.
+        If ``product`` is missing any attributes that are type-hinted,
+        the attribute(dependency) will be assembled and injected.
+
+        Raises
+        ------
+        ProviderNotFoundError
+            If any type-hinted attributes are not populated in ``product`` and
+            any providers for the missing attributes are not found.
```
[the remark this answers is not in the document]
YooSunYoung: Now the ``self.providers`` is ``ProviderGroup`` and it raises ``ProviderNotFoundError`` instead of ``KeyError`` from the ``__getitem__``. I'll add type-hint in the init as well.

src/beamlime/constructors/factories.py
```diff
@@ -0,0 +1,185 @@
...
+        When the ``product`` already has the attribute but it is ``None``,
+            1. if a provider is not found, it skips injection.
+            2. if a provider is found, it overwrites the attribute.
+
+        """
+        provider = self.providers[product_type]
+        attr_dependencies: Dict[str, Any] = {
+            attr_name: attr_spec.dependency_type
+            for attr_name, attr_spec in provider.attr_dep_specs.items()
+            if not hasattr(product, attr_name) or getattr(product, attr_name) is None
```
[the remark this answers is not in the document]
YooSunYoung: For the case it doesn't have attribtue at all, I will replace it to ``if getattr(product, attr_name, None) is None`` thank you!

src/beamlime/constructors/factories.py
```diff
@@ -0,0 +1,185 @@
...
+        from .providers import merge
+
+        merged = merge(*provider_groups)
+        my_providers = copy(self.providers)
+        for product_type in merged:
+            my_providers.pop(product_type)
+
+        try:
+            yield Factory(my_providers, merged)
+        finally:
```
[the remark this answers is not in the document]
YooSunYoung: Yes...! I thought about adding other checks or clean up but I don't think we need any. I'll remove them.

src/beamlime/constructors/inspectors.py
YooSunYoung: Since `ProductSpec` is intended to be used only within the `Provider`,
I wanted to prevent unexpected usage of it as much as possible.
Will it cause any problems later...?

src/beamlime/constructors/providers.py
YooSunYoung: Unfortunately not... it only works with the method of an instance.
And the decorator `ProviderGroup.provider` will not be called when it's instantiated,
but when it's imported,
which means `callable_obj` will be `C.foo` instead of `c.foo` in the example below.

```
provider_gr = ProviderGroup()
class C:
  @provider_gr.provider
  def foo(self):
    ...

c = C()
```

Another way is to check the type of the member of the class by `inspect.classify_class_attrs`,
and then you need to have an access to the class,
which is not possible only from the callable object.

I will maybe add a TODO tag at least to get back to this problem later
since it's not really an explicit way to check if the callable is a method....

tests/constructors/preset_factory.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def make_a_joke(self) -> Joke:
+        return self.joke
+
+    def how_are_you(self) -> Status:
+        return self.status
+
+
+test_provider_group = ProviderGroup(Adult, Parent, give_a_good_telling)
+test_factory = Factory(test_provider_group)
```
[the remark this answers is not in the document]
YooSunYoung: Oh... I didn't know that was this simple....
but it is good that I didn't know that otherwise I wouldn't have written those context managers earlier which is quite useful : D........
I turned them to fixtures thank you...!

And apparently, if you have a nested fixture, it doesn't allow you to call another fixture function directly... it took a little bit of time to figure out...

src/beamlime/constructors/inspectors.py
jl-wynen: > Will it cause any problems later...?

Only really if it is used together with other types. E.g. a function that takes a `ProductSpec` or, say, a function as argument and does an equality check without checking the type.
I doubt that this will be an actual problem. It is just that, AFAIK, the canonical behaviour is to return `False` on type mismatch.

YooSunYoung: Yes, I actually want to strictly forbid it to be used with other types.
I would rather keep it like this and maybe we can change it to return `False` when we want to use it in other places as well.

src/beamlime/constructors/providers.py
YooSunYoung: It's because the input of ``__setitem__`` is different from the output of ``__getitem__``.
It's because it wraps the item with `Provider` class.
And user might want to do something like below.
So I thought it will be confusing if it just returns False without any explanation...?

```
provider_gr = ProviderGroup()
def foo() -> int:
  return 0

provider_gr[int] = foo
provider_gr[int] == foo  # raise NotImlementedError
```

jl-wynen (approved):

PR #60, opened 2023-06-18
Logging methods refurbished!

YooSunYoung:
Custom LogRecord was a bad idea,
so now LogMixin composes message with its class name
instead of pass the application name as an argument in the logging methods.

File handler can be configured by dependency injection.
Unit-tests have examples how to configure custom directory path or log file prefix/extension, but customising logging is not quite expected.


YooSunYoung:
> This includes the dependency injection code. If you want this PR to be reviewed and merged, can you retarget it to the dependency injection branch?

I opened the new one #63 
I'll just keep this open since it'll be merged into main at the end.

PR #70, opened 2023-07-10
Refurbished logging module.

YooSunYoung:


tests/logging/contexts.py
```diff
@@ -0,0 +1,35 @@
...
+    within the sub context using ``log_factory``.
+
+    It will help a test not to interfere other tests.
+    """
+
+    original_logger_names = list(logging.Logger.manager.loggerDict.keys())
+    try:
+        with log_factory.local_factory(*local_log_providers) as local_log_factory:
+            yield local_log_factory
+    finally:
```
YooSunYoung, line 28: It cleans up any loggers that were added to the `logging.Logger.manager` within this context.
It was easier than resetting handlers of each loggers. It also keeps original loggers that are unrelated to the tests as they were.

src/beamlime/logging/resources.py
```diff
@@ -0,0 +1,118 @@
...
+    try:
+        yield _lock
+    finally:
+        _lock.release() if _lock else ...
+
+
+def cleanup_file_handlers(logger: logging.Logger):
+    """Find file handlers that are connectected to
+    non-existing files and remove them from the logger."""
+    from os.path import exists
```
YooSunYoung, line 110: It probably should raise an error instead of just removing the handlers...

src/beamlime/logging/resources.py
```diff
@@ -0,0 +1,119 @@
...
+def create_time_stamp() -> TimeStamp:
+    """
+    Creates a timestamp with the format, ``{cur_timestamp}--{h_time}``.
+    ``htime`` is as an extra information only for the human users,
+    and the ``cur_timestamp`` should be used by the applications or modules.
+    """
+    from datetime import datetime
+
+    cur_timestamp = round(time.time())
+    h_time = datetime.fromtimestamp(cur_timestamp).strftime("%Y-%m-%d-%H-%M-%S")
```
jl-wynen, line 27: Can you change this to use the ISO format?

```suggestion
    h_time = datetime.fromtimestamp(cur_timestamp).isoformat()
```

And ideally even with timezone information. I'm weary of times without timezone information as they can easily be misinterpreted.
You could use the local timezone if you are willing to use a package like dateutil. But I think it is enough to show everything in UTC using:

```suggestion
    from datetime import datetime, timezone

    now = datetime.now(tz=timezone.utc)
    cur_timestamp = round(now.timestamp())
    h_time = now.isoformat()
```

But in general, why do we need the timestamp? If we use the full ISO format, it can be parsed back into a `datetime`, esp. if we force it to be in UTC.

src/beamlime/logging/resources.py
```diff
@@ -0,0 +1,119 @@
...
+    return TimeStamp(f"{cur_timestamp}--{h_time}")
+
+
+LogFilePrefix = NewType("LogFilePrefix", str)
+DefaultPrefix = LogFilePrefix("beamlime")
+
+LogFileExtension = NewType("LogFileExtension", str)
+DefaultFileExtension = LogFileExtension("log")
+
+LogDirectoryPath = NewType("LogDirectoryPath", str)
```
jl-wynen, line 37: Why not use `Path` instead of `str` here? Then you can even use methods on it instead of the functions in `os`.

src/beamlime/logging/resources.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+@log_providers.provider
+def create_log_file_name(
+    time_stamp: TimeStamp,
+    prefix: LogFilePrefix = DefaultPrefix,
+    suffix: LogFileExtension = DefaultFileExtension,
+) -> LogFileName:
+    """Combine prefix, timestamp and suffix into a new file name."""
+    return LogFileName(f"{prefix}--{time_stamp}.{suffix}")
```
jl-wynen, line 70: Did you check if this works on Windows? I don't know if it supports numbers and dashes in filenames like this.

src/beamlime/logging/resources.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+FileHandlerBasePath = NewType("FileHandlerBasePath", Path)
+DefaultDirectoryStatus = DirectoryCreated(True)
+
+
+@log_providers.provider
+def create_log_file_path(
+    directory_ready: DirectoryCreated = DefaultDirectoryStatus,
```
jl-wynen, line 79: Why is there a default argument?

src/beamlime/logging/resources.py
```diff
@@ -0,0 +1,119 @@
...
+
+@contextmanager
+def hold_logging() -> Iterator[RLock]:
+    from . import _lock
+
+    _lock.acquire() if _lock else ...
+    try:
+        yield _lock
+    finally:
+        _lock.release() if _lock else ...
```
jl-wynen, line 104: What are the `if _lock` checks for? And why does this function yield the lock?

src/beamlime/logging/resources.py
YooSunYoung: That is very true... I thought it'll be easier to keep it as `str` in case it'll be user input later but I can just wrap the user input with `Path` so it doesn't matter. I'll update it...!

src/beamlime/logging/resources.py
YooSunYoung: I just tried to keep it consistent with what `logging` library does. But it's no longer used in the `beamlime` anymore so I'll just remove it...! Thank you!

src/beamlime/logging/resources.py
YooSunYoung: Actually no... I removed the default argument. Thank you!

src/beamlime/logging/resources.py
YooSunYoung: Actually no... for now the file path doesn't have `--` anymore but I should probably add `Windows` into CI tests.
I opened an issue #72

src/beamlime/logging/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    BeamlimeLogger,
+    FileHandlerConfigured,
+    LogLevels,
+    get_logger,
+    get_scipp_logger,
+    initialize_file_handler,
+)
+from .resources import LogDirectoryPath, LogFileExtension, LogFileName, LogFilePrefix
+
+_lock = threading.RLock()
```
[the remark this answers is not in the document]
YooSunYoung: Right... I forgot that it doesn't check  `__init__.py` for unused variables...

jl-wynen (approved):

PR #71, opened 2023-07-10
Async scheduling/error handling helper decorator.

YooSunYoung:
It is manly to wait for the new data coming into the data pipeline without blocking other asynchronous calls.
However, it does not allow accurate timeout ability.
It will wait for longer period than expected, depending on other async calls that are used while waiting.

### DIFF
--- a/tests/core/schedulers_test.py
+++ b/tests/core/schedulers_test.py
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+import pytest
+
+
+@pytest.mark.parametrize(
+    ["max_trials", "success"],
+    [(4, True), (5, False)],
+)
+def test_retry_decorator_max_trial(max_trials: int, success: bool):
+    import time
+
+    from beamlime.core.schedulers import retry
+
+    interval = 0.2
+    error_throw_ranges = iter(range(max_trials - int(success) * 1))
+    context = {"run_counts": 0}
+
+    @retry(TimeoutError, max_trials=max_trials, interval=interval)
+    def something_takes_time():
+        context["run_counts"] += 1
+        for _ in error_throw_ranges:
+            raise TimeoutError
+        return True
+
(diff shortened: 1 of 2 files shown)


YooSunYoung:
Here is the example of how it's used in other modules.
[data pipe buffer](https://github.com/scipp/beamlime/pull/74/files#diff-d9743ba727a2f50ba62394720c1ff5264f79926eeff74e9a71abb8383cfac903)

tests/core/schedulers_test.py
```diff
@@ -0,0 +1,75 @@
...
+            something_takes_time()
+    else:
+        assert something_takes_time()
+
+    time_consumed = time.time() - start_time
+    expected_consumed_time = (max_trials - 1) * interval
+    assert time_consumed > expected_consumed_time
+    assert (
+        time_consumed - expected_consumed_time
+    ) < 0.01  # Expected delta for several trials.
```
SimonHeybrock, line 38: If possible, I'd make this larger, 10 ms may make this test flaky.

tests/core/schedulers_test.py
```diff
@@ -0,0 +1,75 @@
...
+    else:
+        assert asyncio.run(something_takes_time())
+
+    assert context["run_counts"] == max_trials
+    time_consumed = time.time() - start_time
+    expected_consumed_time = (max_trials - 1) * interval
+    assert time_consumed > expected_consumed_time
+    assert (
+        time_consumed - expected_consumed_time
+    ) < 0.01  # Expected delta for several trials.
```
SimonHeybrock, line 75: See above.

SimonHeybrock (commented):
Looks reasonable, cannot say much about the larger context.

tests/core/schedulers_test.py
YooSunYoung: Yes... it was very close to the actual difference.
I increased it to `0.05` instead of `0.01`.
I'll add a brief benchmark in the documentation as well. #76

SimonHeybrock (approved):

PR #73, opened 2023-07-11
Cached provider

YooSunYoung:
The previous implementation, delayed registration no longer works since `ProviderGroup` is separated from `Factory`.
Therefore here is the new implementation.

The main purpose of the cached provider is to share a data-pipe instance between applications.

## ProviderGroup.cached_provider
`ProviderGroup` has `cached_provider` method to register a cached provider call.
It will wrap the `Callable` with `CachedProvider` instead of `Provider`, which is a subclass of `Provider`.

## CachedProvider
It is a subclass of `Provider`, which is a function wrapper similar to `partial`.
It wraps the `Callable` with `functools.lru_cache(maxsize=1)`.
The `__call__` will check if it was called with different arguments before and will not allow new ones.

## Provider.__deepcopy__
In order to keep the cache only available within the `Factory` or `ProviderGroup`,
`Provider` has `__deepcopy__` which will create a new instance of the `Provider` or a sub class if not overwritten.
`ProviderGroup` will make a deep copy of a provider dictionary when copied or merged.

## Type Hints
It also has more type hints but `mypy` with `--strict` option still doesn't pass for couple of cases.

### DIFF
--- a/src/beamlime/constructors/inspectors.py
+++ b/src/beamlime/constructors/inspectors.py
@@ -41,12 +41,12 @@ _NewT = NewType("_NewT", int)
 
 
 class NewTypeMeta(type):
-    def __instancecheck__(cls, __instance: Any) -> bool:
+    def __instancecheck__(cls, instance: Any) -> bool:
         return (
-            callable(__instance)
-            and hasattr(__instance, "__supertype__")
-            and _NewT.__module__ == __instance.__module__
-            and _NewT.__qualname__ == __instance.__qualname__
+            callable(instance)
+            and hasattr(instance, "__supertype__")
+            and _NewT.__module__ == instance.__module__
+            and _NewT.__qualname__ == instance.__qualname__
         )
 
 
--- a/src/beamlime/constructors/inspectors.py
+++ b/src/beamlime/constructors/inspectors.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Callable, Dict, Literal, Type, TypeVar, Union
+from typing import Any, Callable, Dict, Generic, Literal, NewType, Type, TypeVar, Union
 
 
 class Empty:
@@ -23,7 +23,7 @@ class UnknownType:
     ...
 
 
-def validate_annotation(annotation) -> Literal[True]:
+def validate_annotation(annotation: Any) -> Literal[True]:
     """
     Check if the origin of the annotation is not Union.
 
@@ -37,11 +37,25 @@ def validate_annotation(annotation) -> Literal[True]:
 
 
 Product = TypeVar("Product")
-ProductType = Type[Product]
+_NewT = NewType("_NewT", int)
 
 
-def extract_underlying_product_type(product_type: ProductType) -> ProductType:
-    if hasattr(product_type, "__supertype__"):  # NewType
+class NewTypeMeta(type):
+    def __instancecheck__(cls, __instance: Any) -> bool:
+        return (
+            callable(__instance)
+            and hasattr(__instance, "__supertype__")
+            and _NewT.__module__ == __instance.__module__
+            and _NewT.__qualname__ == __instance.__qualname__
+        )
+
+
+class NewTypeProtocol(Generic[Product], metaclass=NewTypeMeta):
+    __supertype__: Type[Product]
+
+
+def extract_underlying_product_type(product_type: Type[Product]) -> Type[Product]:
+    if isinstance(product_type, NewTypeProtocol):
         return product_type.__supertype__
     else:
         return product_type
@@ -52,9 +66,9 @@ class ProductSpec:
     Specification of a product (returned value) of a provider.
     """
 
-    def __init__(self, product_type: Union[ProductType, ProductSpec]) -> None:
-        self.product_type: ProductType
-        self.returned_type: ProductType
+    def __init__(self, product_type: Union[Type[Product], ProductSpec]) -> None:
+        self.product_type: Type[Product]
+        self.returned_type: Type[Product]
 
         if isinstance(product_type, ProductSpec):
             self.product_type = product_type.product_type
@@ -73,19 +87,16 @@ class ProductSpec:
             return self.product_type == other.product_type
 
 
-def collect_arg_typehints(callable_obj: Callable) -> Dict[str, Any]:
+def collect_arg_typehints(callable_obj: Callable[..., Product]) -> Dict[str, Type[Any]]:
     from typing import get_type_hints
 
     if isinstance(callable_obj, type):
-        # TODO: When the callable_obj is a class,
-        # investigating ``__init__`` of the class should be okay
-        # but may replace this to the better solution and remove type check ignore tag.
-        return get_type_hints(callable_obj.__init__)  # type: ignore[misc]
+        return get_type_hints(callable_obj.__init__)
     else:
         return get_type_hints(callable_obj)
 
 
-def get_product_spec(callable_obj: Callable) -> ProductSpec:
+def get_product_spec(callable_obj: Callable[..., Product]) -> ProductSpec:
     """
     Retrieve the product of provider.
     If ``callable_obj`` is a function, it is a return type annotation,
@@ -98,7 +109,7 @@ def get_product_spec(callable_obj: Callable) -> ProductSpec:
         return ProductSpec(product)
 
 
-def extract_underlying_dep_type(tp: Any) -> ProductType:
+def extract_underlying_dep_type(tp: Any) -> Any:
     from typing import Union, get_args, get_origin
 
     if get_origin(tp) == Union:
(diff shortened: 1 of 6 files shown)


src/beamlime/constructors/inspectors.py
```diff
@@ -37,11 +37,25 @@ def validate_annotation(annotation) -> Literal[True]:
...
 
 Product = TypeVar("Product")
-ProductType = Type[Product]
+_NewT = NewType("_NewT", int)
 
 
-def extract_underlying_product_type(product_type: ProductType) -> ProductType:
-    if hasattr(product_type, "__supertype__"):  # NewType
+class NewTypeMeta(type):
+    def __instancecheck__(cls, __instance: Any) -> bool:
```
YooSunYoung, line 44: During the runtime, `NewType` returns a dynamically created function, that is same as `lambda x: x`,
and typing module doesn't provide type check of `NewType` object as far as I know.

src/beamlime/constructors/inspectors.py
```diff
@@ -37,11 +37,25 @@ def validate_annotation(annotation) -> Literal[True]:
...
+        return (
+            callable(__instance)
+            and hasattr(__instance, "__supertype__")
+            and _NewT.__module__ == __instance.__module__
+            and _NewT.__qualname__ == __instance.__qualname__
+        )
+
+
+class NewTypeProtocol(Generic[Product], metaclass=NewTypeMeta):
+    __supertype__: Type[Product]
```
jl-wynen, line 54: I don't understand this. Do product types now have to be built using `NewTypeProtocol`? I don't see this in the tests.

Why does `__instancecheck__` check whther `__instance` is callable?

Also, `__instance` shouldn't have any underscores.

src/beamlime/constructors/providers.py
```diff
@@ -199,12 +203,54 @@ def __hash__(self) -> int:
...
         return f"Provider({self.call_name}, *{self.args}, **{self.keywords})."
 
+    @classmethod
+    def __copy__(cls, _obj: Provider[Product]) -> Provider[Product]:
+        return cls(_obj)
+
+    def __deepcopy__(self, _: dict[Any, Any]) -> Provider[Product]:
+        from copy import copy
+
+        return copy(self)
```
jl-wynen, line 213: This doesn't look like a deep copy. It needs to deepcopy all members. Doesn't the default deepcopy work here?

src/beamlime/constructors/providers.py
```diff
@@ -199,12 +203,54 @@ def __hash__(self) -> int:
...
+
+
+class CachedProvider(Provider[Product]):
+    def __init__(
+        self, _constructor: Constructor[Product], /, *args: Any, **kwargs: Any
+    ) -> None:
+        from functools import lru_cache
+
+        super().__init__(_constructor, *args, **kwargs)
+        self.cached_constructor = lru_cache(maxsize=1)(self._constructor)
```
jl-wynen, line 234: This creates reference loops because `self` refers to the cache and the cache refers to `self` via the bound first argument of `self._constructor` .

I am not aware of a good solution here. You may have to implement caching yourself. Since you already check the arguments in the call, this shouldn't be too difficult.

src/beamlime/constructors/providers.py
```diff
@@ -199,12 +203,54 @@ def __hash__(self) -> int:
...
+        self.cache_indicator = lru_cache(maxsize=2)(_dummy_call)
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Product:
+        self.cache_indicator(*self.args, *args, **self.keywords, **kwargs)
+        if self.cache_indicator.cache_info().currsize > 1:
+            err_msg = (
+                f"CachedProvider {self} was called with "
+                "different arguments from the last call."
+            )
+            raise CachedProviderCalledWithDifferentArgs(err_msg)
```
jl-wynen, line 244: If this function is called 3 times with argument sets
1. A
2. B
3. A

That is, the 2nd time with different arguments than the 1st and the 3rd time with the same as the 1st. Does the 3rd call work or also raise an exception?

tests/constructors/provider_group_test.py
jl-wynen, line 1: I don't see any tests for the case of mismatched arguments.

src/beamlime/constructors/inspectors.py
YooSunYoung: All `class`es are callable and `Union` also returns a `function` object so it's no use to check if it's `Callable` or a `function`.

+ I removed underscores...!

jl-wynen: If it's for static type checking, please consider making it a type guard: https://docs.python.org/3/library/typing.html#typing.TypeGuard

src/beamlime/constructors/providers.py
YooSunYoung: It was mainly for making a deep copy of `_providers` dictionary of `ProviderGroup`.

I dug up the `deepcopy` and when you use `deepcopy` of a dictionary,
if the each value doesn't have a specific implementation of `__deepcopy__`,
it simply uses the same instance, not even `copy` of it...
```
from copy import deepcopy
provider = Provider(lambda: 0)
providers = {int: provider}
deepcopy(providers)  # means {deepcopy(int): deepcopy(provider)}
```

I could do something like
```
from copy import copy
new_providers = {key: copy(value) for key, value in providers.items()}
```
Rather than actually makes deep copy of every member of `Provider`....?

jl-wynen: Then I would call this `copy`, not `deepcopy`. If this is all you need, go ahead!

src/beamlime/constructors/inspectors.py
```diff
@@ -52,9 +66,9 @@ class ProductSpec:
     Specification of a product (returned value) of a provider.
     """
 
-    def __init__(self, product_type: Union[ProductType, ProductSpec]) -> None:
-        self.product_type: ProductType
-        self.returned_type: ProductType
+    def __init__(self, product_type: Union[Type[Product], ProductSpec]) -> None:
+        self.product_type: Type[Product]
+        self.returned_type: Type[Product]
```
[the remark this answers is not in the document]
YooSunYoung: I wanted to keep them bound in the `ProviderGroup.__getitem__`.
And then whenever I annotate `product_type` with `type[any]`,
static type checker complains that it's not `type[Product]`...
And since the `returned_type` is the type of what is actually returned from a callable,
it is should always be a super class of `product_type` so I think this annotation makes sense...?
```
class NewInt(int): ...
product = NewInt(1)
```
With the case above,
```
isinstance(product, int)
isinstance(product, NewInt)
```

`returned_type` will be `int` and `product_type` will be `NewInt`,
except that it'll be created by `NewType` instead of inheriting `int`...?

And if it's not `NewType`, they will be exact same thing.

src/beamlime/constructors/inspectors.py
YooSunYoung: That is so cool....! But it's only available from `py3.10`. I'll open an issue to use it. But still we need `NewTypeProtocol` at least I guess...? To do something like `is_newtype(val) -> TypeGuard[NewTypeProtocol]: ...`.

src/beamlime/constructors/providers.py
YooSunYoung: I tried to implement caching myself but hashing `args` and `kwargs` is complicated when you don't know what they will be so I thought it'll be easier to just use `lru_cache`.
And... I don't see the reference loop though, `lru_cache` wraps the `self._constructor` and the wrapper will have the cache.

tests/constructors/provider_group_test.py
YooSunYoung: Yes..! I added them now. Thank you!

YooSunYoung: But they are in the `provider_test.py` not here...!

src/beamlime/constructors/providers.py
YooSunYoung: I just updated it to reset the cache...!

src/beamlime/constructors/providers.py
YooSunYoung: I did...! Thank you!

src/beamlime/constructors/providers.py
```diff
@@ -233,13 +233,17 @@ def __init__(
...
+        self.cached_result: Product
         self.cached_args = CachedArguments()
-        self.cached_constructor = lru_cache(maxsize=1)(self._constructor)
         self.cache_indicator = lru_cache(maxsize=2)(self.cached_args)
 
     def __call__(self, *args: Any, **kwargs: Any) -> Product:
         self.cache_indicator(*self.args, *args, **self.keywords, **kwargs)
-        if self.cache_indicator.cache_info().currsize > 1:
+        if not hasattr(self, "cached_result"):
+            self.cached_result = self.constructor(
```
jl-wynen, line 243: You changed it from `self._constructor` to `self.constructor` without any changes elsewhere. Which one is correct?

YooSunYoung: `self.constructor` is the actual exposed property, so I thought it might be more reasonable to use that one here.
`Provider` doesn't do anything but just gives the `self._constructor` when `self.constructor` is called.

jl-wynen (approved):

PR #74, opened 2023-07-12
Async/sync data sharing buffer.

YooSunYoung:
[Examples of usage](https://github.com/scipp/beamlime/pull/68/files#diff-2ff91e957c5c825368667bc0e24602c3efa7ade43b59a6cdf1ca6532b0b8a5fe)

`Pipe` is a generic type so that it can have a type argument per pipe instance.
It is intended to have only one instance of `Pipe` per dependency injection `Factory`.


YooSunYoung:
I reopened the PR due to remove the commits that are already merged. #81

PR #81, opened 2023-07-12
Async/sync data sharing buffer.

YooSunYoung:
> Same PR as #74 but github for some reason shows unrelated changes from commits that are already merged...

[Examples of usage](https://github.com/scipp/beamlime/pull/68/files#diff-2ff91e957c5c825368667bc0e24602c3efa7ade43b59a6cdf1ca6532b0b8a5fe)

`Pipe` is a generic type so that it can have a type argument per pipe instance.
It is intended to have only one instance of `Pipe` per dependency injection `Factory`.

### DIFF
--- a/src/beamlime/communication/pipes.py
+++ b/src/beamlime/communication/pipes.py
@@ -1,0 +1,238 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+
+from __future__ import annotations
+
+from contextlib import asynccontextmanager, contextmanager
+from queue import Empty, Full
+from typing import (
+    AsyncGenerator,
+    AsyncIterator,
+    Generator,
+    Generic,
+    Iterator,
+    List,
+    Type,
+    TypeVar,
+    Union,
+)
+
+BufferData = TypeVar("BufferData")
+
+
+MAX_CHUNK_SIZE = 100
+DEFAULT_CHUNK_SIZE = int(MAX_CHUNK_SIZE / 10)
+MAX_BUFFER_SIZE = MAX_CHUNK_SIZE * 10
+
+
+class BufferBase(Generic[BufferData]):
+    """
+    Buffer base class that carries a single chunk of data.
+    """
+
+    def __init__(
+        self,
+        *_initial_data: BufferData,
+        max_size: int = DEFAULT_CHUNK_SIZE,
+    ):
+        self._data: List[BufferData] = list(_initial_data)
+        self.max_size: int = max_size
+
+    @property
+    def max_size(self) -> int:
+        return self._max_size
+
+    @max_size.setter
+    def max_size(self, _max_size: int) -> None:
+        if _max_size > MAX_BUFFER_SIZE:
+            raise ValueError("Maximum chunk size is limited to ", MAX_CHUNK_SIZE)
+        self._max_size = _max_size
+
+    def __len__(self) -> int:
+        return len(self._data)
+
+
+class SyncBuffer(BufferBase[BufferData]):
+    """Synchronous reading interfaces of a buffer."""
+
+    def readall(self) -> Generator[BufferData, None, None]:
+        """
+        Returns an iterator that yields the first piece of data
+        until the buffer is empty.
+        """
+        while self._data:
+            yield self._data.pop(0)
+
+    def read(self) -> Union[BufferData, Type[Empty]]:
+        """
+        Pop the first piece of data if there is any data available,
+        else return ``queue.Empty``.
+        """
+        try:
+            return self._data.pop(0)
+        except IndexError:
+            return Empty
+
+
+class AsyncBuffer(BufferBase[BufferData]):
+    """Asynchronous reading interfaces of a buffer."""
+
+    async def readall(self) -> AsyncGenerator[BufferData, None]:
+        """
+        Returns an iterator that yields the first piece of data
+        until the buffer is empty.
+        """
+        while self._data:
+            yield self._data.pop(0)
+
+    async def read(self) -> Union[BufferData, Type[Empty]]:
+        """
+        Pop the first piece of data if there is any data available,
+        else return ``queue.Empty``.
+        """
+        try:
+            return self._data.pop(0)
+        except IndexError:
+            return Empty
+
+
+BufferType = Union[Type[SyncBuffer[BufferData]], Type[AsyncBuffer[BufferData]]]
+Buffer = Union[SyncBuffer[BufferData], AsyncBuffer[BufferData]]
+
+
+class Pipe(BufferBase[BufferData]):
+    """
+    Pipe object can create a buffer to read a data from the pipe,
+    or write a chunk or a piece of data into the pipe.
+
+    If there is an exception raised in the reading data context,
+    it will restore all data from the chunk into the pipe.
+
+    If the context reaches the end without any exception raised,
+    consumed data will be discarded from the buffer
+    and the rest will be restored into the pipe.
+    """
+
+    def __init__(
+        self,
+        *_initial_data: BufferData,
+        max_size: int = MAX_BUFFER_SIZE,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+    ) -> None:
+        super().__init__(*_initial_data, max_size=max_size)
+        self.chunk_size: int = chunk_size
+
+    @property
+    def chunk_size(self) -> int:
+        return self._chunk_size
+
+    @chunk_size.setter
+    def chunk_size(self, _chunk_size: int) -> None:
+        if _chunk_size > MAX_CHUNK_SIZE:
+            raise ValueError("Maximum chunk size is limited to ", MAX_CHUNK_SIZE)
+        self._chunk_size = _chunk_size
+
+    @contextmanager
+    def _open_buffer(
+        self, buffer_type: BufferType[BufferData]
+    ) -> Iterator[Buffer[BufferData]]:
+        """
+        Create a buffer of the ``buffer_type``.
+        """
+        try:
+            from copy import deepcopy
+
+            chunk, self._data = (
+                self._data[: self.chunk_size],
+                self._data[self.chunk_size :],
+            )
+            _buffer = buffer_type(*deepcopy(chunk), max_size=self.chunk_size)
+            yield _buffer
+        except Exception as err:
+            self._data = chunk + self._data
+            raise err
+        else:
+            consumed = max(0, len(chunk) - len(_buffer))
+            self._data = chunk[consumed:] + self._data
+        finally:
+            ...
+
+    @contextmanager
+    def open_readable(
+        self, timeout: int = 0, retry_interval: float = 0.1
+    ) -> Iterator[BufferBase[BufferData]]:
+        """
+        Yield a buffer containing a copy of the first chunk of data.
+        """
+        from ..core.schedulers import retry
+
+        max_trials = int(timeout / retry_interval) + 1
+
+        @retry(Empty, max_trials=max_trials, interval=retry_interval)
+        def wait_for_data() -> None:
+            if not self._data:
+                raise Empty("There is no data left in the pipe.")
+
+        wait_for_data()
+        try:
+            with self._open_buffer(SyncBuffer[BufferData]) as _buffer:
+                yield _buffer
+        finally:
+            ...
+
+    @asynccontextmanager
+    async def open_async_readable(
+        self, timeout: int = 0, retry_interval: float = 0.1
+    ) -> AsyncIterator[BufferBase[BufferData]]:
+        """
+        Yield an async buffer containing a copy of the first chunk of data.
+        """
+
+        from ..core.schedulers import async_retry
+
+        max_trials = int(timeout / retry_interval) + 1
+
+        @async_retry(Empty, max_trials=max_trials, interval=retry_interval)
+        async def wait_for_data() -> None:
+            if not self._data:
+                raise Empty("There is no data left in the pipe.")
+
+        await wait_for_data()
+
+        try:
+            with self._open_buffer(AsyncBuffer[BufferData]) as _buffer:
+                yield _buffer
+        finally:
+            ...
+
+    def write(self, data: BufferData) -> None:
+        """
+        Append a copy of a piece of data into the pipe.
+
+        Raises
+        ------
+        queue.Full
+            If the pipe will have more than ``self.max_size``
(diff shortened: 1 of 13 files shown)


src/beamlime/communication/pipes.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        Create a buffer of the ``buffer_type``.
+        """
+        try:
+            from copy import deepcopy
+
+            chunk, self._data = (
+                self._data[: self.chunk_size],
+                self._data[self.chunk_size :],
+            )
```
SimonHeybrock, line 148: So if I understand correctly, `Buffer` and its subclasses support some way of chunking data for some purpose. It feels like an odd and indirect interface. Can you elaborate a bit on where this is needed, and why you chose this solution?

src/beamlime/communication/pipes.py
```diff
@@ -0,0 +1,238 @@
...
+            _buffer = buffer_type(*deepcopy(chunk), max_size=self.chunk_size)
+            yield _buffer
+        except Exception as err:
```
SimonHeybrock, line 151: What are the things leading to exceptions here, and why is returning data into the buffer a solution?

src/beamlime/communication/pipes.py
```diff
@@ -0,0 +1,238 @@
...
+        ------
+        queue.Full
+            If the pipe will have more than ``self.max_size``
+            of data pieces with the new piece appended.
+        """
+        from copy import copy
+
+        if len(self._data) >= self.max_size:
+            raise Full(f"Pipe is full. There are {self.max_size} pieces of data.")
+        self._data.append(copy(data))
```
SimonHeybrock, line 222: Making a copy? Not sure we can afford this in general. Why is this needed?

src/beamlime/communication/pipes.py
```diff
@@ -0,0 +1,238 @@
...
+
+        await wait_for_data()
+
+        try:
+            with self._open_buffer(AsyncBuffer[BufferData]) as _buffer:
+                yield _buffer
+        finally:
+            ...
+
+    def write(self, data: BufferData) -> None:
```
SimonHeybrock, line 208: Is this meant to be written to from different threads? I can't see any locking mechanism (or other) to make this safe?

SimonHeybrock (commented):
Some questions (have not looked at the tests yet):

SimonHeybrock:
I suppose my top-level question should be: Why not use https://docs.python.org/3/library/asyncio-queue.html?

YooSunYoung:
> I suppose my top-level question should be: Why not use https://docs.python.org/3/library/asyncio-queue.html and/or https://docs.python.org/3/library/queue.html#module-queue?

It had to be a wrapper because each `Pipe[T]` was singleton class until `CachedProvider` was implemented.
Now it doesn't have to be singleton anymore so it lost one of main reasons to exist...

But I agree, it can use `asyncio.Queue` internally, I just avoided it since I didn't want to implement `none-async` and `async` interfaces... but I can not avoid to do it anyways...

src/beamlime/communication/pipes.py
```diff
@@ -0,0 +1,238 @@
...
+    TypeVar,
+    Union,
+)
+
+BufferData = TypeVar("BufferData")
+
+
+MAX_CHUNK_SIZE = 100
+DEFAULT_CHUNK_SIZE = int(MAX_CHUNK_SIZE / 10)
+MAX_BUFFER_SIZE = MAX_CHUNK_SIZE * 10
```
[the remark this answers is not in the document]
YooSunYoung: I wanted to discuss about that actually.
Do we maybe not need those limits...?

src/beamlime/communication/pipes.py
```diff
@@ -0,0 +1,238 @@
...
+    it will restore all data from the chunk into the pipe.
+
+    If the context reaches the end without any exception raised,
+    consumed data will be discarded from the buffer
+    and the rest will be restored into the pipe.
+    """
+
+    def __init__(
+        self,
+        *_initial_data: BufferData,
```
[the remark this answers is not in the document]
YooSunYoung: I think my intention was... `Pipe` shouldn't have `_initial_data` unless it's a partial chunk created from `Pipe` itself.
Should it start with `__` then...?

Or should it be just `initial_data` and allow `Pipe` to be created with initial data by users...?

src/beamlime/communication/pipes.py
```diff
@@ -0,0 +1,238 @@
...
+        Create a buffer of the ``buffer_type``.
+        """
+        try:
+            from copy import deepcopy
+
+            chunk, self._data = (
+                self._data[: self.chunk_size],
+                self._data[self.chunk_size :],
+            )
+            _buffer = buffer_type(*deepcopy(chunk), max_size=self.chunk_size)
```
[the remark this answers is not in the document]
YooSunYoung: It is actually not needed but should it have an option to do so...? Or will it not be needed at all...?

src/beamlime/communication/pipes.py
YooSunYoung: The idea was,
if anything goes wrong, the data should not be consumed from the buffer,
so that we don't lose any data that are not processed.

src/beamlime/communication/pipes.py
YooSunYoung: That is why it has `open_readable_*`  interface.
So it does not guarantee the order of data, but it will not lose any data that are not processed by the consumer.
I thought it'll be enough since `kafka` doesn't guarantee the order of data either.

src/beamlime/communication/pipes.py
YooSunYoung: Yes, you're right... it's not needed.

SimonHeybrock:
> > I suppose my top-level question should be: Why not use https://docs.python.org/3/library/asyncio-queue.html and/or https://docs.python.org/3/library/queue.html#module-queue?
> 
> It had to be a wrapper because each `Pipe[T]` was singleton class until `CachedProvider` was implemented. Now it doesn't have to be singleton anymore so it lost one of main reasons to exist.
> 
> So I agree, it can use `asyncio.Queue` internally, I just avoided it since I didn't want to implement `none-async` and `async` interfaces... but I can not avoid to do it anyways...

My question is: Why can't you use it directly, instead of "using it internally"? Note that `asyncio.Queue` says it is not thread-safe.

src/beamlime/communication/pipes.py
SimonHeybrock: I am not entirely convinced it is so easy to support concurrent writes and reads safely? Or I don't understand the mechanism.

src/beamlime/communication/pipes.py
SimonHeybrock: Goes wrong where? Doesn't this mean that the buffer will just run full, leading to problems elsewhere?

src/beamlime/communication/pipes.py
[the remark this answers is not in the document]
SimonHeybrock: Generally speaking, I think adding things that are not needed but "may be in the future" is almost always a bad idea.

We have quite high performance requirements, so any copying of data must be avoided, unless proven to be necessary.

YooSunYoung:
> My question is: Why can't you use it directly, instead of "using it internally"? Note that asyncio.Queue says it is not thread-safe.

The reasons I thought when I implemented this were,

- Provide easier interface to handle `timeout`
- Consistent interface regardless of using different type of data-stream or data-container (i.e. multi-process queue)
- Not to lose any data while consuming

Another main intention was to extend this `Pipe` to a `Kafka` listener.
`Kafka` consumer should send commit to the `broker` once it's done `consuming` data,
and I thought providing `chunk` of data from the queue within context manager will be the best solution.

SimonHeybrock:
Having consistent interfaces is often good 👍 

I do however wonder what leads to inconsistency in the first place? Below some questions one should ask:

> > My question is: Why can't you use it directly, instead of "using it internally"? Note that asyncio.Queue says it is not thread-safe.
> 
> The reasons I thought when I implemented this were,
> 
>     * Provide easier interface to handle `timeout`

When do these timeouts happen?

> 
>     * Consistent interface regardless of using different type of data-stream or data-container (i.e. multi-process queue)

Do we know yet if we ever will use multi-processing? Complicating the architecture and code now for hypothetical reasons may not be a good strategy.

> 
>     * Not to lose any data while consuming

How? Has it been shown that this is a problem? Why do we lose data when using the builtin `Queue`?

> Another main intention was to extend this `Pipe` to a `Kafka` listener. `Kafka` consumer should send commit to the `broker` once it's done `consuming` data, and I thought providing `chunk` of data from the queue within context manager will be the best solution.

From my outside view on this implementation: It is doing "many" things, making it hard to reason about (and test?):

- It is a pipe.
- It handles chunking, which seems very specialized for a particular case (which one? I assume not all uses of this `Pipe` may want chunking, or care about it?)
- It provides a very specialized way of error handling (putting data back), to me it is unclear where that is needed and what effects it may have in practice.
- It handles (deep)copying data, presumably to allow clients to then change/overwrite original data?

Doing those 4 things (instead of focusing on one) feels odd. It feels like multiple responsibilities handled by the same class?

YooSunYoung:
>> * Provide easier interface to handle `timeout`
>
> When do these timeouts happen?

It's usually at the beginning when it has to wait for the data to come in from the stream.
And at the end, when there is no data coming in from the stream. -> This one can be replaced with shutting down by the control manager somehow once it's finished...?


>> Do we know yet if we ever will use multi-processing? Complicating the architecture and code now for hypothetical reasons may not be a good strategy.

Not between data preprocessing, reducing and data-visualization(plotting),
but the dashboard that has user-interface to control applications should be in a different process I thought,
not to be blocked by data-processing demons, in case something's frozen.
But I agree, we can then implement it when it's needed instead.

>> * Not to lose any data while consuming
>
> How? Has it been shown that this is a problem? Why do we lose data when using the builtin Queue?

You have to `pop` out data from `Queue` and then process it,
which means if anything goes wrong during processing it, you can't easily have it back... I thought...?


> It handles chunking, which seems very specialized for a particular case (which one? I assume not all uses of this Pipe may want chunking, or care about it?)
It provides a very specialized way of error handling (putting data back), to me it is unclear where that is needed and what effects it may have in practice.

Chunking is only needed for `error handling (putting data back)`, so it's not really handling data chunking, it can simply provide all data it has to a consumer, for example.
Main use cases I was thinking of were,
when there are multiple consumers connected to a same `Pipe`, if any of them goes wrong for one of instances, other alive instances should be able to process them.
Or we can easily freeze the status and resume the process from there...? Like `pause` and `resume`.

> It handles (deep)copying data, presumably to allow clients to then change/overwrite original data?

It was concerning about `accidentally` changing data, but now I think we shouldn't presume that original data will be changed.... so we can erase this one...!

SimonHeybrock:
> > > * Provide easier interface to handle `timeout`
> > 
> > 
> > When do these timeouts happen?
> 
> It's usually at the beginning when it has to wait for the data to come in from the stream. And at the end, when there is no data coming in from the stream. -> This one can be replaced with shutting down by the control manager somehow once it's finished...?

But should the pipe handle the timeouts? I thought the consumer side would do that?

> 
> > > Do we know yet if we ever will use multi-processing? Complicating the architecture and code now for hypothetical reasons may not be a good strategy.
> 
> Not between data preprocessing, reducing and data-visualization(plotting), but the dashboard that has user-interface to control applications should be in a different process I thought, not to be blocked by data-processing demons, in case something's frozen. But I agree, we can then implement it when it's needed instead.

Then we should use https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue. It seems to have a consistent interface with https://docs.python.org/3/library/queue.html#queue.Queue, which we wanted?

> 
> > > * Not to lose any data while consuming
> > 
> > 
> > How? Has it been shown that this is a problem? Why do we lose data when using the builtin Queue?
> 
> You have to `pop` out data from `Queue` and then process it, which means if anything goes wrong during processing it, you can't easily have it back... I thought...?

And then what? Who would "have it back"?

> 
> > It handles chunking, which seems very specialized for a particular case (which one? I assume not all uses of this Pipe may want chunking, or care about it?)
> > It provides a very specialized way of error handling (putting data back), to me it is unclear where that is needed and what effects it may have in practice.
> 
> Chunking is only needed for `error handling (putting data back)`, so it's not really handling data chunking, it can simply provide all data it has to a consumer, for example. Main use cases I was thinking of were, when there are multiple consumers connected to a same `Pipe`, if any of them goes wrong for one of instances, other alive instances should be able to process them. Or we can easily freeze the status and resume the process from there...? Like `pause` and `resume`.

Then I misunderstood your code. It seemed it took out 10 items at a time (the default chunk size)?


> 
> > It handles (deep)copying data, presumably to allow clients to then change/overwrite original data?
> 
> It was concerning about `accidentally` changing data, but now I think we shouldn't presume that original data will be changed.... so we can erase this one...!

YooSunYoung:
> But should the pipe handle the timeouts? I thought the consumer side would do that?

Consumer decides how much time it will wait,
but a container of data like `Queue` usually provide `timeout` option along with `get(read)` interface.

> Then we should use https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue. It seems to have a consistent interface with https://docs.python.org/3/library/queue.html#queue.Queue, which we wanted?

`get` and `put` methods are the same but `asyncio.Queue` doesn't provide `timeout` option for the `get` interface.
You have to use `asyncio.wait_for`.
I don't have a good idea to handle `timeout` without a wrapper like this... do you have any...?

>>You have to pop out data from Queue and then process it, which means if anything goes wrong during processing it, you can't >>easily have it back... I thought...?
>
>And then what? Who would "have it back"?

`Pipe[T]` will have it back. And there is still no concrete application that is using this leftover in the `Pipe` yet...

> Then I misunderstood your code. It seemed it took out 10 items at a time (the default chunk size)?

You can set the `chunk_size` as maximum size and then it will return whatever number it has unless it's empty.

src/beamlime/communication/pipes.py
YooSunYoung: If I explain what it does more in detail...

When a consumer reads a buffer from the context manager,
`Pipe` splits a chunk from the original list and give the chunk to the consumer.
Once the context is exited without any error, it discard the chunk and the original list stays the same.
If there is any error raised within the context, it returns the chunk into the original list.

But writing doesn't matter since it will not `write` until the data is ready, so it is not a context manager, it's just appending interface.


Do you think it's not enough to say it's `safe` or is it too handling error too much...?

YooSunYoung:
Closing this for now until this feature is decided to be implemented.

PR #82, opened 2023-07-12
Add and re-organize CI recipes

YooSunYoung:


tests/core/schedulers_test.py
```diff
@@ -72,4 +72,4 @@ async def something_takes_time():
     assert time_consumed > expected_consumed_time
     assert (
         time_consumed - expected_consumed_time
-    ) < interval * 3  # Expected delta for several trials.
+    ) < interval * 5  # Expected delta for several trials.
```
YooSunYoung, line 75: I had to increase it for windows and macos...

tests/logging/logging_resources_test.py
```diff
@@ -120,6 +120,7 @@ def test_check_file_handlers(tmp_path: Path):
     hdlr = BeamlimeFileHandler(tmp_file)
     logger.addHandler(hdlr)
     assert hdlr in logger.handlers
+    hdlr.close()
```
YooSunYoung, line 123: Only `Windows` complained when I tried to remove the file while it's being used by the handler.

tests/core/schedulers_test.py
YooSunYoung: And it still fails time to time...!

YooSunYoung:
Closes due to duplicated PR via copier update #84

PR #84, opened 2023-07-17
Copier update on CI recipes.

YooSunYoung:
Reverse-applying the template created from https://github.com/scipp/beamlime/pull/82.


.copier-answers.yml
```diff
@@ -1,7 +1,8 @@
 # Changes here will be overwritten by Copier; NEVER EDIT MANUALLY
-_commit: 2e7a01f
-_src_path: gh:scipp/copier_template
+_commit: bd89f40
+_src_path: ../copier_template/
```
YooSunYoung, line 3: Will update again once 
`https://github.com/scipp/copier_template/pull/9` is merged...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+      matrix:
+
+        os: ['ubuntu-20.04']
+        python:
+          - version: '3.9'
+            tox-env: 'py39'
+          - version: '3.10'
+            tox-env: 'py310'
+          - version: '3.11'
+            tox-env: 'py311'
```
YooSunYoung, line 40: Four lines above are added on top of the template updates.

SimonHeybrock: I feel having to provide the tox env is a way to mess up. Could we have a default for the tox env, computed from the Python version? Then only someone who really needs a different one has to bother with this.

In the copier template, not here, I mean.

YooSunYoung: It seems like it will need two steps to do that,
1. Create an environment variable with manipulated string via bash command, 
    i.e. `run: echo ENV_VAR='${{ inputs.var }}' >> $GITHUB_ENV`
2. Use that environment variable as a github action variable via `${{ env.ENV_VAR }}`.

` ENV_VAR='${{ inputs.var }}' ` can be replaced with proper manipulation.

YooSunYoung: Manipulating variables in the template is much easier than in the GitHub action itself. 
It will not be written or updated so often anyways, no...?

SimonHeybrock: See my comment on Slack: Can't we determine the Python version from the tox env? It is tied to it anyway.

YooSunYoung: Yes, if you use `tox -e test` it should use the only python version the environment has,
but if you specify a python version under the [testenv], it might use that one instead.
I wasn't sure if that will be the case in other repositories.

And I'm not sure which one tox will use if there are multiple python versions available in the environment...

YooSunYoung: Or if we just do `tox -e py39`, that should just use the python3.9.
But you still need the Python version in the workflow, to set the environment before using tox.
Otherwise, tox can't find the python... as far as I know...?

YooSunYoung:
Closing this in favor of #86

PR #92, opened 2023-09-22
Update runtime type-checker of the dependency injection module.

YooSunYoung:
There are 2 main things updated in the runtime type-checker of the dependency injection module.
- ``NewType`` instance check.
  It was using it's own `NewTypeProtocol` for `NewType` instance check,
  since it is a function object for python version 3.9 or below.
  However, from python version 3.10 or above, ``NewType`` actually returns a class object.
- ``extract_underlying_type``: Extracting runtime type of the ``NewType``.
  Now it retrieves ``__supertype__`` recursively of the ``NewType`` to extract runtime returned type.
  It is mainly for checking the compatibility between a type and the provider.

And another minor thing is that ``extract_dependency_type`` is now a static method of the ``DependencySpec`` so that it is not confused with ``extract_underlying_type``.

### DIFF
--- a/src/beamlime/constructors/inspectors.py
+++ b/src/beamlime/constructors/inspectors.py
@@ -38,13 +38,6 @@ def validate_annotation(annotation: Any) -> Literal[True]:
 
 Product = TypeVar("Product")
 
-
-def _is_type_newtype_py310(tp: Any) -> bool:
-    from typing import NewType
-
-    return isinstance(NewType, type) and isinstance(tp, NewType)
-
-
 _NewT = NewType("_NewT", int)
 
 
@@ -59,7 +52,12 @@ def _is_type_newtype_py39(tp: Any) -> bool:
 
 class NewTypeMeta(type):
     def __instancecheck__(cls, instance: Any) -> bool:
-        return _is_type_newtype_py310(instance) or _is_type_newtype_py39(instance)
+        import sys
+
+        if sys.version_info < (3, 10):  # py39
+            return _is_type_newtype_py39(instance)
+        else:
+            return isinstance(instance, NewType)
 
 
 class NewTypeProtocol(Generic[Product], metaclass=NewTypeMeta):
--- a/src/beamlime/constructors/inspectors.py
+++ b/src/beamlime/constructors/inspectors.py
@@ -37,26 +37,38 @@ def validate_annotation(annotation: Any) -> Literal[True]:
 
 
 Product = TypeVar("Product")
+
+
+def _is_type_newtype_py310(tp: Any) -> bool:
+    from typing import NewType
+
+    return isinstance(NewType, type) and isinstance(tp, NewType)
+
+
 _NewT = NewType("_NewT", int)
+
+
+def _is_type_newtype_py39(tp: Any) -> bool:
+    return (
+        callable(tp)
+        and hasattr(tp, "__supertype__")
+        and _NewT.__module__ == tp.__module__
+        and _NewT.__qualname__ == tp.__qualname__
+    )
 
 
 class NewTypeMeta(type):
     def __instancecheck__(cls, instance: Any) -> bool:
-        return (
-            callable(instance)
-            and hasattr(instance, "__supertype__")
-            and _NewT.__module__ == instance.__module__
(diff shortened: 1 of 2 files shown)


src/beamlime/constructors/inspectors.py
```diff
@@ -37,26 +37,38 @@ def validate_annotation(annotation: Any) -> Literal[True]:
 
 
 Product = TypeVar("Product")
+
+
+def _is_type_newtype_py310(tp: Any) -> bool:
+    from typing import NewType
+
+    return isinstance(NewType, type) and isinstance(tp, NewType)
```
jl-wynen, line 45: What is the first check for?

src/beamlime/constructors/inspectors.py
```diff
@@ -37,26 +37,38 @@ def validate_annotation(annotation: Any) -> Literal[True]:
...
+
 class NewTypeMeta(type):
     def __instancecheck__(cls, instance: Any) -> bool:
-        return (
-            callable(instance)
-            and hasattr(instance, "__supertype__")
-            and _NewT.__module__ == instance.__module__
-            and _NewT.__qualname__ == instance.__qualname__
-        )
+        return _is_type_newtype_py310(instance) or _is_type_newtype_py39(instance)
```
jl-wynen, line 62: Shouldn't this use `sys.version < (3, 10)` or something similar to select one of the two functions?

jl-wynen, line 62: Also, have a look at this: https://github.com/pydantic/pydantic/blob/c83e5fcd2da5693d4dbcf031f749be66a81b322d/pydantic/_internal/_typing_extra.py#L118

I'm not saying your solution is wrong, but theirs looks simpler.

src/beamlime/constructors/inspectors.py
```diff
@@ -109,33 +121,47 @@ def get_product_spec(callable_obj: Callable[..., Product]) -> ProductSpec:
...
-    if get_origin(tp) == Union:
-        if len((args := get_args(tp))) == 2 and type(None) in args:
-            # Optional
-            return args[0] if isinstance(None, args[1]) else args[1]
-        else:
-            raise NotImplementedError(
-                "Union annotation for dependencies is not supported."
-            )
-    else:
-        return tp
```
jl-wynen, line 124: Why did you move it?

tests/constructors/inspection_test.py
```diff
@@ -41,6 +41,16 @@ def test_new_type_underlying_type_retrieved():
     assert product_spec.returned_type is int
 
 
+def test_multi_inheritied_new_type_underlying_type_retrieved():
+    from typing import NewType
+
+    new_type = NewType("new_type", int)
+    new_new_type = NewType("new_new_type", new_type)
```
jl-wynen, line 48: Is this actually used in practice?

src/beamlime/constructors/inspectors.py
YooSunYoung: Because `NewType` is a function not a type, in python 3.9 or below.

src/beamlime/constructors/inspectors.py
YooSunYoung: That is true... I just added the check

YooSunYoung: Yeah it's true, I needed to check the ``.__module__`` and ``.__qualname__`` because checking ``__supertype__`` was not sufficient in some cases though... I'll just keep it this way for now. It'll be unnecessary from py3.10 anyways...

src/beamlime/constructors/inspectors.py
YooSunYoung: Just to specify the scope of this function more in detail and distinguish it from `extract_underlying_type`.
This one is only for argument (dependency) type-checks.

And `extract_underlying_type` is used both for argument and returned type handling.

tests/constructors/inspection_test.py
YooSunYoung: Yes...! I accidentally made a `NewType` of a `NewType` while writing multiple workflows...
And there is no reason not to support these I thought.

jl-wynen (approved):

PR #93, opened 2023-09-22
Singleton provider argument equality check (hashable/unhashable)

YooSunYoung:
Cached provider is internally using ``lru_cache`` to check if it was called with the same arguments.
Therefore it can not be used if the arguments cannot be hashed.


src/beamlime/constructors/providers.py
```diff
@@ -233,10 +233,23 @@ def __init__(
...
         self.cached_args = CachedArguments()
         self.cache_indicator = lru_cache(maxsize=2)(self.cached_args)
 
+    def _check_args_hashability(self) -> None:
+        if unhashable_args := {
+            arg_name: arg_spec
+            for arg_name, arg_spec in self.arg_dep_specs.items()
+            if arg_spec.runtime_type.__hash__ is None
+        }:
+            raise TypeError(
```
nvaytet, line 248: In what kind of situation do you get an unhashable type? Is it relatively common? I.e. do we need to come up with an alternative handling instead of raising an error?

I guess coming up with an alternative may require fundamental changes to the rest of the code base?
Or maybe it is simply because we are using `lru_cache` and we could implement our own caching instead?

(sorry lots of questions...)

tests/constructors/cached_provider_test.py
```diff
@@ -0,0 +1,58 @@
...
+    assert cached_random_number(123) == cached_random_number(123)
+
+    with pytest.raises(CachedProviderCalledWithDifferentArgs):
+        cached_random_number(100)
+
+
+def test_cached_provider_unhashable_arguments_raises():
+    from beamlime.constructors.providers import CachedProvider
+
+    with pytest.raises(TypeError):
```
nvaytet, line 57:
```suggestion
    with pytest.raises(TypeError, match='cannot be cached. Unhashable argument'):
```

tests/constructors/cached_provider_test.py
```diff
@@ -0,0 +1,58 @@
...
+    import random
+
+    random.seed(seed)
+    return random.random()
+
+
+def test_cached_provider_without_arguments():
+    cached_random_number = CachedProvider(random_number)
+    assert cached_random_number() == cached_random_number()
+    assert cached_random_number() != random_number()
```
nvaytet, line 28: Just for my understanding: Is this because of the `maxsize=2`?

src/beamlime/constructors/providers.py
```diff
@@ -233,10 +233,23 @@ def __init__(
         from functools import lru_cache
 
         super().__init__(_constructor, *args, **kwargs)
+        self._check_args_hashability()
+
         self.cached_result: Product
         self.cached_args = CachedArguments()
         self.cache_indicator = lru_cache(maxsize=2)(self.cached_args)
```
nvaytet, line 240: What is the reason for the `maxsize=2`? I guess a comment for clarity would help ;-)

src/beamlime/constructors/providers.py
YooSunYoung: >  do we need to come up with an alternative handling instead of raising an error?
Probably. I just came across this case where;
One of the arguments was not hashable, but it is from a cached provider, so it is guaranteed that it's the same input.
In this case, it shouldn't need to be checked if it is hashable or not.

> Or maybe it is simply because we are using lru_cache and we could implement our own caching instead?
It is internallly using `lru_cache` and `lru_cache` doesn't allow unhashable arguments. That was why it is checking if it is hashable or not.

> (sorry lots of questions...)
Not at all : D...! Thank you for reviewing

src/beamlime/constructors/providers.py
YooSunYoung: Yes...!
`CachedProvider` is not supposed to be called with different arguments.
So it is not necessary to have more than 2 cache size... I thought...

tests/constructors/cached_provider_test.py
YooSunYoung: It's because it always returns the first returned value.
And `cached_random_number` and `random_number` is sharing the same `random` object.
That's why it has another test with `random_number_with_seed`, to initialize random seed between calls.

src/beamlime/constructors/providers.py
nvaytet: Wouldn't it be maxsize=1 then?

nvaytet: In plopp, we implemented a very simple caching mechanism where we store a variable `self._data` in the nodes that is set to `None` by default.

Could you do something similar where you store the args the first time it is called, and then check if the args are the same on subsequent calls?

This would also remove the annoyances with the hashable args.

YooSunYoung: That is actually what I tried at the beginning and then `==` was not very useful for comparison in many cases, since it automatically falls back to reference comparison, which will be `False` even though they have same `values`.

It's also not clear which method to use for comparison.
When you do `==` between two scipp data arrays or variables, it gives a data array or a variable and then it can't be simply used as a condition of `if` statement.... (right...?) so I had to do sth like `(da == da).all().value` to retrieve `True/False`.

Another reason was that `lru_cache` has a fast way to compare the arguments so I didn't have to deal with that.

src/beamlime/constructors/providers.py
```diff
@@ -208,70 +212,204 @@ def __copy__(cls, _obj: Provider[Product]) -> Provider[Product]:
...
+class ArgumentsFilter(_ArgumentsInstanceFilter):
+    """
+    Remember arguments and check if the new arguments match the first one.
+    """
+
+    def __init__(self, func: Callable[..., Any]) -> None:
+        from inspect import Signature, signature
+
+        self._sig: Signature = signature(func)
+        self.unhashable_filter = _ArgumentsInstanceFilter()
```
nvaytet, line 326: I find it a little strange that the `ArgumentsFilter` both inherits from the `_ArgumentsInstanceFilter` and has another instance of `_ArgumentsInstanceFilter` as a member... Does it really need to inherit as well?

src/beamlime/constructors/providers.py
```diff
@@ -208,70 +212,204 @@ def __copy__(cls, _obj: Provider[Product]) -> Provider[Product]:
...
 
-    It remembers only the first returned value and re-use it on next call.
     Therefore passing different arguments from the first call is not allowed.
-    ``functools.lru_cache`` is used to check if the arguments are the same.
+    ``functools.lru_cache`` is used to check if the hashable arguments are the same.
+    For unhashable arguments, ``cached_arguments_filter``
+    check if they are the same instances.
+
+    ``SingletonProvider`` is needed for sharing the same object by multiple objects.
+    The arguments filter prioritize instance equality to hash equality.
```
nvaytet, line 385:
```suggestion
    The arguments filter prioritizes instance equality over hash equality.
```

src/beamlime/constructors/providers.py
YooSunYoung: Yes, because if all `arguments` are the same instances, it does not have to split them into two groups (hashable/unhashable). Therefore `ArgumentsFilter` could use the `_ArgumentsInstanceFilter` methods.

I could just have 3 members,
1. `_ArgumentsInstanceFilter` that checks all `arguments`
2. `_ArgumentsInstanceFilter` that only checks unhashable `arguments`
3. `_ArgumentHashFilter` that only checks hashable `arguments`.

I also thought about only allowing `singleton` dependencies as JL suggested, but then it seems like too strict... that'll end up wrapping all providers with `SingletonProvider`...

src/beamlime/constructors/providers.py
```diff
@@ -208,70 +212,204 @@ def __copy__(cls, _obj: Provider[Product]) -> Provider[Product]:
...
+
+    def _all_kwargs_same_instances(self, new_kwargs: dict):
+        return all(
+            [
+                kwarg is n_kwarg
+                for kwarg, n_kwarg in zip(self.kwargs.values(), new_kwargs.values())
+            ]
+        )
+
+    def filter_arguments(self, args: tuple, kwargs: dict) -> bool:
```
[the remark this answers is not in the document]
YooSunYoung: Yes, I also thought it was a confusing name.
How about... `arguments_filter` ...?

src/beamlime/constructors/providers.py
YooSunYoung: Yeah that sounds like a very good idea....!

YooSunYoung:
Last three commits are just fixing docstring and shortening import line as `SingletonProvider` is now exposed from the `constructors` module...!

src/beamlime/constructors/providers.py
```diff
@@ -303,18 +240,83 @@ def split_dict_by_filter(
...
+        if not self.was_called():
+            self.save_arguments(arguments)
+            filtered, rest = arguments, dict()
+        else:
+            filtered, rest = split_dict_by_filter(self.arguments_filter, arguments)
+
+        return rest if reverse else filtered
+
+
+class _ArgumentsHashFilter(_ArgumentsInstanceFilter):
```
nvaytet, line 279: Sorry to be a pain, but I still think it's a little weird that the `_ArgumentsHashFilter` inherits from the `_ArgumentsInstanceFilter`, as they are doing quite different things.

If you don't want to duplicate the `__call__` and `was_called` methods, then I think it would be cleaner, and make more sense conceptually, to have maybe a `_ArgumentsFilter` mixin, and then both Instance and Hash filters could inherit from that?

YooSunYoung: Sorry I found this comment very late...! I made an abstract class `_ArgumentFilter` as you suggested and now each type of argument filter inherit the abstract class instead of having multi-depth inheritance...! Thank you!

nvaytet (approved):

PR #94, opened 2023-09-22
Add prototypes for benchmark testing.

YooSunYoung:
There are 2 minimum prototype daemons with dummy workflows for benchmark testings.
This PR does not have benchmarking code. It only has unit tests of each prototypes.

## Review points
- ``prototype_mini.StopWatch`` is used to benchmark the process.
- Three main applications, ``DataStreamListener``/``DataReduction``/``Visualization`` of prototypes.
- Dummy workflows in ``workflows``.

## Minimum prototype
These are the minimum components of the real-time data reduction testing.
- ``tests/prototypes/prototype_mini.py``
- ``tests/prototypes/parameters.py``
- ``tests/prototypes/random_data_providers.py``
- ``tests/prototypes/workflows.py``.

``prototype_mini`` can also be run as a script, ``python -m tests.prototypes.prototype_mini``.
It simulates the data stream between async daemons.
The dummy workflow contains binning, transform coordinates, and histogram.

## Kafka streaming tests
``tests/prototypes/prototype_kafka.py`` contains the same daemons as ``prototype_mini.py`` but with real kafka api and ``protobuf`` serialization/deserialization process.
``kafka`` broker should be running in the same machine in order to run this test.

## Kafka related tests filter.
`kafka` related tests will not be included in the CI test (for now) since it requires `kafka` broker running in the same machine.
Those tests can be filtered by requesting ``kafka_test`` fixture.
``kafka_test`` fixture will then check if there was a ``--kafka-test`` flag included as an argument of ``pytest`` command and then skip the test if it was not.
You can include those tests by ``pytest --kafka-test``.

## The reason for using ``fixture`` and not ``mark``.
Adding extra marks is also an option to include/exclude marked tests.
It should be configured in the ``pytest.ini``.
``fixture`` was used for now since ``mark`` has limitation of how to decide whether it is skipping the test or not.


tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+
+        However, sometimes it is automatically handled within the thread,
+        and it caused errors which was hard to be debugged under ``pytest`` session.
+        For example,
+        - The new event loop was not closed properly as it is destroyed.
+        - The new event loop was never started until it is destroyed.
+        ``Traceback`` of ``pytest`` did not show
+        where exactly the error is from in those cases.
+        It was resolved by using ``get_event_loop``,
+        or manually closing the event loop at the end of the test.
```
YooSunYoung, line 368: Test case to reproduce similar error: 
```python
async def some_coroutine() -> bool:
    import asyncio
    for _ in range(100):
        try:
            raise RuntimeError
        except RuntimeError:
            await asyncio.sleep(0.01)
    return True


def wrong_way_of_running_async_calls():
    import asyncio
    new_loop = asyncio.new_event_loop()
    return new_loop.run_until_complete(some_coroutine())


def test_foo():
    assert wrong_way_of_running_async_calls()

def test_foo1():
    assert wrong_way_of_running_async_calls()

def test_foo2():
    assert wrong_way_of_running_async_calls()

```

YooSunYoung:
Script for deleting all randomly generated topic with prefix of ``BEAMLIMETEST``.

```python
from confluent_kafka.admin import ClusterMetadata

def filter_beamlime_test_topics(cluster_meta: ClusterMetadata) -> list[str]:
    test_topic_prefix = "BEAMLIMETEST"
    return [topic for topic in cluster_meta.topics if topic.startswith(test_topic_prefix)]

if __name__ == "__main__":
    import time

    from confluent_kafka.admin import AdminClient
    
    admin = AdminClient({'bootstrap.servers': 'localhost:9092'})
    topic_list = filter_beamlime_test_topics(admin.list_topics())
    admin.list_topics().topics
    
    if topic_list:
        print("Deleting all existing topics: ", topic_list)
        admin.delete_topics(topic_list)

        while (topic_list:=filter_beamlime_test_topics(admin.list_topics())):
            time.sleep(1)
        
        print('Test topics all deleted.')
    else:
        print("No test topics to delete.")


```

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+
+        data_monitor = self.data_pipe_monitor(self.visualized_data_pipe)
+
+        if hasattr(self, "fig"):
+            del self.fig
+
+        if await data_monitor():
+            self.first_data = self.visualized_data_pipe.pop(0)
+            self.debug("First data as a seed of histogram: %s", self.first_data)
+            self.stream_node = pp.Node(lambda: self.first_data)
```
nvaytet, line 269:
```suggestion
            self.stream_node = pp.Node(self.first_data)
```

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+        yield loop
+        loop.close()
+
+
+class BasePrototype(BaseApp, ABC):
+    data_stream_listener: DataStreamListener
+    data_merge: DataMerge[Events, MergedData]
+    data_binning: DataBinning[MergedData, PixelGrouped]
+    data_reduction: DataReduction[PixelGrouped, ReducedData]
+    data_plotter: DataHistogramming[ReducedData, Histogrammed]
```
nvaytet, line 314:
```suggestion
    data_merger: DataMerge[Events, MergedData]
    data_binner: DataBinning[MergedData, PixelGrouped]
    data_reducer: DataReduction[PixelGrouped, ReducedData]
    data_histogrammer: DataHistogramming[ReducedData, Histogrammed]
```
?

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+        For example,
+        - The new event loop was not closed properly as it is destroyed.
+        - The new event loop was never started until it is destroyed.
+        ``Traceback`` of ``pytest`` did not show
+        where exactly the error is from in those cases.
+        It was resolved by using ``get_event_loop``,
+        or manually closing the event loop at the end of the test.
+
+        **When to use ``asyncio.new_event_loop``.**
+          - ``asyncio.get_event_loop`` raises ``RuntimeError``
```
nvaytet, line 371: In what cases does this happen?

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+            self.data_reduction,
+            self.data_plotter,
+            self.visualizer,
+        ]
+
+    def run(self):
+        """
+        Collect all coroutines of daemons and schedule them into the event loop.
+
+        Notes
```
nvaytet, line 331: :+1:

tests/prototypes/random_data_providers.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def calculate_event_per_frame(
+    frame_rate: FrameRate, event_rate: EventRate
+) -> EventFrameRate:
+    return EventFrameRate(int(event_rate / frame_rate))
+
+
+def provide_time_coords(rng: RNG, ef_rate: EventFrameRate) -> TimeCoords:
+    dummy_zeros = [zr for zr in range(int(13620492e11), int(13620492e11) + ef_rate)]
```
nvaytet, line 38: where does the number `13620492e11` come from?

tests/prototypes/random_data_providers.py
```diff
@@ -0,0 +1,113 @@
...
+    et_offset = sc.array(
+        dims=["event"], values=rng.random((ef_rate,)) * 800 + 200, unit='ns'
+    )
+
+    return TimeCoords({"event_time_zero": et_zero, "event_time_offset": et_offset})
+
+
+def provide_dummy_counts(ef_rate: EventFrameRate) -> DetectorCounts:
+    return DetectorCounts(
+        sc.ones(dims=["event"], shape=(ef_rate,), unit=sc.units.counts)
```
nvaytet, line 49:
```suggestion
        sc.ones(sizes={"event": ef_rate}, unit='counts')
```

tests/prototypes/workflows.py
```diff
@@ -0,0 +1,122 @@
...
+def transform_coords(
+    binned: PixelGrouped,
+    c_a: ConstantA = default_c_a,
+    c_b: ConstantB = default_c_b,
+    c_c: ConstantC = default_c_c,
+) -> ReducedData:
+    return ReducedData(
+        binned.transform_coords(
+            ['L', 'wavelength'],
+            graph={
```
nvaytet, line 82: Is that the same graph as on L49? Can we avoid the duplication?
Maybe this requires `c_a`, `c_b`, and `c_c` to also be part of the graph, but maybe that is a good thing?

tests/prototypes/workflows.py
```diff
@@ -0,0 +1,122 @@
...
+
+def bin_pixel_id(da: MergedData, pixel_bin_coord: PixelIDEdges) -> PixelGrouped:
+    return PixelGrouped(da.group(pixel_bin_coord))
+
+
+def transform_coords(
+    binned: PixelGrouped,
+    c_a: ConstantA = default_c_a,
+    c_b: ConstantB = default_c_b,
+    c_c: ConstantC = default_c_c,
```
nvaytet, line 77: Not sure about having default values here?

nvaytet (commented):
I have to admit I don't understand how all the different parts come together here, I would have to spend much longer looking at this to understand it all.

However, I had a superficial look and tried to spot things/patterns that may seem odd.

About:

> kafka related tests will not be included in the CI test (for now) since it requires kafka broker running in the same machine.

It would be nice to get this up and running soon, because there is a lot of code that pertains to Kafka that is currently not being automatically tested (we rely on the contributors to run those tests locally, which is quite dangerous).

tests/prototypes/prototype_kafka.py
```diff
@@ -0,0 +1,395 @@
...
+        self.stop_watch.start()
+
+    async def run(self) -> None:
+        with self.consumer_cxt() as consumer:
+            self.start_stop_watch()
+            data_chunk: Events = Events([])
+            i_frame = 0
+            while (event := self.poll_one_data(consumer)) is not None:
+                data_chunk.append(event)
+                if len(data_chunk) >= self.chunk_size and (i_frame := i_frame + 1):
```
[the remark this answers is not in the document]
YooSunYoung: No, it was just to bind it with the condition `len(data_chunk) >= self.chunk_size` .... so that I don't forget to change `i_frame` as well if the condition changes.
Is it more confusing than convenient....?

tests/prototypes/prototype_kafka.py
```diff
@@ -0,0 +1,395 @@
...
+
+    from .parameters import EventRate, NumPixels
+    from .prototype_mini import run_prototype
+
+    kafka_factory = kafka_prototype_factory()
+    kafka_factory[BeamlimeLogger].setLevel(logging.DEBUG)
+
+    run_prototype(
+        kafka_factory,
+        parameters={EventRate: 10**4, NumPixels: 10**4, NumFrames: 140},
```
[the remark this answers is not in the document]
YooSunYoung: I had to run this many times so... maybe...?

tests/prototypes/random_data_providers.py
YooSunYoung: It's just a random datetime I picked... `2013-02-28T12:00:00` : D.... Actually event_time_zero was not used in this dummy workflow. Maybe it should be...

tests/prototypes/workflows.py
YooSunYoung: Yes probably...?
so the idea was to keep `workflow_script` from L36 to L58 to be a through workflow script from the input to the output,
and the rest is granular steps of that workflow, probably run by `sciline` later.
So they are meant to be duplicated.
But the graph might be separated and re-used I agree.

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+    output_pipe: List[OutputType]
+
+    def __init__(self) -> None:
+        self.input_type = self._retrieve_type_arg('input_pipe')
+        self.output_type = self._retrieve_type_arg('output_pipe')
+        self.data_counts = 0
+        super().__init__()
+
+    @classmethod
+    def _retrieve_type_arg(cls, attr_name: str) -> type:
```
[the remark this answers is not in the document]
YooSunYoung: It is retrieving type argument from generic types, for example `Union[int, float]` and `int` and `float` are type arguments. So I would keep the name but I'll add some docstring...!

tests/prototypes/prototype_mini.py
YooSunYoung: > ``RuntimeError`` if there has been an event loop set in the thread object before but it is now removed.
I added an indicator to the related information above. Does this make sense...?

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+                raise TimeoutError
+
+        async def is_pipe_filled() -> bool:
+            try:
+                await wait_for_preferred_size()
+            except TimeoutError:
+                ...
+            return len(pipe) >= target_size
+
+        return is_pipe_filled
```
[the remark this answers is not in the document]
YooSunYoung: It should return the function object `is_pipe_filled`.

It has triple steps, 
in `data_pipe_monitor`
1. `wait_for_preferred_size` will be re-run until it reaches the condition, `len(pipe) < preferred_size` in this case, and throw an error if it does't meet the condition.
2. `is_pipe_filled` will use `wait_for_preferred_size` to wait until the `pipe` is filled up to `preferred size`, and returns True/False based on the `target_size`.

`target_size` is what you `need` at the minimum and `preferred_size` is what you want to wait for.
For example, you might want to wait until there are 10 pieces of data in the list but you only need 1 to proceed.
At the end `data_pipe_monitor` will give you this handle `is_pipe_filled` to see if the `pipe` object is filled with `target_size` pieces of data.

3. `data_pipe_monitor` create two functions above and return an async function `is_pipe_filled` so that it could be awaited in while loop or in an if condition.

`data_pipe_monitor` could also just handle the error and return `len(pipe) >= target_size` but then
it has to create `wait_for_preferred_size` every time it checks the size of the pipe so I wrapped it this way.

Does this make sense....?

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,472 @@
...
+                "Stop-timestamp is not available. ``stop`` was never called."
+            )
+        else:
+            return self._stop_timestamp
+
+    def start(self) -> None:
+        import time
+
+        try:
+            self.start_timestamp
```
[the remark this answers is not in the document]
YooSunYoung: `start/stop_timestamp` is used like 
```python
self.stop_watch.start()
self.stop_watch.stop()
self.info(
                "Benchmark result: %s",
                self.stop_watch.stop_timestamp - self.stop_watch.start_timestamp,
            )
```

And `start` and `stop` is for fixing values of those two properties.
I guess `stop_timestamp` - `start_timestamp` could be wrapped into one property, sth like `record`...?

tests/prototypes/prototype_kafka.py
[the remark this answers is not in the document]
nvaytet: > Is it more confusing than convenient....?

Maybe?

tests/prototypes/prototype_mini.py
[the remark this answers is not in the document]
nvaytet: Can you make an additional property, e.g. `self.duration` which would just compute the difference between the timestamps?

tests/prototypes/prototype_kafka.py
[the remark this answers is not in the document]
YooSunYoung: I added argparser as well. Now we can run this like
```bash
python -m tests.prototypes.prototype_mini --event-rate 10_000_000
```

tests/prototypes/workflows.py
```diff
@@ -0,0 +1,105 @@
...
+    da_list: Events, num_pixels: NumPixels, histogram_bin_size: HistogramBinSize
+) -> Histogrammed:
+    merged = sc.concat(da_list, dim='event')
+    pixel_ids = sc.arange(dim='pixel_id', start=0, stop=num_pixels)
+    binned = merged.group(pixel_ids)
+
+    graph = provide_coord_transform_graph()
+
+    transformed = binned.transform_coords(['L', 'wavelength'], graph=graph)
+    return transformed.hist(wavelength=histogram_bin_size).sum('L')
```
SimonHeybrock, line 57: For instrument like DREAM we need thousands of bins. This will run out of memory, instead we will have to use something like

```python
return transformed.bins.concat('L').hist(wavelength=histogram_bin_size)
```

tests/prototypes/workflows.py
```diff
@@ -0,0 +1,141 @@
...
+    return CoordTransformGraph(
+        {
+            'L1': lambda pixel_id: (pixel_id * c_a) + c_b,
+            'L2': lambda pixel_id: (pixel_id * c_a) + c_b,
+            'L': lambda L1, L2: L1 + L2,
+            'tof_min': lambda L: (L * scale_factor * lambda_min).to(unit=sc.units.ns),
+            'frame_offset': lambda event_time_zero: event_time_zero - first_pulse_time,
+            'time_offset_pivot': time_offset_pivot,
+            'tof': tof_from_time_offset,
+            'wavelength': wavelength_from_tof,
```
YooSunYoung, line 74: @SimonHeybrock I added converting `event_time_offset`, `event_time_zero` to `tof` in this graph.
Or should I just use `scippneutron` ...?

Haven't run the whole set of benchmarks yet though.

tests/prototypes/prototype_mini.py
```diff
@@ -0,0 +1,508 @@
...
+    input_pipe: List[Events]
+    output_pipe: List[MergedData]
+
+    def format_received(self, data: Any) -> str:
+        return f"{len(data)} pieces of {self.input_type.__name__}"
+
+    def process_first_data(self, data: Events) -> None:
+        sample_event = data[0]
+        first_pulse_time = sample_event.coords['event_time_zero'][0]
+        self.workflow.providers[FirstPulseTime] = lambda: first_pulse_time
```
YooSunYoung, line 252: `DataReductionApp` has separate `process_first_data` method so that it can retrieve the `FirstPulseTime` (which is the `event_time_zero` of the first incoming data in this case).

And then it updates the `workflow` providers to return the retrieved value.

tests/prototypes/prototype_kafka.py
```diff
@@ -0,0 +1,395 @@
...
+        return None
+
+
+class KafkaListener(KafkaListenerBase):
+    def deserialize(self, data_list: list[bytes]) -> sc.DataArray:
+        import numpy as np
+        from streaming_data_types.eventdata_ev44 import deserialise_ev44
+
+        data = deserialise_ev44(self.merge_bytes(data_list))
+        event_zeros = np.full(len(data.pixel_id), data.reference_time[0])
```
SimonHeybrock, line 317: Initializing a full numpy array and copying into a variable is a bit inefficient, should directly init a scipp variable. Use `scipp.full`?

tests/prototypes/workflows.py
SimonHeybrock: Yes, we absolutely should use `scippneutron`. If we start replicating logic here we will run into big problems, since things will never be in sync.

tests/prototypes/workflows.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def workflow_script(
+    da_list: Events,
+    num_pixels: NumPixels,
+    histogram_bin_size: HistogramBinSize,
+    frame_rate: FrameRate,
+) -> Histogrammed:
+    merged = sc.concat(da_list, dim='event')
+    pixel_ids = sc.arange(dim='pixel_id', start=0, stop=num_pixels)
```
SimonHeybrock, line 86: This will be used for every processed chunk, can we avoid recreating it all the time?

tests/prototypes/workflows.py
```diff
@@ -0,0 +1,141 @@
...
+    binned = merged.group(pixel_ids)
+
+    graph = provide_coord_transform_graph(frame_rate)
```
SimonHeybrock, line 89: Why is this recreated every time?

tests/prototypes/workflows.py
```diff
@@ -0,0 +1,141 @@
...
+) -> Histogrammed:
+    return reduced_data.hist(wavelength=bin_size).sum('L')
+
+
+Workflow = NewType("Workflow", Factory)
+
+
+def provide_workflow(
+    num_pixels: NumPixels, histogram_binsize: HistogramBinSize, frame_rate: FrameRate
+) -> Workflow:
```
SimonHeybrock, line 125: All the code here seems to be duplicating what is in `workflow_script` above. Are both used?

SimonHeybrock (commented):
I mainly looked at the places where data is touched/transformed, I believe @nvaytet took care of reviewing the other code?

YooSunYoung (commented):
I'm not sure if this is what you meant... @SimonHeybrock but I replaced the existing graph with frame unwrapping helper from ``scippneutron`` package.

SimonHeybrock (approved):

PR #102, opened 2023-11-01
Very simple kafka topic managing tool.

YooSunYoung:
This kind of tool was needed for the kafka integration tests.
Those tests create random topics and delete them at the end,
but sometimes it couldn't reach there when the test crashes or interrupted so I had to manually delete the them not to use infinite amount of storage/memory.

## Demo
It shows/delete topics that are not protected ones(which starts with ``_`` or ``__``) of the connected broker.

As I mentioned in the documentation, if more functionalities are needed,
we can use existing web applications, that are available as docker containers.

I used textual because it works in the terminal...!

kafka_tool_demo


tests/executables/kafka_topic_gui.py
```diff
@@ -0,0 +1,320 @@
...
+        self.topics.remove()
+        self.topics = SelectionList()
+        self.topic_container.mount(self.topics)
+
+    @work(exclusive=True)
+    async def delete_confirmed(self, selected: list[str]) -> None:
+        import asyncio
+
+        selected = self.topics.selected
+        self.admin.delete_topics(selected)
```
jokasimr, line 171: Doesn't the `delete_topics` method return some kind of future that indicates when the topics are deleted?
If it does then I think it would be better to await that future instead of sleeping for a fixed time.

jokasimr (commented):
The demo looks very good. However I would also suggest taking a look at the official command line interface [kafka-tools.html#kafka-topics-usage](https://docs.confluent.io/kafka/operations-tools/kafka-tools.html#kafka-topics-usage).

Would something like
```./bin/kafka-topics.sh --zookeeper localhost:1234 --delete --topic 'BEAMLIMETEST*'```
work in this case and delete all the temporary topics? Note that it accepts regex.

YooSunYoung:
> Would something like `./bin/kafka-topics.sh --zookeeper localhost:1234 --delete --topic 'BEAMLIMETEST*'` work in this case and delete all the temporary topics? Note that it accepts regex.

Yes, I actually tried wrapping them into bash shell script,
but it's always wrapped in docker so I would have to do sth like `docker exec container-name ${KAFKAHOME}/bin/kafka-topics.sh blah blah` so... there were some trivial extra steps needed to make the script robust... 
and the kafka server might not be running on the same machine, then there are even more extra steps...

But we're indeed missing regex,
actually we need a regex filter helper on topics since they are gonna have different prefix per instrument... 🤔

tests/executables/kafka_topic_gui.py
YooSunYoung: Yes, that makes more sense. I'll fix it that way.

jokasimr (approved):

tests/executables/kafka_topic_gui.py
YooSunYoung: I fixed it so that the users will see this waiting screen while kafka topics are deleted, I add artificially added async sleep line so that you can see the waiting screen but it almost never shows up since topics are automatically deleted in the testing environments.

wait_screen

PR #104, opened 2023-11-06
Benchmark session/runner helper.

YooSunYoung:
I put all the benchmark helper modules/functions under `tests/benchmarks`.

Loading benchmark results, visualizing and more pre-defined benchmark functions will be in the next PR.

## Context (The reason why not using existing ones.)

Existing benchmark helpers, extensions interfaces accept callable and arguments
so that they can run the callable object with arguments
and then save it to the session object or returns the benchmark results.

However, with `beamlime`, we build an application by dependency injection,
which means most of the arguments should be already defined when we build the callable object,
not when the callable object is called.

So, if we include the building stages in the benchmark, then it will include non-trivial time consuming computation which is not in our interest of benchmarking, and we would need to exclude the overhead.

Therefore we needed a benchmark helper that
- accepts different parameters/arguments and build the application from them 
- only measures the performance of the processes of our interest.
- works with async calls.

## Review points
- ``tests/benchmarks/runner.py``
The `BenchmarkSession` class is a bridge object that has a container of benchmark reports, and benchmark runner, file manager and configurations.
The benchmark reports(`BenchmarkReports`), file manager (`BenchmarkFileManager`) and configurations (`BenchmarkSessionConfiguration`) are not expected to be replaced with custom ones.

However, the benchmark runner (``BenchmarkRunner``) are meant to be replaced with custom runners so it is abstract class.

``SimpleRunner`` demonstrates what a runner should do,
and it is similar to existing benchmark helpers that measure consumed time of a callable object to be completed.

- `tests/prototypes/prototype_mini.py`
The runner for workflow prototype (`PrototypeRunner`) is in this file.

- `tests/prototypes/prototpye_test.py`
It has benchmark session object generator as a fixture and use ``PrototypeRunner`` to run a prototype.

## Benchmark result example
Here is the benchmark result saved under `.benchmarks` from the command `pytest --benchmark-test`.

```json
{
  "environment": {
    "benchmark_run_id": "1bf01bbc31d44c7998cc0b401ce8aba4",
    "git_commit_id": "7aa6d01eab0b8b1e44657e77e58fa47246d69b5b",
    "timestamp": "2023-11-06T09:25:47",
    "hardware_spec": {
      "operating_system": "Linux",
      "operating_system_version": "#34~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Sep  7 13:12:03 UTC 2",
      "platform_desc": "Linux-6.2.0-34-generic-x86_64-with-glibc2.35",
      "machine_type": "x86_64",
      "total_memory": [134, "GB"]
    }
  },
  "target_names": ["mini_prototype", "mini_prototype", "mini_prototype"],
  "measurements": {
    "time": {
      "value": [0.0913536548614502, 0.08919787406921387, 0.09025907516479492],
      "unit": ["s", "s", "s"]
    },
    "space": {
      "value": [null, null, null],
      "unit": [null, null, null]
    }
  },
  "arguments": {
    "num_pixels": [10000, 10000, 10000],
    "event_rate": [10000, 10000, 10000],
    "histogram_bin_size": [50, 50, 50],
    "chunk_size": [28, 28, 28],
    "frame_rate": [14, 14, 14],
    "num_frames": [140, 140, 140],
    "random_seed": [123, 123, 123],
    "test-run": [true, true, true]
  }
}

```

### DIFF
--- a/tests/benchmarks/runner.py
+++ b/tests/benchmarks/runner.py
@@ -1,0 +1,262 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from abc import ABC, abstractmethod
+from contextlib import contextmanager
+from dataclasses import asdict, dataclass
+from typing import Callable, Generic, NewType, Optional, TypeVar
+
+import scipp as sc
+
+from beamlime.constructors import Factory, ProviderGroup
+
+from .environments import (
+    BenchmarkEnvironment,
+    BenchmarkResultFilePath,
+    BenchmarkTargetName,
+    env_providers,
+)
+
+
+@dataclass
+class TimeMeasurement:
+    """Time measurement results."""
+
+    value: float
+    unit: str
+
+
+@dataclass
+class SpaceMeasurement:
+    """Memory measurement results."""
+
+    value: float
+    unit: str
+
+
+@dataclass
+class BenchmarkResult:  # Measurement results should always have value and unit.
+    time: TimeMeasurement
+    space: Optional[SpaceMeasurement] = None
+
+
+_Item = TypeVar("_Item")
+
+
+def _append_row(
+    obj: dict[str, list[Optional[_Item]]], row: dict[str, _Item]
+) -> dict[str, list[Optional[_Item]]]:
+    """
+    Helper function to extend Pandas Dataframe-like dictionary.
+    All columns(Corresponding to the highest level keys)
+    in the ``obj`` should always have the same length of rows.
+    """
+    # Reference count of rows.
+    _ref = min(len(v) for _, v in obj.items()) if obj else 0
+
+    for missing_k in (k for k in obj.keys() if k not in row):
+        # Append ``None`` if there is an existing key
+        # that does not exist in ``row``.
+        obj[missing_k].append(None)
+
+    for k, v in row.items():
+        # Fills with ``[None]*_ref`` if there is a new key
+        # that does not exist in the ``obj``.
+        obj.setdefault(k, [None] * _ref).append(v)
+
+    return obj
+
+
+R = TypeVar("R")
+
+
+@dataclass
+class SingleRunReport(Generic[R]):
+    callable_name: BenchmarkTargetName
+    benchmark_result: BenchmarkResult
+    arguments: dict
+    output: Optional[R] = None
+
+
+@dataclass  # Need dataclass decorator to use ``as_dict``.
+class BenchmarkReport:
+    """Benchmark report template."""
+
+    environment: BenchmarkEnvironment
+    target_names: list[BenchmarkTargetName]
+    measurements: dict[str, dict[str, list]]
+    arguments: dict[str, list]
+
+    def __init__(self) -> None:
+        self.target_names = list()
+        self.measurements = dict()
+        self.arguments = dict()
+
+    def append_measurement(self, result: BenchmarkResult) -> None:
+        measurement = asdict(result)
+        for meas_dim in measurement:
+            _append_row(
+                self.measurements.setdefault(meas_dim, dict(value=[], unit=[])),
+                measurement.get(meas_dim) or dict(value=None, unit=None),
+            )
+
+    def append(self, single_run_result: SingleRunReport) -> None:
+        self.target_names.append(single_run_result.callable_name)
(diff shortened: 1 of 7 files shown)


SimonHeybrock (commented):
Code looks clean (I have not looked at all details). Are there instructions on how to configure and use this?

SimonHeybrock:
Can you add some documenation, or a readme file, explaining how to use all this? The docstrings are useful, but do not give on overall idea (and it will be hard to figure out where to look first).

YooSunYoung:
Loader&visualization helpers will come in another PR!

docs/developer/benchmark.ipynb
```diff
@@ -0,0 +1,587 @@
...
+   "execution_count": null,
+   "id": "db5d1f7c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from rich.pretty import Pretty\n",
+    "from tests.benchmarks.runner import create_benchmark_runner_factory, BenchmarkSession\n",
+    "\n",
+    "benchmark_factory = create_benchmark_runner_factory()\n",
+    "session = benchmark_factory[BenchmarkSession]\n",
```
SimonHeybrock, line 64: Can you start by briefly outlining (above) what the factory, the session, and the runner are and do?

docs/developer/benchmark.ipynb
```diff
@@ -0,0 +1,587 @@
...
+    "session.run(test_func, 0.1)\n",
+    "Pretty(session.report.measurements)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7b63a66a",
+   "metadata": {},
+   "source": [
+    "The ``BenchmarkSession.run`` passes all arguments to ``runner`` and append its result into the ``report``.\n",
```
SimonHeybrock, line 98: I can't see `runner` in the code, what are you referring to? Same for `report`.

docs/developer/benchmark.ipynb
```diff
@@ -0,0 +1,587 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "26671cfc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from tests.benchmarks.runner import BenchmarkRunner\n",
+    "runner = benchmark_factory[BenchmarkRunner]  # SimpleRunner\n",
```
SimonHeybrock, line 151: So `benchmark_factory` is not a factory for benchmarks, but a factory for sessions and runners? Why is it the same factory? What does it do?

tests/benchmarks/runner.py
```diff
@@ -0,0 +1,242 @@
...
+def create_benchmark_runner_factory(
+    runner_type: type[BenchmarkRunner] = SimpleRunner,
+) -> Factory:
+    session_providers = ProviderGroup()
+    session_providers[BenchmarkSession] = BenchmarkSession
+    session_providers[BenchmarkReport] = BenchmarkReport
+    session_providers[BenchmarkRunner] = runner_type
+    session_providers[BenchmarkFileManager] = SimpleFileManager
+    session_providers[BenchmarkSessionConfiguration] = BenchmarkSessionConfiguration
+    return Factory(env_providers, session_providers)
```
SimonHeybrock, line 242: Can you explain what this is doing?

SimonHeybrock (commented):
The notebook helps a lot :+1: 

Some questions:

tests/benchmarks/runner.py
YooSunYoung: It is just a helper to collect all providers to build ``BenchmarkSession`` with dependency injection module.
I also added a doc string.

docs/developer/benchmark.ipynb
YooSunYoung: No, sorry for confusing names... This factory builds a benchmark session object not the benchmarks.
There is another factory to build a prototype.
I renamed it to have clear meaning.

docs/developer/benchmark.ipynb
YooSunYoung: I rephrased them to be consistent with the code blocks... sorry...!

docs/developer/benchmark.ipynb
YooSunYoung: Yes, I added the outline above.

SimonHeybrock (approved):

PR #117, opened 2024-01-09
Logos and icons

YooSunYoung:
The logo works both in the bright background and dark background so dark logo is just as same as the original logo.

## Static Files
- favicon.ico
- logo.svg
- logo-dark.svg
## Scripts
- draw_icon.py
- draw_logo.py


docs/_static/draw_icon.py
jl-wynen, line 1: Did you check whether these scripts get deployed to the built docs? They might because they are in `_static`.

docs/_static/favicon.ico
jl-wynen, line 1: You can alternatively use an svg. There is no need any more for converting to ico.

docs/_static/draw_icon.py
YooSunYoung: Yes...! I'll move it to `resources` instead.

docs/_static/favicon.ico
jl-wynen: Yes. Do we have something like this in the template?

jl-wynen (approved):

docs/_static/favicon.ico
YooSunYoung: No, but maybe it should go into the release process instruction.

PR #118, opened 2024-01-10
Document benchmark results with visualization helper.

YooSunYoung:
I added visualizing helper and added documentation page to collect the results.

It only contains the representing result with 63 cpu cores with 128 GB memory.

Other results will be added in a separate PR.

TODO:
- [ ] Add plot of performance comparing between just workflow and beamlime


YooSunYoung:
Closing this one in favor of #122

PR #125, opened 2024-02-12
Add overhead benchmark test result in docs.

### LINKED ISSUE #38: Performance tests

YooSunYoung:
Fixes #38 

Probably last performance plot in the docs...!
The test results are collected via ``workflow_test.py``.

Overhead performance plot:

output


SimonHeybrock:
What is the meaning of "online", "offline", "standard speed", and "speed"?

YooSunYoung:
> What is the meaning of "online", "offline" in relation to "standard speed", and "speed"?

It's either if we processed them chunk by chunk through smaller steps, or just run the workflow directly on the whole set of data.
There's an same explanation in the markdown cell. Is it too ambiguous....?

> A negative percentage is hard to think about. Is there a better quantity we could plot, which would tell us more about the actual impact?

I just kept it consistent with other plots because it was the least confusing if you already understood what other plots above are.
But I could plot the difference of the `time/frame` it consumed instead of speed if that's more intuitive...?

Something like this...?

time_per_frame_comparison


> Furthermore, this plot (choice of quantity being plottet) appears to highlight the irrelevant parts (large overhead for tiny data?), but the more interesting part (overhead for large data) is hidden close to 0.

I thought large overhead for tiny data is also meaningful.
Then I can just make them into separate plots.

separate_comparison


like this...??

SimonHeybrock:
> It's either if we processed them chunk by chunk through smaller steps, or just run the workflow directly on the whole set of data.
> There's an same explanation in the markdown cell. Is it too ambiguous....?

That was defining online and offline, but it was not clear to me which of the two (if any) is "standard speed" and which is "speed".

SimonHeybrock:
> I thought large overhead for tiny data is also meaningful.

How about plotting the overhead in seconds per frame? Is this number defined based on your measurement?

YooSunYoung:
> > I thought large overhead for tiny data is also meaningful.
> 
> How about plotting the overhead in seconds per frame? Is this number defined based on your measurement?

Like this...?

process_time_per_frame

SimonHeybrock:
Not sure I understand the plots correctly (what is on the Y axis?), but this suddenly looks like there is a massive overhead for 107 pixels or more? Didn't you previously determine that the overhead from Beamlime is negligible?

YooSunYoung:
> Not sure I understand the plots correctly (what is on the Y axis?), but this suddenly looks like there is a massive overhead for 107 pixels or more? Didn't you previously determine that the overhead from Beamlime is negligible?

It's because the ``offline`` reduction is done on the whole dataset, not per chunk(28 frames in this case).
If we iterate 28 frames at once just like what ``online`` does,
the result is like this:

onoff_chunk_process_time

Here you can see the the orange arrow and green yellow are mostly on top of each other.

SimonHeybrock:
> It's because the offline reduction is done on the whole dataset, not per chunk(28 frames in this case).

Then this benchmark cannot tell us anything about the overhead caused by Beamlime. We must compare the same operations on the same data.

YooSunYoung:
Final version of the plot (maybe...?)

onoffworkflow

SimonHeybrock:
What is `s/counts`? `s/chunk` or something else?

YooSunYoung:
> What is `s/counts`? `s/chunk` or something else?

Hmm... it's meant to be per 1 frame, which is 1 pulse in the experiment.
I'll just show the total time and mention the number of pulses it processed like this:

onoff_chunk_process_time

YooSunYoung:
I'll close it for now since we're almost re-writing everything : D

PR #144, opened 2024-02-23
Refactoring prototypes (without kafka yet).

### LINKED ISSUE #130: Move prototype to `src/beamlime` and add executable scipt
- Move prototype to `src`
  - Add option to choose input source (selecting Kafka vs. fake)
- Add script (see https://github.com/scipp/chexus/blob/f843830d1966aa2468b41c42a6f5b0e198c2de4d/pyproject.toml#L38-L39 for example).

YooSunYoung:
Fixes #130 

## Try out
```bash
beamlime --image-path beamlime_plot.png
```

This command will save the plot into ``beamlime_plot.png``.

### Core
- Application components concept, ``handler``, ``daemon`` and ``application``.
  - In the previous prototypes in ``test`` helper package, all steps were implemented as daemons.
  - They are still just concept, and doesn't have a concrete low-level helper modules or a framework yet.
  - I just tried using those terms for easier communication among us within the project.
- Multiple temporary/constant providers setting context manager moved from helper package to ``constructors`` module.

### Pre-made handlers and daemons
- Random event data generator (no log or start-message)
- Dummy workflow (Wavelength converting and histogramming, uses some hard-coded values)
- Plotting handler that saves a png file.
- MessageRouter daemon 
  - It's not a lower-level message routing framework that I tried to implement here. It's just one of pre-made daemons that took some of works that ``DataReductionApp`` was doing.
- Workflow handler that runs the dummy workflow on the chunk of data.
- ``DataReductionApp`` as a highest-level interface.

### Documentation
- Jupyter and command line examples [docs/getting-started/live_data_reduction.ipynb](https://github.com/scipp/beamlime/pull/144/files#diff-87e5bab61b7d085b7fc70079c284ea6250291293a16bf97fadb971e255dc04b2)

## Related next tasks

TODO
- [x] Update docstrings
- [x] Update documentation and write more explanations  (Maybe briefly, mostly about future implementation from here?)
- [x] Update/add tests if needed(?)


YooSunYoung:
Currently looks like this:
[Screencast from 02-26-2024 09:35:57 AM.webm](https://github.com/scipp/beamlime/assets/17974113/24c9cf45-e588-4802-858e-87a4f2b0469f)

SimonHeybrock:
Unless there are essential to-dos, we should get the ready as quickly as possible (do the rest in later PRs), since it will simplify coordination for follow-up tasks. Moving the Kafka bits is not required yet.

YooSunYoung:
> Unless there are essential to-dos, we should get this ready as quickly as possible (do the rest in later PRs), since it will simplify coordination for follow-up tasks. 

@SimonHeybrock 
Yes...! I tried to extract the bare minimum and make it as clear as possible, what each part of code does...

> Moving the Kafka bits is not required yet.

Yeah, that's why I didn't include it here.

src/beamlime/applications/base.py
```diff
@@ -0,0 +1,176 @@
...
+            await self.route(message)
+        self.debug("Routing the rest of the message done.")
+
+    async def send_message_async(self, message: BeamlimeMessage) -> None:
+        self.message_pipe.append(message)
+        await asyncio.sleep(0)
+
+
+class BaseHandler(LogMixin, ABC):
+    """Base class for message handlers.
```
SimonHeybrock, line 158: Above it says daemons handle messages.

src/beamlime/applications/base.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Message handlers are expected to register methods to a message router,
+    so that the message router can trigger the methods
+    with the certain type of message.
+
+    Registered methods should accept a message as the only positional argument.
```
SimonHeybrock, line 164: This seems to be a new mechanism? Not sure why this was added now? I suppose it has not been benchmarked? Is this required, given the quite static nature of Beamlime, which will always have the same components?

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,199 @@
...
+
+class DataReductionMessageRouter(MessageRouter):
+    async def run(self) -> None:
+        await super().run()
+        await self.route(
+            StopWatch.Stop(
+                sender=DataReductionMessageRouter,
+                receiver=StopWatch,
+                content=None,
+            )
```
SimonHeybrock, line 22: The message router is generating messages? That seems odd. Shouldn't it route messages?

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,199 @@
...
+                        receiver=StopWatch,
+                        content=None,
+                    )
+                )
+
+        await self.messenger.send_message_async(
+            self.messenger.StopRouting(
+                sender=DataStreamSimulator, receiver=self.messenger, content=None
+            )
+        )
```
SimonHeybrock, line 74: Seems unrelated to the business of a stream simulator. Remove?

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,199 @@
...
+DataStreamListener = NewType("DataStreamListener", BaseDaemon)
+PlotHandler = NewType("PlotHandler", BaseHandler)
+WorkflowHandler = NewType("WorkflowHandler", BaseHandler)
+
+
+class DataReductionDaemon(BaseDaemon, ABC):
+    """Data reduction daemon that handles all part of the data reduction process."""
+
+    # Daemons
+    data_stream_listener: DataStreamListener
```
SimonHeybrock, line 88: Why does this daemon contain another daemon?

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,199 @@
...
+    # Handlers
+    data_reduction_handler: WorkflowHandler
+    plot_handler: PlotHandler
+    stop_watch: StopWatch
+
+    def collect_sub_daemons(self) -> list[BaseDaemon]:
+        return [
+            self.data_stream_listener,
+            self.message_router,
+        ]
```
SimonHeybrock, line 99: Why do we need sub daemons?

src/beamlime/executables/prototypes.py
```diff
@@ -0,0 +1,124 @@
...
+):
+    from beamlime.constructors import (
+        multiple_constant_providers,
+        multiple_temporary_providers,
+    )
+
+    tmp_factory = Factory(prototype_factory.providers)
+    with multiple_constant_providers(tmp_factory, parameters):
+        with multiple_temporary_providers(tmp_factory, providers):
+            yield tmp_factory
```
SimonHeybrock, line 31: Why do we need this?

tests/prototypes/prototype_test.py
```diff
@@ -1,40 +1,100 @@
...
 
-from beamlime.constructors import Factory
-from tests.benchmarks.runner import BenchmarkSession
-from tests.prototypes.parameters import EventRate, NumPixels, PrototypeParameters
-from tests.prototypes.prototype_mini import (
-    BenchmarkTargetName,
-    PrototypeBenchmarkRecipe,
-    mini_prototype_factory,
-)
+from beamlime.applications._parameters import EventRate, NumPixels, PrototypeParameters
```
SimonHeybrock, line 9: We should probably just remove most of these tests, since most of this will be handled completely differently. Otherwise we will waste endless time refactoring these tests again and again.

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,199 @@
...
+            if not loop.is_running():
+                loop.run_until_complete(asyncio.gather(*self.tasks))
+
+
+class Prototype(DataReductionDaemon):
+    """Prototype of the data reduction daemon."""
+
+    data_stream_listener: DataStreamSimulator
+    plot_handler: PlotSaver
+    data_reduction_handler: DataReductionHandler
```
SimonHeybrock, line 168: Contained in base already.

src/beamlime/applications/handlers.py
```diff
@@ -0,0 +1,173 @@
...
+
+        self.output_da.values += output.values
+        self.stream_node.notify_children("update")
+        await self.messenger.send_message_async(
+            HistogramUpdated(
+                sender=DataReductionHandler,
+                receiver=Any,
+                content=self.figure,
+            )
+        )
```
SimonHeybrock, line 126: To be honest, I do not get much of the entire daemon/handler thing. Daemons send messages. Handlers send messages. Daemons contain daemons. Daemons contain handlers, but don't do anything with them. Handlers send and process messages that are not their concern. Etc., ...

I cannot tell why the (seemingly?) high complexity and (seemingly high?) coupling between the various components is needed. I thought I roughly understood how things work when I reviewed the prototypes (being "moved" here), but now I am honestly lost again.

I am not sure how to proceed. Proceed and then trim down in subsequent steps until only essentials are left (and refactor to remove coupling)?

src/beamlime/applications/base.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def __init__(self, message_pipe: List[BeamlimeMessage]):
+        self.handlers: dict[type, List[Callable[[BeamlimeMessage], Any]]] = dict()
+        self.awaitable_handlers: dict[
+            type, List[Callable[[BeamlimeMessage], Awaitable[Any]]]
+        ] = dict()
+        self.message_pipe = message_pipe
+        self._break_routing_loop = False
+
+        class StopRouting(BeamlimeMessage):
```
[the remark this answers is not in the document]
YooSunYoung: The original one was using ``while-loop`` that breaks based on the ``StopWatch``
because it was mainly for benchmarking tests.
``MessageRouter`` was inserted to replace that.
I didn't have better idea than this... if it's too confusing I can remove it and go back to the original way.

src/beamlime/applications/base.py
YooSunYoung: What I tried to implement was:
Daemons and handlers both handle messages
but the daemons have routine that runs until it's done,
and handlers only have methods that can handle certain type of messages.

src/beamlime/executables/prototypes.py
YooSunYoung: It was to build application instances easily from a factory without affecting existing one.

src/beamlime/applications/handlers.py
YooSunYoung: Sorry for confusing... 
I renamed this class to ``Application``.
So the idea was to have both ``daemons`` and ``handlers`` under the ``Application``.
``daemon``  should have a ``run`` method, that can be tasks in the event loop and
``handler`` should have a method that can be called for handling certain type of event/message.

src/beamlime/applications/daemons.py
YooSunYoung: It's for overwriting the type of those properties.
The ``plot_handler`` and ``DataStreamListener`` can be changed.

tests/prototypes/prototype_test.py
YooSunYoung: I just removed all the tests for now.
I'll add tests later if needed.

src/beamlime/applications/base.py
YooSunYoung: But it's basically doing the same thing, and makes much less coroutine calls,
because in the original one, all daemons had to monitor their own lists,
but now it's only one instance that monitors the queue...

pyproject.toml
```diff
@@ -41,6 +41,9 @@ dependencies = [
```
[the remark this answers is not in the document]
YooSunYoung: I think so...?
| packages | role |
| -- | -- |
| "ess-streaming-data-types" | Flat-buffer (de)serialising |
| "numpy", "plopp",  "sciline", "scipp>23.08.0", "scippneutron" | Data reduction example |
| "rich" | logging |

YooSunYoung:
I tried to refactor it in a more clear way...

```mermaid
classDiagram
    HandlerInterface <|-- DataReductionHandler
    HandlerInterface <|-- PlotSaver

    DaemonInterface <|-- DataStreamSimulator
    DaemonInterface <|-- MessageRouter

    ApplicationInterface <|-- DataReductionApp

    class HandlerInterface {
        <<interface>>
    }
    class DataReductionHandler {
        process_message()
    }
    class PlotSaver {
        save_histogram()
    }
    class DaemonInterface{
        <<interface>>
        async run()
    }
    class DataStreamSimulator
    class MessageRouter {
        async route()
        async send_message_async()
    }
    class ApplicationInterface{
            <<interface>>
            daemons: tuple[DaemonInterface]
            tasks: list[asyncio.Task]
            loop: asyncio.EventLoop
            register_handlers()            
            run()
        }
        class DataReductionApp {
            data_stream_listener: DataStreamSimulator
            message_router: MessageRouter
            data_reduction_handler: DataReductionHandler
            plot_handler: PlotSaver
        }
```

And I also tried to run a benchmark with this implementation and it looks very similar:
``online-5-chunk(s)-old`` is the previous implementation, and the ``online-5-chunk(s)-new`` is the new one.
Both of ``offline-*`` are also collected again so it was using same pipeline as ``online-5-chunk(s)-new``.

benchmark

src/beamlime/applications/daemons.py
YooSunYoung: I removed one layer of inheritance for now.

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,173 @@
...
+from ._random_data_providers import RandomEvents
+from ._workflow import Events
+from .base import BeamlimeMessage, DaemonInterface
+from .handlers import RawDataSent
+
+
+class MessageRouter(DaemonInterface):
+    """A message router that routes messages to handlers."""
+
+    message_pipe: List[BeamlimeMessage]
```
jl-wynen, line 17: Shouldn't this be a queue? (https://docs.python.org/3/library/queue.html)
Also, you don't need a class annotation for it.

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,173 @@
...
+        if isinstance(result, BeamlimeMessage):
+            self.message_pipe.append(result)
+        elif isinstance(result, tuple):
+            msgs = (_msg for _msg in result if isinstance(_msg, BeamlimeMessage))
+            self.message_pipe.extend(msgs)
+
+    async def route(self, message: BeamlimeMessage) -> None:
+        # Synchronous handlers
+        if handlers := self.handlers.get(type(message), []):
+            for handler in handlers:
```
jl-wynen, line 89: Useless `if`: (same below)
```suggestion
        for handler in self.handlers.get(type(message), []):
```

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,173 @@
...
+                await asyncio.sleep(0)  # Let others use the event loop.
+                with self._handler_wrapper(handler, message):
+                    self._check_result(handler(message))
+
+        # Asynchronous handlers
+        if awaitable_handlers := self.awaitable_handlers.get(type(message), []):
+            for handler in awaitable_handlers:
+                with self._handler_wrapper(handler, message):
+                    self._check_result(await handler(message))
```
jl-wynen, line 98: This means that you run the handlers in sequence, i.e., they cannot run in parallel. It would be better to spawn them all at the same time.

src/beamlime/applications/daemons.py
```diff
@@ -0,0 +1,173 @@
...
+    @contextmanager
+    def _handler_wrapper(
+        self, handler: Callable[..., Any], message: BeamlimeMessage
+    ) -> Generator[None, None, None]:
+        import warnings
+
+        try:
+            self.debug(f"Routing event {type(message)} to handler {handler}...")
+            yield
+        except Exception as e:
```
jl-wynen, line 46: This ultimately ignores the error. So it could end up running the handler at every update cycle even though it fails every time. Intentional?

src/beamlime/applications/handlers.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class DataReductionHandler(HandlerInterface):
+    """Data reduction handler to process the raw data and update the histogram.
+
+    It receives a list of events, and reduces them into a histogram.
+    It also triggers the update of a plot stream node.
+    """
+
+    pipeline: WorkflowPipeline
```
jl-wynen, line 33: Where does `pipeline` get initialised? I don't see it in the class.

src/beamlime/applications/handlers.py
```diff
@@ -0,0 +1,114 @@
...
+            grid=True,
+        )
+
+    async def process_data(self, data: Events) -> Histogrammed:
+        self.info("Received, %s", self.format_received(data))
+        self.pipeline[Events] = data
+        return self.pipeline.compute(Histogrammed)
+
+    async def process_message(
+        self, message: RawDataSent | BeamlimeMessage
```
jl-wynen, line 63: I think the annotation is equivalent to `message: BeamlimeMessage` because `RawDataSent` is a subclass of `BeamlimeMessage`.

jl-wynen (commented):
I think the overall architecture makes sense. But I'm confused about `ApplicationInterface`. Why does it need a base class? Does any part of the code depend on the application?

YooSunYoung:
> I think the overall architecture makes sense. But I'm confused about ApplicationInterface. Why does it need a base class? Does any part of the code depend on the application?

That is true... what I really wanted is, to allow customizing daemons and handlers.

src/beamlime/applications/handlers.py
YooSunYoung: It's assembled by ``Factory``.
The provider is here:
https://github.com/scipp/beamlime/blob/c0264d838e02e73ba977644bc31a0dde324c77eb/src/beamlime/applications/_workflow.py#L134

src/beamlime/applications/handlers.py
YooSunYoung: I wanted to annotate it as ``RawDataSet`` so that it's easy to see that this method is meant to handle ``RawDataSent`` and then pylint was not happy about it...
I'll just keep it as ``BeamlimeMessage`` for now.

jl-wynen:
> > I think the overall architecture makes sense. But I'm confused about ApplicationInterface. Why does it need a base class? Does any part of the code depend on the application?
> 
> That is true... what I really wanted is, to allow customizing daemons and handlers.

In the current implementation, that would require writing a completely new application class, no? So the interface doesn't accomplish much.

src/beamlime/applications/daemons.py
YooSunYoung: That is true....!

src/beamlime/applications/daemons.py
YooSunYoung: It was intentional but I think it's not a good idea... changing...

YooSunYoung:
> > > I think the overall architecture makes sense. But I'm confused about ApplicationInterface. Why does it need a base class? Does any part of the code depend on the application?
> > 
> > 
> > That is true... what I really wanted is, to allow customizing daemons and handlers.
> 
> In the current implementation, that would require writing a completely new application class, no? So the interface doesn't accomplish much.

No, not really... you just need to overwrite ``daemons`` property or ``register_handling_methods``.
then you can use ``run`` method as it is, in the interface class.

I made it this way so that you can just type-hint the class attribute
so that ``Factory`` can just inject those properties.
But I had a second thought about this... I think it should have a method that allows to register handling methods and daemons into this application instance...

src/beamlime/applications/handlers.py
jl-wynen: Can you point to the place where the attribute gets set?

Does this mean that the handler is in an invalid state until it gets patched up by the factory later?

jl-wynen:
> But I had a second thought about this... I think it should have a method that allows to register handling methods and daemons into this application instance...

Yes, try using composition instead of inheritance.

src/beamlime/applications/handlers.py
YooSunYoung: Ah...! I understood what you meant. It seems wrong...
I'll move it to the constructor instead.

> Does this mean that the handler is in an invalid state until it gets patched up by the factory later?

And yes...! It's invalid if it's not patched... I'll change...

YooSunYoung:
Now it looks like this...!



```mermaid
classDiagram
    HandlerInterface <|-- DataReductionHandler
    HandlerInterface <|-- PlotSaver

    DaemonInterface <|-- DataStreamSimulator
    DaemonInterface <|-- MessageRouter

    class HandlerInterface {
        <<interface>>
    }
    class DataReductionHandler {
        process_message()
    }
    class PlotSaver {
        save_histogram()
    }
    class DaemonInterface{
        <<interface>>
        async run()
    }
    class DataStreamSimulator
    class MessageRouter {
        async route()
        async send_message_async()
    }
    class Application {
        daemons: list[DaemonInterface]
        tasks: list[asyncio.Task]
        loop: asyncio.EventLoop
        register_handling_method()
        register_daemon()
        run()
    }
```

And instead of ``Application`` registering handlers and daemons itself,
it should be done after building ``Application`` instance.
So that ``Application`` class doesn't have to know anything about ``handlers`` and ``daemons`` from the beginning.

So collecting providers is also separated from the ``Application``.

pyproject.toml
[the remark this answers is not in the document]
SimonHeybrock: Ultimately, I think dependencies that the data-reduction needs should (and must) be dependencies of the package implementing a workflow. It should not be Beamlime?

src/beamlime/applications/base.py
SimonHeybrock: So handlers run only once? And daemons handle all messages?

src/beamlime/applications/handlers.py
```diff
@@ -0,0 +1,115 @@
...
+    async def process_message(self, message: MessageProtocol) -> MessageProtocol:
+        if not isinstance(message, RawDataSent):
+            raise TypeError(f"Message type should be {RawDataSent.__name__}.")
+
+        if not hasattr(self, "output_da"):
+            self.process_first_input(message.content)
+            output = await self.process_data(message.content)
+            self.process_first_output(output)
+        else:
+            output = await self.process_data(message.content)
```
SimonHeybrock, line 74: Can you remind me why have have the nested `await/async`? Why is `process_data` `async`, and not just `process_message`?

src/beamlime/applications/handlers.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def process_first_output(self, data: Histogrammed) -> None:
+        self.output_da = Histogrammed(sc.zeros_like(data))
+        self.info("First data as a seed of histogram: %s", self.output_da)
+        self.stream_node = pp.Node(self.output_da)
+        self.figure = pp.figure1d(
+            self.stream_node,
+            title="Wavelength Histogram",
+            grid=True,
+        )
```
SimonHeybrock, line 58: Can we move the plotting code out of the data-reduction handler? Instead, put it into the `PlotSaver`? This coupling of the processing with the visualization adds coupling.

SimonHeybrock (commented):
I think it looks much better and simpler now 👍. Not an in-depth review (I think Jan-Lukas did that), but two more questions:

src/beamlime/applications/handlers.py
YooSunYoung: Hmm... just copied-pasted without thinking... I think it's better to keep methods/functions non-async as much as possible.

jokasimr:
> And I also tried to run a benchmark with this implementation and it looks very similar: `online-5-chunk(s)-old` is the previous implementation, and the `online-5-chunk(s)-new` is the new one. Both of `offline-*` are also collected again so it was using same pipeline as `online-5-chunk(s)-new`.
> 
> benchmark

Is there a reason for using a log-lin scale here instead of log-log? I imagine we expect the time to be linear in the number of events, and in that case a log-log plot illustrates that relationship better.

pyproject.toml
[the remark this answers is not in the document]
YooSunYoung: That is quite right... I'll update it in another PR ...!

SimonHeybrock (approved):
I think we can merge this now, moving on to the next tasks in the milestone.

Issue #127, opened 2024-02-13
Event-driven application controlling.


YooSunYoung:
The applications need to be controlled during the run-time, i.e. `kill` or `restart`.
Current prototype applications have time-out and a dedicated flag that they are watching.
For more flexible controlling of those apps, maybe we can use event routers.

For example, kafka brokers or [message bubbling](https://textual.textualize.io/api/message/) and [on method](https://textual.textualize.io/api/on/) in [textual](https://textual.textualize.io).

We could somehow use `textual` or `flask` or... similar third party packages, or implement them ourselves...?

YooSunYoung:
Idea note from #144 

Now ``Daemon`` is sending messages to a router directly.
But, maybe it just have a async-generator that yields ``Message``s as it runs.
And ``Application`` can use the generator and send the message that it yields instead.

YooSunYoung:
It is already implemented as a message router.

PR #146, opened 2024-03-01
DaemonInterface as a generator.

YooSunYoung:
Related to #127 

# Main Changes
- ``DaemonInterface.run`` is no longer just coroutine, but it is now ``AsyncGenerator``.
  It yields either a message or ``None``, so the application knows if it should forward the output of the generator to the router or not.
- Message handling methods use ``try``-``except`` instead of ``hasattr``.

# Pros
- De-coupling ``MessageRouter`` from all daemons.
  Now ``DaemonInterface`` doesn't have to have ``MessageRouter`` as a member to send a message. 
- Better syntax to write a daemonic-process.
  Instead of calling ``message_router.send_message_async``, it can simply ``yield`` messages along the process.
- Easier to control the interval of daemon process all at once.
  ``Application`` can influence how often the daemon has the event loop.
- More clear when/how to call ``asyncio.sleep`` in the daemon.
  For example, since re-calling of the async-generator is done by the ``Application``, daemons no longer need to worry about re-trying and simply add an if statement that calls ``await asyncio.sleep``.
  So helpers such as ``async_retry`` or ``data_pipe_monitor`` is no longer needed.

# Cons?
- Daemons must call the ``asyncio.sleep`` or ``yield`` something in the ``run`` generator.
  Otherwise it will occupy the event loop for quite sometime...
  It was always the case even without this change,
  but you didn't have to think about it with ``data_pipe_monitor`` helper since it is doing it itself.
  You also have to think about the timing of when ``asyncio.sleep`` is called  in the generator.
- AsyncGenerator type-annotation is a bit tedious.
- You need to make a generator to write a test. (One more step for testing.)

# Memo

I also tried making handlers ``AsyncGenerator`` to try using ``asend``.
I wanted to try this to avoid checking attribute to see if it is the very first message to process.
But it was not so neat after all since you always have to ``asend`` ``None``,
and it seems like making things a bit slower.....
So I tried using try-exception instead of checking attributes to save time in the run-time.


src/beamlime/applications/base.py
```diff
@@ -206,25 +149,17 @@ async def route(self, message: MessageProtocol) -> None:
...
-            if await data_monitor():
-                message = self.message_pipe.get()
-                await self.route(message)
-                await asyncio.sleep(0)
-
-        self.debug("Breaking routing loop. Routing the rest of the message...")
-        while not self.message_pipe.empty():
-            await self.route(self.message_pipe.get())
-        self.debug("Routing the rest of the message done.")
+        while True:
```
jl-wynen, line 156: How does the loop get stopped?

src/beamlime/applications/base.py
```diff
@@ -206,25 +149,17 @@ async def route(self, message: MessageProtocol) -> None:
...
         """Message router daemon."""
-        data_monitor = self.data_pipe_monitor(
-            self.message_pipe,
-            timeout=0.1,
-            interval=0.1,
-            size_monitor_func=self.message_pipe.qsize,
-        )
-
-        while not self._break_routing_loop:
-            if await data_monitor():
```
jl-wynen, line 219: So you removed the monitor and the check for the number of messages. Why is this not needed any more?

src/beamlime/applications/base.py
YooSunYoung: This is a generator so once it reaches yield, it's stopps there.

src/beamlime/applications/base.py
YooSunYoung: Because message router doesn't decide when to route/stop anymore.
It's a generator so the ``Application`` instance will call ``next`` on it.

SimonHeybrock:
There are merge conflicts here. Is this PR still relevant, or will it be replaced?

src/beamlime/applications/base.py
jl-wynen: It pauses, but doesn't stop and clean up completely. So does the caller simply drop its reference to the generator?

YooSunYoung:
> There are merge conflicts here. Is this PR still relevant, or will it be replaced?

I'll fix it!

src/beamlime/applications/base.py
YooSunYoung: Ah, now I understood...!

https://github.com/scipp/beamlime/blob/cfa63d08701e5f36f558ca4b7fe93ab1425e8e48/src/beamlime/applications/base.py#L222-L231

This method is used for creating a routine and it has a reference to ``self(Application instance)._break``.

So ``run_daemon`` coroutine is done, the generator should just stop...?

jl-wynen: Ok. It should be fine because `run` doesn't hold onto any resources that need to cleaned up.

src/beamlime/applications/base.py
```diff
@@ -242,24 +177,38 @@ class Application(LogMixin):
 
     """
 
+    @dataclass
+    class Stop:
```
jl-wynen, line 181: I can't find where this class is instantiated. Is it currently unused?

src/beamlime/applications/base.py
```diff
@@ -242,24 +177,38 @@ class Application(LogMixin):
...
+        """A message to break the routing loop."""
+
+        content: Optional[Any]
+        sender: type
+        receiver: type
```
jl-wynen, line 186: Does it need these attributes? The protocol requires them. But does is that necessary? It seems to me that if you have distinct message types that don't inherit from a common base, they should be able to set their own contents freely.

src/beamlime/applications/daemons.py
```diff
@@ -1,63 +1,54 @@
...
 
 
 @dataclass
-class BeamlimeMessage:
-    """A message object that can be exchanged between daemons or handlers."""
-
-    content: Any
-    sender: type = Any
-    receiver: type = Any
+class RawDataSent(MessageProtocol):
```
jl-wynen, line 13: Why does this class inherit from `MessageProtocol`? (`Stop` doesn't. And it's a protocol, not a base class.)

src/beamlime/applications/base.py
YooSunYoung: Yeah... I was not sure about the ``MessageProtocol``, how it should be
so I tried to keep it very general and flexible...

src/beamlime/applications/daemons.py
YooSunYoung: That is true...! I think it's just I forgot to remove the inheritance after turning it into protocol....

src/beamlime/applications/base.py
jl-wynen: Depends on the usage. I see several different approaches that can work. I don't know beamlime well enough to judge which one is best. But using protocols like you did does not seem like the best approach to me.

1. Single message type
```python
class Message:
    content: Any  # contains the actual message
    sender: type
    receiver: type
```
This is useful if there are fields (`sender`, `receiver`) that are needed by the router or all handlers. The message type is then represented by the type of data in `content`.

2. Use a base class
```python
class Message:
    sender: type
    receiver: type

class DataMessage(Message):
    data: DataArray
```
This is also useful if there are fields (`sender`, `receiver`) that are needed by the router or all handlers. If effectively inlines the concrete messages into the object but is otherwise the same as 1. Handlers would inspect the type of the message where they would inspect `content` in 1.

This could also be done with a protocol but that would require repeating the common fields.

3. No commonality
```python
class DataMessage:
    data: DataArray
    sender: type
    receiver: type

class StopMessage:
   ...
```
This is useful if there is no need for common attributes. In that case, handlers anyway have to check the type (or use duck-typing) and only look at the attributes they care about.

YooSunYoung: I would like to keep it as protocol so that the handlers or daemons don't have to depend on beamlime,
probably plugin packages...? Or we can maybe change it later... 
but I did remove the ``sender`` and ``receiver`` for now.

PR #149, opened 2024-03-01
Fake Log-message.

YooSunYoung:
I tried writing an example of Log message in between event-data message...

Hmm...


SimonHeybrock:
> I tried writing an example of Log message in between event-data message...

We should make the NXlog messages matching the run-start message that we will be using. Therefore, we should first decide which one to use, so we can have log messages for all the logs in there.

YooSunYoung:
Closing this and will open another one later.

Issue #97, opened 2023-09-26
Kafka integration test in CI.


YooSunYoung:
For now, kafka related test is excluded by a flag ``kafka-test`` among `pytest` test cases.

Those tests should be automated later once there are more complicated kafka integration test cases.

There are a few ideas about how to run kafka integration test

## Mock Kafka clients
Unit tests with Mocked kafka objects.
Most of interfaces could be unit-tested with mocked object.
See [mocked kafka broker example](https://github.com/scipp/beamlime/blob/legacy/tests/communications/mock_helpers.py)

## Local kafka instance in the CI environment
Similarly what we are doing on the local machine, we can set up a kafka broker in the same machine/virtual environment and use it.
It may require a lot of work to maintain the kafka container/deployment to be consistent with the service version otherwise it may not be sufficient to test with the local kafka broker instance.

## Tester credential
We can set up a dummy user in the cluster with a limited access to topics/read-write permissions and use it for testings.
It might be as easy as setting a secret in the github repository but it may also not be possible for security issue...

MridulS:
Just dropping this here for future use: https://www.confluent.io/blog/apache-kafka-ci-cd-with-github/

PR #150, opened 2024-03-04
Add stateless workflow interface

### LINKED ISSUE #145: Define stateless workflow interface
In #131 we ran into a number of complications that might not be resolvable in a short time frame without trying it out in practice. Therefore, I think we should split the problem into two steps:

1. Define a minimal stateless interface (now). This will allow for milestone completion as well as continuation of follow-up work.
2. Gather experience and data in practice, so we can make a more informed choice in #131, for a full workflow interface.

The stateless workflow interface should:

- Take the combined static + time-dependent data as input (from #134) as JSON NeXus with wrapper so it can use ScippNexus to load.
  - Note: In #134 we had assumed that the static- and time-dependent pieces would be "loaded" independently. From the perspective of the stateless interface here we should merge before "loading", and pass the resulting JSON to the workflow.
  - See https://github.com/scipp/scippneutron/blob/22ee6ebe3598d499b1a140bb6cd3aacb33732625/src/scippneutron/io/nexus/load_nexus.py#L400-L414, code like this should be part of Beamlime, not the workflow, essentially we will be passing a `JSONGroup` to the workflow. The loaders in reduction workflows thus need to be able to accept an "open" file/group (currently the open files themselves).
- Return result for this chunk, without any accumulation or other options. That is, we will not perform accumulation in this version but simply display the result of the latest chunk.

I think we can simply make this a function for now, which would also naturally emphasize the stateless nature.

SimonHeybrock:
Fixes #145.


src/beamlime/applications/handlers.py
```diff
@@ -71,51 +53,43 @@ def process_message(self, message: MessageProtocol) -> MessageProtocol:
 def random_image_path() -> ImagePath:
     import uuid
 
-    return ImagePath(pathlib.Path(f"beamlime_plot_{uuid.uuid4().hex}.png"))
+    return ImagePath(pathlib.Path(f"beamlime_plot_{uuid.uuid4().hex}"))
```
YooSunYoung, line 56: Can we change the name ``ImagePath`` to ``ImageNamePrefix``?
And together with ``--image-path`` argument parsing in ``executables/prototypes.py``, line #97.

YooSunYoung (approved):

src/beamlime/applications/handlers.py
SimonHeybrock: I think this may be removed soon anyway, if not, fix it when touching this code next.

PR #151, opened 2024-03-04
Workflow plugins

### LINKED ISSUE #132: Workflow plugin architecture
- Make workflows loadable as plugins.
- See https://github.com/ArjanCodes/2021-plugin-architecture for an example.

jokasimr:
Fixes #132 

Will rebase on main once [stateless-workflow](https://github.com/scipp/beamlime/tree/stateless-workflow) is merged

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,3 +89,6 @@ enable_error_code = [
 ]
 show_error_codes = true
 warn_unreachable = true
+
+[project.entry-points."beamlime.stateless"]
+dummy = "beamlime.stateless_workflow:dummy_workflow"
--- a/src/beamlime/executables/options.py
+++ b/src/beamlime/executables/options.py
@@ -8,6 +8,12 @@ from typing import Optional
 
 def build_arg_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="BEAMLIME configuration.")
+    parser.add_argument(
+        "--workflow",
+        default="dummy",
+        help="Name of the workflow to run",
+        type=str,
+    )
     return parser
 
 
@@ -70,11 +76,4 @@ def visualization_arg_parser(
         help="Set logging level. Default is INFO.",
         type=str,
     )
-    group.add_argument(
-        "--workflow-name",
-        default="dummy",
-        help="Name of the workflow that produces the visualization",
-        type=str,
-    )
-
     return parser
--- a/src/beamlime/executables/options.py
+++ b/src/beamlime/executables/options.py
@@ -70,5 +70,11 @@ def visualization_arg_parser(
         help="Set logging level. Default is INFO.",
         type=str,
     )
+    group.add_argument(
+        "--workflow-name",
+        default="dummy",
+        help="Name of the workflow that produces the visualization",
+        type=str,
+    )
 
     return parser
--- a/src/beamlime/stateless_workflow.py
+++ b/src/beamlime/stateless_workflow.py
@@ -1,9 +1,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
-from typing import Protocol
+from importlib.metadata import entry_points
+from typing import NewType, Protocol
 
 import scipp as sc
 from scippneutron.io.nexus.load_nexus import JSONGroup
+
+Workflow = NewType('Workflow', str)
+'''Name of the workflow plugin to use'''
 
 WorkflowResult = dict[str, sc.DataArray]
 
@@ -33,6 +37,9 @@ class DummyWorkflow:
         }
 
 
-def provide_stateless_workflow() -> StatelessWorkflow:
-    # TODO implement plugin mechanism here, see #132
-    return DummyWorkflow()
+def provide_stateless_workflow(workflow: Workflow) -> StatelessWorkflow:
+    stateless_workflows = entry_points(group='beamlime.stateless')
+    return stateless_workflows[workflow].load()
+
+
+dummy_workflow = DummyWorkflow()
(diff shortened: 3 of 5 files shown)


src/beamlime/executables/options.py
```diff
@@ -70,5 +70,11 @@ def visualization_arg_parser(
         help="Set logging level. Default is INFO.",
         type=str,
     )
+    group.add_argument(
+        "--workflow-name",
+        default="dummy",
+        help="Name of the workflow that produces the visualization",
+        type=str,
+    )
```
YooSunYoung, line 78: I think plugins should be able to add it's own argument group in to the parser.

For example, ``beamlime`` expects plugins to have a function that adds argument group, sth like... ``workflow_arg_parser(parser: argparse.ArgumentParser)`` so that it adds its own argument group into the parser.
So that ``beamlime`` doesn't have to know about the arguments that only those plugins use.

Or did you want to add an argument for the name of the "workflow plugin", like ``esssans``...?

SimonHeybrock:
> Will rebase on main once [stateless-workflow](https://github.com/scipp/beamlime/tree/stateless-workflow) is merged

For the future: Just choose the previous feature branch as base of the PR, GitHub will then update the base branch automatically once the previous feature is merged.

src/beamlime/executables/options.py
SimonHeybrock: > I think plugins should be able to add it's own argument group in to the parser.

We should not do this for now. Configuration handling is not part of this milestone, and neither of the next (but the one after that).

src/beamlime/stateless_workflow.py
```diff
@@ -33,6 +37,7 @@ def __call__(self, group: JSONGroup) -> WorkflowResult:
         }
 
 
-def provide_stateless_workflow() -> StatelessWorkflow:
-    # TODO implement plugin mechanism here, see #132
-    return DummyWorkflow()
+def provide_stateless_workflow(workflow_name: WorkflowName) -> StatelessWorkflow:
+    stateless_workflows = entry_points(group='beamlime.stateless')
+    Workflow = stateless_workflows[workflow_name].load()
+    return Workflow()
```
SimonHeybrock, line 43: I think we should not assume that a workflow is an instance of a type, it might just be a function (the protocol defines that it has to be callable). The instantiation should thus be done by the module defining the entry point, not here. So this should change to 
```suggestion
    return stateless_workflows[workflow_name].load()
```
Alongside that change, maybe rename `WorkflowName` -> `Workflow` (as well as the command line arg), since that may lead to confusion related to this?

pyproject.toml
```diff
@@ -89,3 +89,6 @@ enable_error_code = [
 ]
 show_error_codes = true
 warn_unreachable = true
+
+[project.entry-points."beamlime.stateless"]
+dummy = "beamlime.stateless_workflow:DummyWorkflow"
```
SimonHeybrock, line 94: You probably need to change this now, e.g., to a function that returns and instance. Did you try running `beamlime`, to see if it still works after your latest changes?

jokasimr:
> > Will rebase on main once [stateless-workflow](https://github.com/scipp/beamlime/tree/stateless-workflow) is merged
> 
> For the future: Just choose the previous feature branch as base of the PR, GitHub will then update the base branch automatically once the previous feature is merged.

Yeah much better!

pyproject.toml
jokasimr: No I haven't opened the repo yet ;) but I'll make sure to test it before merge

src/beamlime/executables/options.py
YooSunYoung: Yeah what I was trying to say was,
**If** it is done, it should be done by the plugins.

SimonHeybrock:
```
beamlime: error: unrecognized arguments: --workflow-name dummy
```
@jokasimr What am I doing wrong?

src/beamlime/executables/options.py
```diff
@@ -70,5 +70,11 @@ def visualization_arg_parser(
         help="Set logging level. Default is INFO.",
         type=str,
     )
+    group.add_argument(
+        "--workflow-name",
+        default="dummy",
+        help="Name of the workflow that produces the visualization",
```
SimonHeybrock, line 76: - This should not be part of the `visualization` arg parser, I think?
- Rename to `--workflow`?

SimonHeybrock:
I see, there is duplication of the arg parser bits in `options.py` and `prototypes.py`. @YooSunYoung can we remove one of those?

YooSunYoung:
> I see, there is duplication of the arg parser bits in `options.py` and `prototypes.py`. @YooSunYoung can we remove one of those?

The argument parsing bits are only for specific workflow daemons/handlers,
which I expect to be handled by plugins themselves later.

So none of the higher-level arguments, such as ``--workflow`` should go here.
``options`` was meant to be the place for functions that handle these highest-level entry-point arguments stuff.

We can in principle remove all the argument builder in ``prototypes`` for now,
but I would rather keep it for easier testing while development...

SimonHeybrock (approved):
Works for me locally :+1:

jokasimr:
> > > I see, there is duplication of the arg parser bits in `options.py` and `prototypes.py`. @YooSunYoung can we remove one of those?
> > 
> > 
> > The argument parsing bits in `prototypes` are only for specific workflow daemons/handlers, which I expect to be handled by plugins themselves later.
> > So none of the higher-level arguments, such as `--workflow` should go here. `options` was meant to be the place for functions that handle these highest-level entry-point arguments stuff.
> > We can in principle remove all the argument builder in `prototypes` for now, but I would rather keep it for easier testing while development...
> 
> The problem is that it is duplicate code, and Johannes changes did not work initially, since there were two copies of the same function.

Nah I'd say the changes didn't work because I was a bit sloppy and forgot to make sure the command line argument was handled the right way. I blame that this is my first contribution to Beamlime so I'm not yet completely clear on how the code is supposed to work.

YooSunYoung:
Ah...! I thought they were deleted in the ``options.py``...! I'll clean them up sorry...!

Issue #153, opened 2024-03-05
Add generator of fake events for `FakeListener`


SimonHeybrock:
Part of #139, follow-up to #152.

- Generate fake events in the `FakeListener`.
  - This may include events for multiple banks, if present in the run-start message (used for initializing the `FakeListener`).
  - `pixel_id` must be setup correctly for the pixel IDs of each bank.
  - Generate random noise, no special shape or distribution for now.
- Produce messages to Beamlimes message router with updates, e.g., once every N seconds.
  - Use #134 to merge the static data with the fake event data.
- We assume the original messages from Kafka have been concatenated within the listener over a period of time. The updates should thus contain a nested json (structure given my the run-start message) containing updates from multiple ev44 (later this will also include f144 / tdct, do not add this for now) Kafka messages.
- The event data should be based on [ev44](https://github.com/ess-dmsc/streaming-data-types/blob/master/schemas/ev44_events.fbs), excluding the `source_name` (which gives the location in the json) and the `message_id` (which is something the listener would handle internally, but not forward to the rest of Beamlime).

YooSunYoung:
Fixed in #165

PR #156, opened 2024-03-05
test: add fake listener

### LINKED ISSUE #152: Add `FakeListener`
Add a fake that can take the place of a real listener and produces messages as they would have been obtained from Kafka. This is part of #139.

- Add `class FakeListener` (similar to the current `StreamSimulator`?)
- Can be initialized from a JSON run-start message, as found in https://gitlab.esss.lu.se/ecdc/nexus-json-templates.
- Will send a single message to the message router, containing the run-start message *with empty NXevent_data and NXlog* groups.
- Messages with event data and log data will be added in follow-up tasks.

jokasimr:
Fixes #152


tests/applications/fake.py
```diff
@@ -0,0 +1,27 @@
+from beamlime import DaemonInterface
+
+
+class FakeListener(DaemonInterface):
+    messenger: MessageRouter
+
+    def __init__(self, run_start_message: dict):
+        self.run_start_message = run_start_message
```
SimonHeybrock, line 8: The fake will need some way of loading a run-start message from a file.

jokasimr:
I'm a bit unsure on where this belongs in the project. Does it belong as a helper the test-suit or should it be part of the beamlime package?

jokasimr:
Not sure how to test it properly. There are some fields I don't know how to initialize, the `messenger: MessageRouter` and the `logger`. Can we have a call @YooSunYoung where you explain some of how that works?

SimonHeybrock:
This should replace the current `StreamSimulator`. I think that can be removed (or in the next step)?

src/beamlime/applications/daemons.py
```diff
@@ -61,3 +74,36 @@ async def run(self) -> None:
         )
 
         self.info("Data streaming finished...")
+
+
+class FakeListener(DaemonInterface):
```
SimonHeybrock, line 79: I think we should replace https://github.com/scipp/beamlime/blob/8958ea820d994e60941c56cd178797443f3db507/src/beamlime/executables/prototypes.py#L29 so it uses this instead. The `DataStreamSimulator` should be removed. What do you think @YooSunYoung?

src/beamlime/applications/daemons.py
```diff
@@ -61,3 +74,36 @@ async def run(self) -> None:
...
+        '''Read nexus structure from json file'''
+        with open(path) as f:
+            nexus_structure = json.load(f)
+        return cls(nexus_structure)
+
+    async def run(self) -> None:
+        self.info("Fake data streaming started...")
+
+        await self.messenger.send_message_async(
+            RunStart(
```
YooSunYoung, line 96: `RunStart` is a bit confusing.
I thought `StatelessWorkflow` won't know about run-start and `KafkaListener` will handle them itself.
For now, it can only send an empty structure and it will be able to fill some data into the structure  as #153 implemented.

src/beamlime/applications/daemons.py
jokasimr: Me and @YooSunYoung talked about this yesterday and then we came to the conclusion that it's better to remove it in the PR where fake events are implemented

src/beamlime/applications/daemons.py
SimonHeybrock: I would just ignore this for now, and fix in the follow-up when we add events. Otherwise this is ill-defined and a waste of time to "fix" here.

YooSunYoung (approved):

PR #163, opened 2024-03-12
Final proposal of ``MessageProtocol``

YooSunYoung:
I propose that this should be the final shape of the `MessageProtocol`.

It now carries ``args: tuple``  and ``kwargs: dict``, which are positional arguments and key-word arguments respectively.

https://github.com/scipp/beamlime/blob/45832a9b9a9e9094814af3bb59fc11cc4750f889/src/beamlime/applications/base.py#L30-L35

So that the ``MessageRouter`` can directly call the handlers by unpacking those arguments like this.
https://github.com/scipp/beamlime/blob/45832a9b9a9e9094814af3bb59fc11cc4750f889/src/beamlime/applications/base.py#L144

Here are some pros&cons. And I tried to argue with the ``cons`` myself : D

## Pros
- ``handling method`` don't have to know about the ``MessageProtocol``.
  ``handling method``s don't have to receive ``Message`` object itself, so you can write a regular method and use it as a handler.
  It means, you don't necessarily have to ``wrap`` an existing function/method to make handler.
  You might still have to wrap it if it needs to return a ``Message`` as a handler, but not as a regular method.
- More loose condition of registering ``handling method`` (related to the previous one)
- More explicit type annotations when you populate the message, if you want.

## Cons
- Type annotation might be more verbose or confusing and duplicating.
   Now that there are two fields in the message, which can replace each other (if you are expecting the handlers to accept arguments in both positional, key-word based)
  And if an expected handler requires many arguments, and you want to annotate all of them, it might be very verbose.
  In that case, we can maybe use a typed dict.
- Not strict shape of handling method.
  Since now the arguments are more flexible, it might require more work to make the messages compatible with the handlers and vice versa.
  But it's always the case, the proposed protocol just makes it fails earlier in the router, not in the handler.
- Can't accept subset of arguments and ignore the rest after receiving the message.
   For example, if you want to trigger the handler but you don't need any of the arguments, (Maybe they are meant for other handlers) the handlers still need to have at least ``*args``, ``**kwargs``.
   But it also means that it can simply pick up subset of the arguments by the method definition itself, without filtering them in the handler.

### DIFF
--- a/src/beamlime/applications/base.py
+++ b/src/beamlime/applications/base.py
@@ -4,6 +4,7 @@ import asyncio
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from dataclasses import dataclass
+from types import MappingProxyType
 from typing import (
     Any,
     AsyncGenerator,
@@ -12,18 +13,34 @@ from typing import (
     Coroutine,
     Generator,
     List,
+    Mapping,
     Optional,
     Protocol,
+    Tuple,
     runtime_checkable,
 )
 
 from ..logging import BeamlimeLogger
 from ..logging.mixins import LogMixin
 
+EMPTY_ARGS = tuple()
+EMPTY_KWARGS = MappingProxyType(dict())
+
 
 @runtime_checkable
 class MessageProtocol(Protocol):
-    content: Any
+    """Container of arguments and keyword arguments for handlers."""
+
+    args: Tuple
+    kwargs: Mapping
+
+
+@dataclass
+class MessageBase:
+    """Base class for easier implementation of ``MessageProtocol``."""
+
+    args: Tuple = EMPTY_ARGS
+    kwargs: Mapping = EMPTY_KWARGS
 
 
 class DaemonInterface(LogMixin, ABC):
@@ -84,11 +101,9 @@ class MessageRouter(DaemonInterface):
     def _register(
         self,
         *,
-        handler_list: dict[
-            type[MessageProtocol], List[Callable[[MessageProtocol], Any]]
-        ],
+        handler_list: dict[type[MessageProtocol], List[Callable[..., Any]]],
         event_tp: type[MessageProtocol],
-        handler: Callable[[MessageProtocol], Any],
+        handler: Callable[..., Any],
     ):
         if event_tp in handler_list:
             handler_list[event_tp].append(handler)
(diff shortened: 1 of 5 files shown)


src/beamlime/applications/base.py
```diff
@@ -12,18 +13,34 @@
...
+    args: Tuple
+    kwargs: Mapping
+
+
+@dataclass
+class MessageBase:
+    """Base class for easier implementation of ``MessageProtocol``."""
+
+    args: Tuple = EMPTY_ARGS
+    kwargs: Mapping = EMPTY_KWARGS
```
jl-wynen, line 43: Can't you just use `dataclasses.field(default_factory=tuple)` instead of a global default variable?

src/beamlime/applications/base.py
jl-wynen, line 1: I'm not really sure what the purpose of this is. In my understanding, handlers are specific to a message type. So their signature must match a given message and vice versa. The flexibility we get from varargs and kwargs doesn't really apply here. Just take a look at the concrete message types. They all have a specific type annotation for `args`.

So what is the benefit of this new approach over dedicated fields in the messages? That is
```python 
@dataclass
class RunStart(MessageBase):
    chunk: JSONGroup

def handler(message: RunStart):
    # use message.chunk here
```

YooSunYoung: It was mainly for avoiding wrapping the function as a handler.
Because you might want to use the function without using messages as well.
There'll be one less step to test those functions as well.

jl-wynen: Then how about this?
```python
@dataclass
class RunStart(MessageBase):
    chunk: JSONGroup

def handler(chunk: JSONGroup):
    ...

# call handler
handler(**dataclasses.asdict(message))
```
This requires keyword args so it limits the handler interface a little. But is that an issue?

YooSunYoung: > This requires keyword args so it limits the handler interface a little. But is that an issue?

That was why I didn't go for it but it would be very rare that we use a function that only allows positional arguments
so we can make a wrapper only for those cases...?

And do you think it makes sense that we just remove MessageBase and accept any ``dataclass``...? By checking `is_dataclass`...? Then we don't even need any protocol or a base class.

YooSunYoung:
Closing it for now 😔...

PR #164, opened 2024-03-12
Support dynamic registration of handlers.

YooSunYoung:
This feature was needed for allowing daemons to separate some tasks from the ``run`` routine.

For example, ``KafkaListener`` might need to handle arbitrary types of messages according to the ``run_start_message``.
With this change, ``KafkaListener`` can now register handlers whenever it receives ``run_start_message`` and don't have to handle them in the ``run`` routine. 
It can simply wrap received messages as a certain types of ``MessageProtocol`` and yield them, 
instead of wrap them **and** handle them within the routine.

### DIFF
--- a/src/beamlime/applications/base.py
+++ b/src/beamlime/applications/base.py
@@ -202,6 +202,12 @@ class Application(LogMixin):
         args: Tuple[type]
         """Sender of the message."""
 
+    @dataclass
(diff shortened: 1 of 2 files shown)


YooSunYoung:
Closing it for now 😔 since #163 was not proceeded yet ...

Issue #134, opened 2024-02-19
Update static data with new time-dependent information from (merged) message


SimonHeybrock:
Initially we get non-time-dependent information from the stream (the run-start message). From then on, messages contain "updates" of time-dependent information, in particular `NXevent_data` and `NXlog`. The "static" run-start message needs to be merged with the updates for the time-dependent info (provided by #133).

Therefore:
- Static info from run-start message needs to be kept around (in the `KafkaListener` or `FakeListener`).
- For every (merged) message containing updated event and log data, insert update in the correct places.
- We need to avoid full deep copies, but keep in mind that we pass the result to downstream handlers, i.e., we must avoid modifying those and thus use copies of dicts on the levels where updates are inserted.
- Send new message to Beamlimes message router.

PR #165, opened 2024-03-13
Random event generator

### LINKED ISSUE #153: Add generator of fake events for `FakeListener`

YooSunYoung:
Fixes #153 

## ``ESSKafkaStreamSimulator`` (previously: ``FakeListener``)
``FakeListener`` was implemented before ``DaemonInterface.run`` became a generator.
I renamed it because it is simulating both ``Producer`` and ``Listener``, not only ``Listener``.

## ``NexusContainer``, Nexus template parsing helpers
It is a wrapper for a ``nexus_structure`` dictionary.
It is currently using dictionary because ``module`` placeholders need to be populated first in order to turn them into ``Group``.

For ``ESSKafkaStreamSimulator``, it was needed to parse the ``detector`` information to generate events based on it.

## Related TODOs
- Smaller ``json`` template with ``NXdetector``, or ask ECDC to have ``NXdetector`` in `Ymir`...?
- There should be another handler object that assembles the received messages using ``NexusContainer``. #134


src/beamlime/applications/_nexus_helpers.py
```diff
@@ -0,0 +1,303 @@
...
+
+FBIdentifier = Literal["ev44", "f144", "tdct"]
+ValuesRecipe = MappingProxyType[str, type]
+
+_EV44_RECIPE = MappingProxyType(
+    {
+        "reference_time": list[float],
+        "reference_time_index": list[int],
+        "time_of_flight": list[float],
+        "pixel_id": list[int],
```
SimonHeybrock, line 19: In practice, I hope these are Numpy arrays, not lists? Use `typing.Sequence`?

src/beamlime/applications/_random_data_providers.py
```diff
@@ -24,108 +18,37 @@
 RandomEvents = NewType("RandomEvents", List[sc.DataArray])
 
 
-def provide_rng(random_seed: RandomSeed) -> RNG:
-    from numpy.random import default_rng
-
-    return default_rng(random_seed)
-
+DetectorNumberCands = NewType("DetectorNumberCands", List[int])
```
SimonHeybrock, line 21: What is `Cands`?

src/beamlime/applications/_random_data_providers.py
```diff
@@ -24,108 +18,37 @@
...
-                unit=sc.units.dimensionless,
-            )
+    et_zero = ReferenceTimeZero(13620492**11)  # No reason
+    while True:
+        cur_event_number = int(
+            ef_rate * (rng.integers(99, 101) / 100)
+        )  # 1% of fluctuation
+        yield EV44(
+            reference_time=[et_zero],
+            reference_time_index=[0],
```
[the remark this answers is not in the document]
SimonHeybrock: Good question. Will the stream use the time index to refer to the history (then this should be something like the cumsum of the number of events in all previous messages since the run-start), or only the current message?

src/beamlime/applications/daemons.py
```diff
@@ -1,72 +1,63 @@
...
 
+NexusTemplatePath = NewType("NexusTemplatePath", str)
 
-class DataStreamSimulator(DaemonInterface):
-    """Data that simulates the data streaming from the random generator."""
 
-    random_events: RandomEvents
-    chunk_size: ChunkSize
-    data_feeding_speed: DataFeedingSpeed
+class ESSKafkaStreamSimulator(DaemonInterface):
```
SimonHeybrock, line 29: This name is confusing. The fake that we are implementing is meant to stand in for the Kafka stream listener. It does not do anything with the Kafka side of things (and as explained on Tuesday, we may later have a `FakeKafka`, so we can test the "real" listener. What was wrong with `FakeListener`? Sure it has to "produce" something that it forwards, but it is "faking" the real listener (which will be implemented later).

tests/applications/test_kafka_simulator.py
SimonHeybrock, line 1: See above, rename to not include Kafka.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        return nested_dict_select(
+            nexus_container,
+            'children',
+            partial(match_nx_class, cls_name='NXentry'),
+            'children',
+            partial(match_nx_class, cls_name='NXinstrument'),
+        )
+
+    def prune_pixel_info(self) -> dict:
```
SimonHeybrock, line 284: Where is this used?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -0,0 +1,303 @@
...
+        pixel_offsets = [
+            child
+            for child in self.detector_dict['children']
+            if _match_dataset_name_suffix(child, 'pixel_offset')
+        ]
+        self.pixel_offsets = {
+            offset['config']['name']: offset['config']['values']
+            for offset in pixel_offsets
+        }
+        self._pixel_offsets_idx = {
```
SimonHeybrock, line 231: What is this private field used for?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -0,0 +1,303 @@
...
+
+        The values are extended if they are list and replaced otherwise.
+        """
+        if 'values' not in self.config_container or not isinstance(values, list):
+            self.config_container['values'] = values
+        elif isinstance(self.config_container['values'], list):
+            self.config_container['values'].extend(values)
+
+
+class ModularDataGroupContainer:
```
SimonHeybrock, line 157: Where is this used?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -0,0 +1,303 @@
...
+            if _match_module_name(cand, fb_id):
+                collected.append(cand_idx)
+            else:
+                _recursive_collect_indices(cand, cand_idx)
+
+    _recursive_collect_indices(nexus_container, tuple())
+    return collected
+
+
+class ModularDatasetContainer(dict):
```
SimonHeybrock, line 130: Where is this used?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -0,0 +1,303 @@
...
+    return child.get('module', None) == name
+
+
+def _match_dataset_name_suffix(child: dict, suffix: str) -> bool:
+    return child.get('module', None) == 'dataset' and child.get('config', {}).get(
+        'name', ''
+    ).endswith(suffix)
+
+
+def collect_nested_module_indices(
```
SimonHeybrock, line 108: Where is this used?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -0,0 +1,303 @@
...
+_TDCT_RECIPE = MappingProxyType({"name": str, "timestamps": list[int]})
+
+
+MODULE_RECIPES: MappingProxyType[FBIdentifier, ValuesRecipe] = MappingProxyType(
+    {
+        "ev44": _EV44_RECIPE,
+        "f144": _F144_RECIPE,
+        "tdct": _TDCT_RECIPE,
+    }
+)
```
SimonHeybrock, line 36: Used only be unused(?) class below?

src/beamlime/applications/_random_data_providers.py
YooSunYoung: ``Candidates`` sorry...
Since the geometry will carry all possible ``pixel id``s, not a range,
it can just receive that list and choose among them.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: It is a helper to prune pixel ids and pixel x/y/z offsets so that I can dump the json structure to another file and investigate it. The original files are too large so that the highlighter doesn't work, for example.

I guess we can just remove it though...

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Just for pruning them later...?

src/beamlime/applications/_random_data_providers.py
SimonHeybrock: Can you rename it?

src/beamlime/applications/_nexus_helpers.py
SimonHeybrock: Yes, please remove everything unrelated to the task.

src/beamlime/applications/_nexus_helpers.py
SimonHeybrock: Later where?

src/beamlime/applications/daemons.py
YooSunYoung: I don't know... I agree that it should not include `Kafka` but `Listener` is already 
but `FakeListener` sounds like a stub or a mock rather than something actually has a functionality...
But name doesn't really matter... I can revert the name

SimonHeybrock: A "fake" is generally more than a simple stub or mock. Here is a ChatGPT summary:

```
In software development and testing, stubs, mocks, and fakes are all used to simulate parts of a system that are not yet available or to isolate components for testing purposes. While they serve similar purposes, they have different characteristics and usage contexts:

1. **Stub**:
   - A stub is a minimal implementation of a component or module that provides canned responses to function calls or method invocations.
   - Stubs are typically used in integration testing when a component depends on another component that is not yet implemented or available.
   - Stubs do not have full functionality and are often hardcoded to return specific values or execute predefined behavior.
   - They are simple placeholders that allow developers to continue working on parts of the system independently.

2. **Mock**:
   - A mock is a test double that mimics the behavior of a real object or component in a controlled way.
   - Mocks are used primarily in unit testing to verify interactions between objects and to test the behavior of the object being tested.
   - Unlike stubs, mocks are programmed with expectations about how they should be called (i.e., which methods and with what parameters).
   - Mocks can simulate both expected and unexpected behaviors, allowing developers to test edge cases and error handling.

3. **Fake**:
   - A fake is a simplified functional implementation of a dependency that is used in place of the actual implementation.
   - Fakes are often used in testing when the real implementation is too complex, slow, or resource-intensive to use in testing scenarios.
   - Fakes are more sophisticated than stubs because they provide a functional alternative, but they are simpler than the actual implementation.
   - For example, a fake database implementation might store data in memory rather than on disk to speed up testing.

In summary, stubs provide basic, predefined responses to method calls, mocks are used to verify interactions and behavior during unit testing, and fakes provide functional but simplified alternatives to real dependencies for testing purposes. Each serves a different purpose in software development and testing, helping developers write robust and maintainable code.
```

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: In the ``prune_pixel_info``, which is deleted so this part is also deleted.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Sorry these are for the #134 I'll move it

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Removed it since it's only for #134

SimonHeybrock (approved):
👍

SimonHeybrock (approved):

src/beamlime/applications/_random_data_providers.py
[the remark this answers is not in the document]
YooSunYoung: I started a thread of QnA in #168 to collect questions and remind us to document them.

PR #166, opened 2024-03-13
Message router handle stops before it raises.

YooSunYoung:
Once coroutines are created, it is no longer bound to each other.
Therefore, raising an error in one of coroutines won't stop other coroutines to stop.
Therefore we need to make sure ``Application.Stop`` is handled before the message handlers raise any errors.

Or we can also have a safety net in the coroutine to check if any tasks are failed...? I don't know...
Currently handling ``Applicaiton.Stop`` before raising an error was sufficient, and was tested in #165 

The reason is explained in the comment of the code as well.

### DIFF
--- a/src/beamlime/applications/base.py
+++ b/src/beamlime/applications/base.py
@@ -70,13 +70,18 @@ class MessageRouter(DaemonInterface):
     def _handler_wrapper(
         self, handler: Callable[..., Any], message: MessageProtocol
     ) -> Generator[None, None, None]:
-        import warnings
-
         try:
             self.debug(f"Routing event {type(message)} to handler {handler}...")
             yield
         except Exception as e:
-            warnings.warn(f"Failed to handle event {type(message)}", stacklevel=2)
+            # Make sure ``Stop`` event is handled.
+            # It is because raising an exception will
+            # only destroy the ``MessageRouter`` coroutine, not other coroutines.
+            # Therefore other coroutines may not be stopped in
+            # some environments, e.g. Jupyter Notebooks.
+            for stop_handler in self.handlers.get(Application.Stop, []):
+                stop_handler(Application.Stop(content=None))
+            self.warning(f"Failed to handle event {type(message)}")
             raise e
         else:
             self.debug(f"Routing event {type(message)} to handler {handler} done.")
@@ -137,11 +142,7 @@ class MessageRouter(DaemonInterface):
 
         # No handlers
         if not (handlers or awaitable_handlers):
-            import warnings
-
-            warnings.warn(
-                f"No handler for event {type(message)}. Ignoring...", stacklevel=2
-            )
+            self.warning(f"No handler for event {type(message)}. Ignoring...")
 
         # Re-route the results
         for result in results:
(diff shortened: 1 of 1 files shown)


src/beamlime/applications/base.py
jl-wynen, line 1: This seems brittle. Is there a way to force-stop the event loop and all coroutines? Or would this cause problems because it doesn't shut down cleanly?

SimonHeybrock: Note that application control such as pause/resume/cancellation is part of the *next* **but one** milestone, i.e., we just need something that works for now. See https://github.com/scipp/ess-project/blob/main/planning/high-level-plan-2024.md#beamlime.

YooSunYoung: > This seems brittle. Is there a way to force-stop the event loop and all coroutines? Or would this cause problems because it doesn't shut down cleanly?

I think handling them in the ``Application`` is worth adding as well as routing ``Stop`` message. So that daemons or handlers can subscribe to the ``Stop`` message for clean stop as well.
It'll be just one more try-exception so I'll add them ...!

src/beamlime/applications/base.py
```diff
@@ -215,14 +210,43 @@ def register_daemon(self, daemon: DaemonInterface) -> None:
...
+            # It is because raising an exception will destroy only the task
+            # that had an error raised and may not affect other tasks in some cases,
+            # e.g. in Jupyter Notebooks.
+            self.error(f"Daemon {daemon} failed. Cancelling all other tasks...")
+            # Break all daemon generator loops.
+            self._break = True
+            # Let other daemons/handlers clean up.
+            await self.message_router.route(self.Stop(None))
+            # Make sure all other async tasks are cancelled.
+            self.cancel_all_tasks()
```
jl-wynen, line 236: Is there actually time for the handlers to stop? Or does this just cancel all tasks before they can be resumed and handle the stop message?

Or are message handlers and tasks completely separate?

YooSunYoung: > Or are message handlers and tasks completely separate?

Yes, they are separate.
Even if ``message_router.run`` is canceled, since ``route`` is directly called here so it should be able to call all handlers that are subscribing ``Stop``.

src/beamlime/applications/base.py
jl-wynen: Not async handlers. They may never resume to handle the stop message.

src/beamlime/applications/base.py
YooSunYoung: The coroutines (that are used to create the tasks) are dynamically created, using `DaemonInterface` run coroutines when `Application.run` is called.

And the handlers are called by ``MessageRouter.route``, which is constantly called in ``MessageRouter(DaemonInterface).run``.

So, if a ``MessageRouter.run`` task is canceled, even if ``Stop`` message is populated, it will not be routed automatically anymore.

But daemon objects including `MessageRouter` are still alive,  so we can still use ``route`` method  to route``Stop`` to handlers.

Did it answer your question....?

jl-wynen (approved):

PR #167, opened 2024-03-14
Fix benchmark helper for MacOS.

YooSunYoung:
``psutil`` doesn't support ``cpu_affinity`` in ``MacOS``.

It's not important/urgent but it was annoying when I run ``tox`` tests/docs build...


YooSunYoung:
I'll just merge it since it's a minor change in a test helper.

Issue #171, opened 2024-03-18
Handle ``f144(log)`` and ``tdct(chopper)``.


YooSunYoung:
Currently ``DataReductionHandler`` can only insert ``ev44(event data)`` into the nexus structure.

But as mentioned in the #134, we also need to handle ``f144`` and ``tdct`` messages.

jokasimr:
This was implemented in #134 but not tested.
So probably we only need to make sure that the current implementation works, or tweak it if it doesn't.

Issue #168, opened 2024-03-15
Data Streaming QnA.


YooSunYoung:
List of things that should be answered by ECDC and documented in the beamlime.

- [x] Will the stream use the time index to refer to the history (then this should be something like the cumsum of the number of events in all previous messages since the run-start), or only the current message?
- [x] How can you associate these module place holders from the message? Are they defined per message type('ev44' or 'f144' for example).

YooSunYoung:
- [ ] How can you associate these module place holders from the message? Are they defined per message type('ev44' or 'f144' for example). 

Like, how can you fill this:

```json
"module": "f144",
"config": {
    "topic": "loki_motion",
    "source": "source",
    "dtype": "double"
}
```

from this

```json
table f144_LogData {
  source_name: string (required);      // EPICS PV name
  timestamp: long;                     // Nanoseconds since UNIX epoch
  value: Value (required);             // May be scalar or array
}
```

YooSunYoung:
Here is a link to the ECDC document about kafka topics and message schemas: https://confluence.esss.lu.se/display/ECDC/Kafka+Topics+Overview+for+Instruments

PR #172, opened 2024-03-18
Insert event data into nexus structure.

### LINKED ISSUE #134: Update static data with new time-dependent information from (merged) message

YooSunYoung:
Fixes #134

I tried to make it generic so that ``DataReductionHandler`` can insert any time-dependent data into the nexus structure,
but the ``source`` of the ``f144(log``, ``ev44(event data)`` and ``tdct(chopper)`` don't seem to be generic...

We are collecting related questions here: #168 and in the mean-time we decided to just use event data.

I created related issue to fix the rest of them #171.


src/beamlime/applications/handlers.py
```diff
@@ -23,17 +22,22 @@ class DataReductionHandler(HandlerInterface):
...
-    def process_message(self, message: MessageProtocol) -> MessageProtocol:
-        content = message.content
-        self.info("Received, %s", self.format_received(content))
-        return WorkflowResultUpdate(content=self.workflow(content))
+    def process_message(self, message: DetectorDataReceived) -> WorkflowResultUpdate:
+        self.info("Received data from %s.", message.content["source_name"])
+        self.nexus_container.insert_ev44(message.content)
+        return WorkflowResultUpdate(
+            content=self.workflow(self.nexus_container.nexus_dict)
+        )
```
SimonHeybrock, line 40: This makes the stateless workflow handler stateful. This seems to undo what we have worked towards in the past couple of weeks.

- The stateless workflow must be called on data that contains something akin to a time-slice of an NeXus file. We probably do not want to do this on something where only a single ev44 update is included.
- Please read the description in #134 (which this PR is supposed to fix). The message merging should *not* be part of the workflow. It is done in, e.g., the (Fake)Listener, or possibly a step in between.

Aside from the above, can you minimize package-wide changes that are unrelated to the PR? It makes reviewing really hard.

SimonHeybrock (changes requested):

YooSunYoung:
This PR will be split into multiple ones starting with #175 ...

PR #176, opened 2024-03-19
Add hypothetical detectors in the ymir json template.

YooSunYoung:
For testing event data assembling.

### DIFF
--- a/tests/applications/test_listeners.py
+++ b/tests/applications/test_listeners.py
@@ -43,7 +43,8 @@ def fake_listener() -> FakeListener:
 def test_fake_listener_constructor(
     fake_listener: FakeListener,
 ) -> None:
-    assert len(fake_listener.nexus_container.detectors) == 0  # ymir has no detectors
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
Why can't we use Loki template? This is what we want to use for demoing data reduction anyway.

YooSunYoung:
> Why can't we use Loki template? This is what we want to use for demoing data reduction anyway.

Loki template is too large... I guess I can just prune them though. Should I...?

SimonHeybrock:
Too large for what? The size is something we have to be able to deal with, and since we want to start working on Loki workflow with Beamlime very soon it cannot be postponed long.

YooSunYoung:
> Too large for what? The size is something we have to be able to deal with, and since we want to start working on Loki workflow with Beamlime very soon it cannot be postponed long.

For putting somewhere accessible and download it, etc to use it for unit tests.
I already tested it with loki json file as well.

SimonHeybrock (approved):

PR #177, opened 2024-03-19
Insert event data into nexus structure by data assembling handler.

### LINKED ISSUE #134: Update static data with new time-dependent information from (merged) message

YooSunYoung:
Fixes #134 
Continued from #172


src/beamlime/applications/_nexus_helpers.py
```diff
@@ -2,11 +2,56 @@
...
+        }
+    )
+
+
+class DatasetRecipe(NamedTuple):
+    dtype: type
+    ndim: int
+
+
+FBIdentifier = Literal["ev44", "f144", "tdct"]
```
SimonHeybrock, line 39:
```suggestion
FlatBufferIdentifier = Literal["ev44", "f144", "tdct"]
```
?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -131,3 +269,24 @@ def _get_instrument_group(self, nexus_container: dict) -> dict:
             'children',
             partial(match_nx_class, cls_name='NXinstrument'),
         )
+
+    def _collect_ev44(self) -> dict[str, ModularDataGroupContainer]:
+        """Collect the ev44 modules from the detector group."""
```
SimonHeybrock, line 274: ... and monitor groups

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -66,6 +111,95 @@ def _match_dataset_name(child: dict, name: str) -> bool:
...
+            self.config_dict['values'] = np.append(
+                self.config_dict['values'],
+                new_values
+                # if ``new_values`` is a sequence, it will be extended
+                # if it is a single value, it will be appended
+            )
+
+
+class ModularDataGroupContainer:
+    """Module dataset container."""
```
SimonHeybrock, line 150:
```suggestion
class ModularGroupContainer:
    """Modular group container."""
```
?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -2,11 +2,56 @@
...
+
+
+class DatasetRecipe(NamedTuple):
+    dtype: type
+    ndim: int
+
+
+FBIdentifier = Literal["ev44", "f144", "tdct"]
+ValuesRecipe = MappingProxyType[str, DatasetRecipe]
+GROUP_RECIPES: MappingProxyType[FBIdentifier, ValuesRecipe] = _nested_mapping_proxy(
```
SimonHeybrock, line 41: Why do we need the `_nested_mapping_proxy`?

src/beamlime/applications/handlers.py
```diff
@@ -3,37 +3,59 @@
...
+        from ._nexus_helpers import NexusContainer
+
+        self.nexus_container: NexusContainer
+
+    def set_run_start(self, message: RunStart) -> None:
+        self.nexus_container = message.content
+
+    def assemble_detector_data(self, message: DetectorDataReceived) -> DataReady:
+        self.info("Received data from detector %s", message.content['source_name'])
+        self.nexus_container.insert_ev44(message.content)
```
SimonHeybrock, line 43: Won't this do an in-place update of `self.nexus_container.nexus_dict`? Since we are forwarding this we must avoid modification (see item in #134). Or is this taken care of somehow in `update` , in a way that I cannot see?

SimonHeybrock (commented):
The handler setup looks cleaner now 👍 

Some questions:

src/beamlime/applications/handlers.py
YooSunYoung: You're right. But how do we partially update a nested dictionary without making a copy...?
One nexus structure can be up to GB of size in the memory as far as I understood.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: I wanted to keep it read-only... but since the ``DatasetRecipe`` is a data class, it won't be read-only anyways. I'll just remove it.

src/beamlime/applications/handlers.py
SimonHeybrock: We only copy the subdicts that get updated. I presume most of the size is in datasets, or subdicts that do not change.

YooSunYoung: Hmm so you're saying we should copy, for example, ``NXeventdata``...?
But those should be somehow inserted into the whole dataset though.

SimonHeybrock: I think you need to copy (not deep-copy) all dicts in the path to the event data as well.

YooSunYoung: You mean we make a nested shallow copy and keep the original one somewhere else?

SimonHeybrock: Yes, something like that.

YooSunYoung:
I'll reopen it when it's ready

YooSunYoung:
I'll split it into multiple PRs...

YooSunYoung:
Closing this as it was split into multiple PRs in #183 , #181 and #180

PR #178, opened 2024-03-20
Nightly tests with large files (loki template)

YooSunYoung:
nexus templates are typically large-size, (~200MB for loki.json) so it had to be in the public server, instead of `tests` folder.
So I moved ymir.json and loki.json file to server and added pooch helper.

I assumed that the list of template files will become longer so that we can test more templates in the future.

However, I modified ``ymir.json`` file so I kept it in the repository and added a test for checking if we wanted to update it...
I couldn't come up with a better solution...


YooSunYoung:
I tested it manually as well: https://github.com/scipp/beamlime/actions/runs/8360200104/job/22885339987

tests/applications/test_nexus_helpers.py
```diff
@@ -31,6 +41,26 @@ def ev44_generator() -> RandomEV44Generator:
     )
 
 
+def test_ymir_template_checksum() -> None:
+    """Test that the ymir template is updated.
+
+    ``ymir.json`` is modified to include detector data.
```
SimonHeybrock, line 47: I suggest renaming the modified version, so we can (if we ever want to) also use the original.

YooSunYoung: @SimonHeybrock  Done...!

SimonHeybrock (approved):

Issue #182, opened 2024-04-05
Async lock or threading lock needed for appending dataset.


YooSunYoung:
For now, ``DataAssembler``, through ``NexusContainer``, create a shallow copy of the original nexus dictionary and update it. 
Therefore, it is not guaranteed that ``NexusContainer.nexus_dict`` is always the latest one.

For example, if there is an unnecessary ``asyncio.sleep`` in the middle of handling method like below,
```python
nexus_dict = {'event' = []}
nexus_container = {'nexus_dict': nexus_dict}

async def handling_method() -> None:
  copied = nested_shallow_copy(nexus_container['nexus_dict'])
  copied['event'].append(3)
  await asyncio.sleep(0)
  nexus_container['nexus_dict'] = copied

await handling_method()
await handling_method()
   
```
The expected result of ``nexus_container`` should be
```python
{
  'nexus_dict': {'event': [3, 3]}
}
```
But it might be 
```python
{
  'nexus_dict': {'event': [3]}
}
```
Because of the ``asyncio.sleep`` in the middle, because the latest result was not updated in the ``nexus_container``.

So this should be handled when we want to append data in the ``nexus_container``.
It may be handled by async ``Lock``.

jl-wynen:
In this example, it is a bit strange that you copy the dict and then insert the copy back into the outer dict. Instead, it seems more reasonable to have an atomic operation on the nested dict and modify it in place along the lines of

```python
async def handling_method() -> None:
  new_value = some_function_that_maybe_uses_nexus_container()
  await asyncio.sleep(0)
  atomic_append_event(nexus_container, new_value)

def atomic_append_event(container, value):
  # no lock needed in async context,
  # otherwise use thread mutex here
  container['nexus_dict']['event'].append(value
```

Does this also apply to the real code or is the example simplified too much?

YooSunYoung:
>  Instead, it seems more reasonable to have an atomic operation on the nested dict and modify it in place along the lines of

You mean, we should make sure that the `old` one is used 
and then update it with new values in-place...?

jl-wynen:
I don't understand what you mean. Let's have a quick call tomorrow!

YooSunYoung:
This is not relevant anymore since we make sure we collect enough of the dataset first before we inject them into the whole structure.

Issue #133, opened 2024-02-19
Merge messages


SimonHeybrock:
We get relatively small messages from Kafka, e.g., one per pulse. This is too inefficient to feed through the data-reduction. We therefore need to merge messages before passing the data to the reduction workflow.

Notes:
- Merging is generic, does not need deep NeXus or workflow knowledge.
- We need to concat fields of `NXevent_data` and `NXlog`
- Merging should be done on the raw JSON, operating on NumPy arrays. That is, *before* we parse using ScippNexus.

PR #183, opened 2024-04-05
Insert event data into nexus structure by data assembling handler.

### LINKED ISSUE #134: Update static data with new time-dependent information from (merged) message

YooSunYoung:
Continued from #177 
Fixes #134

## Main Changes
Handler, ``DataAssembler`` receives raw data messages (for now only ``ev44`` and it doesn't matter if it's fake or not) and use ``NexusContainer`` to assemble them into the whole nexus structure and send the assembled nexus structure to the message router again.

``_nexus_helpers.replace_subdataset`` create a shallow copy from root, to the point of insertion so that the previously assembled dictionary is not affected by the update.

In order to update the nexus structure (in this case, dictionary) without affecting the previous data, there are helper functions for handling/filtering nested dictionary/list and validating the whole structure.

## Following-up Issues
For now, ``NexusContainer.insert_ev44`` method only replaces the data, but since it is now shallow copying the original structure, it is not guaranteed that the ``DataAssembler`` is looking at the ``latest`` data if they are being ``appended``, not ``replaced``. Related issue: #182.

And I expected ``DataAssembler`` will later take care of merging multiple messages by simply concatenating them or using the average. Related issue: #133


src/beamlime/applications/_nexus_helpers.py
```diff
@@ -2,10 +2,72 @@
...
+    >>> a['c']['c-a']  # The original dictionary is also changed
+    4
+    """
+    from copy import copy
+
+    root_obj = obj.copy()
+    cur_parent = root_obj
+
+    try:
+        for idx in indices:
```
SimonHeybrock, line 64: So this is more like a path, not independent indices?

src/beamlime/applications/handlers.py
```diff
@@ -3,37 +3,59 @@
...
+
+    def set_run_start(self, message: RunStart) -> None:
+        self.nexus_container = message.content
+
+    def assemble_detector_data(self, message: DetectorDataReceived) -> DataReady:
+        self.info("Received data from detector %s", message.content['source_name'])
+        self.nexus_container.insert_ev44(message.content)
+        return DataReady(
+            content=self.nexus_container.nexus_dict  # should be a JSONGroup later.
+        )
```
SimonHeybrock, line 46: This looks like it is inserting only a single detector, so the data passed to the data reduction always only contains a single (but different every time) detector? I think this will not work. The listener must merge this data (as mentioned in the issue #134), such that the assembled result is roughly equivalent to reading a time-slice from the file.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Yes...! I'll rename it for clarity...

src/beamlime/applications/handlers.py
YooSunYoung: No, the ``insert_ev44`` method also replaces the ``nexus_dict`` with the updated one, so it still has other detectors.

SimonHeybrock: So this just accumulates indefinitely? I thought we wanted to accumulate for some time and *not* push the entire data to the data reduction with every update.

YooSunYoung: No, it replaces the dataset for each detector.
We only accumulate the dataset of the number of detectors(ev44 place holders) at maximum.

Merging messages is not implemented yet.

YooSunYoung: Should we have a call instead...?

SimonHeybrock: Ok, so you are saying that:
- Every time *one* detector gets an update...
- ... we replace that but keep the others
- Which means we will always have the "most recent" state (except at the beginning).
- We forward to data reduction *with every update*

Correct?

I think this will still not work, since at the beginning not all detectors will have received updates yet. Also, we call data reduction to often.

I am not saying that message merging needs implementing here, but shouldn't we gather updates from all detectors, such that what we pass to data reduction is consistent?

YooSunYoung: > Correct?

Yes, the idea was to keep the ``most recent`` state.

> I am not saying that message merging needs implementing here, but shouldn't we gather updates from all detectors, such that what we pass to data reduction is consistent?

I thought that should be done as a part of ``merge``.
It's not so simple when you don't know when exactly all the detectors are sending their messages.
So we have to either
1. Flush the dataset(Trigger the data reduction) only when whole set of dataset from all detectors arrive
2. Merge `n*detectors` number of messages, hoping that they are usually received in order.

And I thought the option `1.` is not really feasible or robust because then we know which detector are sending data or not from the beginning. So it'll be broken as soon as part of detectors are not functioning or turned off for some reason. And even if we go for this option, aligning multiple messages based on the `source` should be done as a separate task not here.


Is this current implementation too far to be considered 'working'...?

SimonHeybrock: I think it will be

3. Merge M seconds, use empty data for datasets without updates.

But yes, we can do it in the follow up. I will write this down more clearly in an issue.

SimonHeybrock (approved):

src/beamlime/applications/handlers.py
SimonHeybrock: I updated https://github.com/scipp/beamlime/issues/133. Please read the description of that task an let me know what is unclear!

PR #184, opened 2024-04-17
Merge messages

### LINKED ISSUE #133: Merge messages

jokasimr:
Fixes #133 

There are some large refactorings of the `nexus_helpers` in the PR.
They're not necessary to fix the issue, but I think they make it easier, and implementing the merging here without refactoring and then refactoring anyway would have been unnecessary, so I thought it better to refactor directly.

The improvements from the refactoring are mainly
  * reduced code size
  * less state
  * fewer datatypes

### DIFF
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -14,6 +14,9 @@ from typing import (
     Union,
 )
 
+import numpy as np
+from deepcopy import deepcopy
+
 Path = Union[str, bytes, os.PathLike]
 DTypeDesc = Literal["int", "float", "string"]
 
@@ -249,6 +252,99 @@ class NXDetectorContainer:
             self.detector_dict, 'children', pixel_id_filter, 'config', 'values'
         )
         self.pixel_ids = nested_dict_getitem(self.detector_dict, *self._pixel_ids_idx)
+
+
+def _merge_ev44(s, m):
+    accumulation_fields = (
+        'time_of_flight',
+        'pixel_id',
+    )
+    for field in set(m.keys()) - set(accumulation_fields):
+        s[field] = m[field]
+    for field in accumulation_fields:
+        s[field] = np.concatenate((s.get(field, []), m[field]))
+
+
+def _merge_f144(s, m):
+    accumulation_fields = ('timestamp', 'value')
+    for field in set(m.keys()) - set(accumulation_fields):
+        s[field] = m[field]
+    for field in accumulation_fields:
+        s[field] = np.stack((s.get(field, []), [m[field]]))
+
+
+def _merge_tdct(s, m):
+    for field in set(m.keys()) - {'timestamps'}:
+        s[field] = m[field]
+    s['timestamps'] = np.concatenate((s.get('timestamps', []), m['timestamps']))
+
+
+def merge(acc, message):
+    kind, content = message
+    if kind == 'ev44':
+        _merge_ev44(acc, content)
+    elif kind == 'f144':
+        _merge_f144(acc, content)
+    elif kind == 'tdct':
+        _merge_tdct(acc, content)
+    else:
+        raise NotImplementedError
+
+
+def node_name(n):
+    config = n.get('config', {})
+    return n.get('name', config.get('name', config.get('source', '')))
+
+
+def iter_nexus_structure(structure, root=()):
+    if not isinstance(structure, dict):
+        return
+    path = (*root, node_name(structure))
+    yield path, structure
+    for child in structure.get('children', []):
+        yield from iter_nexus_structure(child, root=path)
+
+
+def find_nexus_structure(structure, path):
+    if path == ():
+        return structure
+    head, *tail = path
+    for child in structure['children']:
+        if head == node_name(child):
+            return find_nexus_structure(child, tail)
+
+
+def path_to_message_container(structure, message):
+    kind, content = message
+    for path, node in iter_nexus_structure(structure):
+        if (
+            node.get('module') == kind
+            and node.get('config', {}).get('source') == content['source_name']
+        ):
+            return path
+
+
+def merge_message_into_store(store, structure, message):
+    path = path_to_message_container(structure, message)
+    merge(store.setdefault(path, {}), message)
+
+
+def combine_store_and_structure(store, structure):
+    structure = deepcopy(structure)
+    for path, data in store.items():
+        c = find_nexus_structure(structure, path)
+        c.pop('module')
+        c['children'] = [
+            dict(
+                module='dataset',
+                config=dict(
+                    name=name,
+                    values=values,
+                ),
+            )
+            for name, values in data.items()
+        ]
+    return structure
 
 
 class NexusContainer:
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
 import os
+from copy import deepcopy
 from functools import partial
 from typing import (
     Any,
@@ -15,7 +16,6 @@ from typing import (
 )
 
 import numpy as np
-from deepcopy import deepcopy
 
 Path = Union[str, bytes, os.PathLike]
 DTypeDesc = Literal["int", "float", "string"]
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -1,256 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
-import os
-from functools import partial
-from typing import (
-    Any,
-    Callable,
-    Literal,
-    Mapping,
-    NamedTuple,
-    Optional,
-    Tuple,
-    TypeVar,
-    Union,
-)
+from typing import Iterable, Mapping, Optional, Tuple
 
 import numpy as np
-
-Path = Union[str, bytes, os.PathLike]
-DTypeDesc = Literal["int", "float", "string"]
-
-
-class DatasetRecipe(NamedTuple):
-    name: str
-    dtype_description: DTypeDesc
-
-
-FlatBufferIdentifier = Literal["ev44", "f144", "tdct"]
-GROUP_RECIPES: Mapping[FlatBufferIdentifier, Tuple[DatasetRecipe, ...]] = {
-    "f144": (DatasetRecipe("timestamp", "int"), DatasetRecipe("value", "float")),
-    "tdct": (DatasetRecipe("name", "string"), DatasetRecipe("timestamps", "int")),
-    "ev44": (
-        DatasetRecipe("reference_time", "int"),  # In nanoseconds
-        DatasetRecipe("reference_time_index", "int"),
-        DatasetRecipe("time_of_flight", "int"),  # In nanoseconds
-        DatasetRecipe("pixel_id", "int"),
-    ),
-}
-
-
-NestedObjectT = TypeVar("NestedObjectT", dict, list[dict])
-
-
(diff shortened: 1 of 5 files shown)


src/beamlime/applications/daemons.py
```diff
@@ -49,45 +96,25 @@ def __init__(
...
-        }
-        self.data_feeding_speed = speed
-        self.num_frames = num_frames
 
-    def _collect_detector_ev44_modules(self) -> list[str]:
-        """Assumig 'detector' is included in the topic of the module."""
+        with open(nexus_template_path) as f:
+            self.nexus_structure = json.load(f)
+
+        if self.nexus_structure['children'][0]['children'][-1] == '$USERS$':
```
jokasimr, line 103: Aha, this should be removed.
The nexus structure templates in gitlab have some templating variables that I assume will not be present in the run-start messages. The presence of the templating variables messes up the parsing of the nexus structure, thats why this was needed here to make the tests pass.

I think the better option would be to remove the templating variable directly in the json file and store that version in putch.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,327 +1,147 @@
...
-        raise KeyError(f"Invalid path: {path_to_target}") from e
-
-    return root_obj
-
-
-def find_index(
-    obj: dict | list[dict],
-    *filter_or_indices: Callable | str | int,
-    _cur_idx: Optional[tuple[str | int, ...]] = None,
-) -> tuple[str | int, ...]:
```
YooSunYoung, line 77: These helpers are also for finding attributes of the nexus not just children. But I guess we can put them back when they are actually needed...

jokasimr: Yes we can add it back in that case. My feeling right now is that it is a too general helper function. It simplifies things to assume the kind of structure we expect (that is, nexus-json), instead of arbitrary nested lists/dictionaries.

SimonHeybrock (commented):
Looking mostly at a high level, I think this is what we want, and the cleanup makes sense.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,327 +1,147 @@
...
+
+
+def combine_store_and_structure(store, structure):
+    '''Creates a new nexus structure, replacing the stream modules
+    with the datasets in `store`, while avoiding
+    to copy data from `structure` if unessecary'''
+    if len(store) == 0:
+        return structure
+    if (None,) in store:
+        return store[(None,)]
```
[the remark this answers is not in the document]
jokasimr: The stream modules in the nexus json template are "single children" to a group.
Filling the structure with data from the stream replaces the group with a new group that has datasets as children, where the datasets contain the data from the stream.

The stream module (the single child) has no name, that is, the path from the parent group to the stream module is `(None,)`.
If we encounter such a path in the `store` (a mapping from paths to groups created from data streams), replace the group with the group associated with the stream module.

YooSunYoung:
I made them stateful to prevent any overhead when processing the real-time messages
by arranging necessary information at the start and not again during the processing.

And the data structures were for making it clear what's stored where and how,
and also easily add handles for certain part/attributes of the data that is often needed.

I usually prefer stateless and functional way but I'm not sure if that's the best solution for beamlime as well.
But maybe these are not so much of overhead compared to the actual workflow...

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Yeah that's true. I made them generic assuming the structure is an arbitrary dictionary. 
Cause I thought... if we are so sure of the structure, we can just use scippnexus directly and no need to have these from the first place.

SimonHeybrock:
> I made them stateful to prevent any overhead when processing the real-time messages by arranging necessary information at the start and not again during the processing.

Are you referring to walking the tree structure every time?

> And the data structures were for making it clear what's stored where and how, and also easily add handles for certain part/attributes of the data that is often needed.

Can you give an example?

tests/applications/test_nexus_helpers.py
```diff
@@ -18,22 +20,25 @@
...
+from beamlime.applications.daemons import fake_event_generators
+from tests.applications.data import get_path
 
 
 @pytest.fixture
-def ymir_detectors_container() -> NexusContainer:
-    from tests.applications.data import get_path
-
-    return NexusContainer.from_template_file(get_path('ymir_detectors.json'))
+def ymir():
```
YooSunYoung, line 28: I made the name specific because this one is modified from the original `ymir` template.
Can we keep the `detectors` in the name?

YooSunYoung:
> > I made them stateful to prevent any overhead when processing the real-time messages by arranging necessary information at the start and not again during the processing.
> 
> Are you referring to walking the tree structure every time?

Yeah ah but I found that the path is retrieved only at the beginning as well.

> > And the data structures were for making it clear what's stored where and how, and also easily add handles for certain part/attributes of the data that is often needed.
> 
> Can you give an example?

Like, if we just use dictionary, we don't know each item means.

So I wanted to wrap them, for example, the path to the stream modules and the type of the module.
I thought it might be easier to store them in a dedicated data structure, in case we want to configure data processing for each module, e.x. de/selecting specific detectors/monitors.

jokasimr:
> I made them stateful to prevent any overhead when processing the real-time messages
by arranging necessary information at the start and not again during the processing.

My thinking was that if this becomes relevant for performance we can make some wrapper that caches the return value of  `path_to_message_group`

SimonHeybrock:
> > > And the data structures were for making it clear what's stored where and how, and also easily add handles for certain part/attributes of the data that is often needed.
> > 
> > 
> > Can you give an example?
> 
> Like, if we just use dictionary, we don't know each item means.
> 
> So I wanted to wrap them, for example, the path to the stream modules and the type of the module. I thought it might be easier to store them in a dedicated data structure, in case we want to configure data processing for each module, e.x. de/selecting specific detectors/monitors.

You mean to avoid any overhead, even before passing this to the data reduction? Wouldn't that be handled by not subscribing to the Kafka topic for those detectors? In any case, maybe we can worry about that when/if we get there?

YooSunYoung:
> > > > And the data structures were for making it clear what's stored where and how, and also easily add handles for certain part/attributes of the data that is often needed.
> > > 
> > > 
> > > Can you give an example?
> > 
> > 
> > Like, if we just use dictionary, we don't know each item means.
> > So I wanted to wrap them, for example, the path to the stream modules and the type of the module. I thought it might be easier to store them in a dedicated data structure, in case we want to configure data processing for each module, e.x. de/selecting specific detectors/monitors.
> 
> You mean to avoid any overhead, even before passing this to the data reduction? Wouldn't that be handled by not subscribing to the Kafka topic for those detectors? In any case, maybe we can worry about that when/if we get there?

Users might want to just try turning on/off some detectors without connecting to the stream again.
Handling kafka clients are even more tedious/time-consuming regarding both human efforts and computing resources than just filtering messages...

But I just talked to Johannes offline and I think we can worry about them later as you said.

src/beamlime/applications/handlers.py
```diff
@@ -27,23 +34,50 @@ class WorkflowResultUpdate:
...
+
 class DataAssembler(HandlerInterface):
     """Receives data and assembles it into a single data structure."""
 
-    def __init__(self) -> None:
-        from ._nexus_helpers import NexusContainer
-
-        self.nexus_container: NexusContainer
+    def __init__(self, merge_every_nth=float('inf'), max_seconds_between_messages=5):
+        self._store = {}
```
YooSunYoung, line 56: Can you add more explanation what the `_store` is?
Or make it as a typed dict...?

jokasimr: It's a `Mapping[Tuple[str, ...], Mapping]`, a mapping from paths to nexus-json group objects.

It stores the nexus group objects created by the merged messages, and the paths where they should be inserted into the structure.

After a `DataReady` messages is emitted the store is reset so that the next `DataReady` message doesn't contain data from earlier messages.

jokasimr: I'll add that type annotation to the code

src/beamlime/applications/_nexus_helpers.py
jokasimr: Yeah I kind of have the same thought. Why can we not just use the stuff in `scippneutron` / `scippnexus` that deals with this nexus-json structure?

YooSunYoung: It was because we didn't want to depend on the `assumption` of the how the nexus structure looks like,
and scippnexus is kind of depending on the `assumption` of how it looks like.

For example, we should be able to flexibly patch the structure to be compatible with the scippnexus if needed in the production I thought.... since it's meant to be run as a real-time application.

tests/helpers/tests/applications/data.py
```diff
@@ -47,3 +49,17 @@ def get_checksum(name: str) -> str:
...
+    with open(get_path('ymir_detectors.json')) as f:
+        data = json.load(f)
+    return data
+
+
+@pytest.fixture
+def loki(large_file_test: bool):
+    assert large_file_test
+    with open(get_path('loki.json')) as f:
+        return json.load(f)
```
YooSunYoung, line 65: I think they should go to `conftest.py` instead.
This module is also used by documentations.

tests/applications/ymir_detectors.json
```diff
@@ -1576,8 +1576,7 @@
           ],
           "name": "daq_publication",
           "type": "group"
-        },
-        "$USERS$"
```
YooSunYoung, line 1580: Can you add a note of this modification in the related docstring...?

jokasimr: Do you mean in the docstring of the `ymir` fixture?

SimonHeybrock:
What remains to be done here? Can it get merge so we can move forward?

src/beamlime/applications/handlers.py
```diff
@@ -11,10 +13,21 @@
...
+from .daemons import (
+    ChopperDataReceived,
+    DetectorDataReceived,
+    LogDataReceived,
+    RunStart,
+)
 
 Events = NewType("Events", list[sc.DataArray])
+MergeMessageEveryNth = NewType("MergeMessageEveryNth", Number)
+'''Every n:th message the data assembler receives, the data reduction is run'''
```
YooSunYoung, line 27: Not so important but, how about keeping the names similar with `MergeMessageTimeInterval`...?
```suggestion
MergeMessageCountInterval = NewType("MergeMessageCountInterval", Number)
'''Every MergeMessageCountInterval-th message the data assembler receives, the data reduction is run'''
```

src/beamlime/applications/handlers.py
```diff
@@ -27,23 +40,59 @@ class WorkflowResultUpdate:
     content: WorkflowResult
 
 
+def nth_or_maxtime(n: Number, maxtime: Number):
+    if n <= 0:
+        raise ValueError('n must be positive')
+    if maxtime <= 0:
+        raise ValueError('maxtime must be positive')
```
YooSunYoung, line 47: Also here.

```suggestion
def maxcount_or_maxtime(maxcount: Number, maxtime: Number):
    if maxcount <= 0:
        raise ValueError('maxcount must be positive')
    if maxtime <= 0:
        raise ValueError('maxtime must be positive')
```

YooSunYoung (approved):
I don't have any complains except for some minor opinions on namings.
Feel free to ignore them.

src/beamlime/applications/handlers.py
jokasimr: I think this was a very good suggestion. I'll add a separate pr for those since this is already merged now.

src/beamlime/applications/handlers.py
jokasimr: Also much better.

PR #187, opened 2024-04-25
fix: plot update; default handler arguments to match old behavior

jokasimr:
Tried to run the example notebook and no figure was produced because the default config of the merge handler was to merge five messages, but the example only produces fewer messages so the workflow was never run.

Also encountered some error with the figure update function. Don't know what that was about, but now it seems fixed.

### DIFF
--- a/src/beamlime/applications/handlers.py
+++ b/src/beamlime/applications/handlers.py
@@ -66,8 +66,9 @@ class DataAssembler(HandlerInterface):
 
     def __init__(
         self,
-        merge_every_nth: MergeMessageCountInterval = float('inf'),
-        max_seconds_between_messages: MergeMessageTimeInterval = 5,
+        *,
+        merge_every_nth: MergeMessageCountInterval = 1,
+        max_seconds_between_messages: MergeMessageTimeInterval = float('inf'),
     ):
         self._store = {}
         self._should_send_message = maxcount_or_maxtime(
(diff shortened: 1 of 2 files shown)


src/beamlime/applications/handlers.py
```diff
@@ -66,8 +66,9 @@ class DataAssembler(HandlerInterface):
 
     def __init__(
         self,
-        merge_every_nth: MergeMessageCountInterval = float('inf'),
-        max_seconds_between_messages: MergeMessageTimeInterval = 5,
+        *,
+        merge_every_nth: MergeMessageCountInterval = 1,
+        max_seconds_between_messages: MergeMessageTimeInterval = float('inf'),
```
jokasimr, line 71: By default, run the workflow every time you receive a message.

src/beamlime/applications/handlers.py
```diff
@@ -135,7 +136,7 @@ def plot_item(self, name: str, data: sc.DataArray) -> None:
             self.artists[name] = next(iter(plot.artists))
             self.figures[name] = plot
         else:
-            figure.update(data, key=self.artists[name])
```
jokasimr, line 138: is this old plopp code?

SimonHeybrock: @nvaytet?

YooSunYoung: Probably...?

YooSunYoung: Yeah, it was...! I'll update the plopp version in the dependencies : D

YooSunYoung (approved):

PR #190, opened 2024-04-26
Dependency updates

YooSunYoung:

PR #192, opened 2024-05-03
Update fixed dependencies.

YooSunYoung:
Continued from #190 

We need to update `plopp`


YooSunYoung:
I'll just merge it since it's just a version upgrades of dependencies.

PR #194, opened 2024-05-08
Grid plots from multiple figures.

### LINKED ISSUE #136: Default visualization
The result of the data reduction needs to be visualized. There should be a default visualization option that does not require any configuration.

- Depending on number of results, choose a reasonable grid layout.
- Use `plopp` to plot.
- Note #159, which returns `I(Q)` that should be plottable. Use this for (manual) testing of the visualization.

YooSunYoung:
Fixes #136 

`PlotSaver`/`PlotStreamer` now make a grid of plots, if there are multiple figures to render.
For easier testing, I made the dummy workflow to return multiple data-arrays.

# In terminal

https://github.com/scipp/beamlime/assets/17974113/a096d061-9d11-47b1-a326-a850f67345de

# In Jupyter notebook

https://github.com/scipp/beamlime/assets/17974113/2f5b2863-7557-4ad4-9de9-75bd9a565b13

### DIFF
--- a/src/beamlime/applications/handlers.py
+++ b/src/beamlime/applications/handlers.py
@@ -119,17 +119,27 @@ def random_image_path() -> ImagePath:
     return ImagePath(pathlib.Path(f"beamlime_plot_{uuid.uuid4().hex}"))
 
 
+MaxPlotColumn = NewType("MaxPlotColumn", int)
+DefaultMaxPlotColumn = MaxPlotColumn(2)
+
+
 class PlotStreamer(HandlerInterface):
-    def __init__(self, logger: BeamlimeLogger) -> None:
+    def __init__(
+        self,
+        *,
+        logger: BeamlimeLogger,
+        max_column: MaxPlotColumn = DefaultMaxPlotColumn,
+    ) -> None:
         self.logger = logger
         self.figures = {}
         self.artists = {}
+        self.max_column = max_column
         super().__init__()
 
     def plot_item(self, name: str, data: sc.DataArray) -> None:
         figure = self.figures.get(name)
         if figure is None:
-            plot = data.plot()
+            plot = data.plot(title=name)
             # TODO Either improve Plopp's update method, or handle multiple artists
             if len(plot.artists) > 1:
                 raise NotImplementedError("Data with multiple items not supported.")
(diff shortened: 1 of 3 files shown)


src/beamlime/applications/handlers.py
```diff
@@ -143,23 +153,55 @@ def update_histogram(self, message: WorkflowResultUpdate) -> None:
...
+                    HBar(
+                        figures[
+                            i_row * self.max_column : i_row * self.max_column
+                            + self.max_column
+                        ]
+                    )
+                    for i_row in range(n_rows)
+                ]
+            )
+        )
```
nvaytet, line 174: I think you don't need the `VBar` and `HBar`.
You should be able to use nested lists, e.g. `Box([[f1, f2, f3], [f4, f5]])` would give you 2 rows with 3 and 2 figs on each row, respectively.

src/beamlime/applications/handlers.py
```diff
@@ -143,23 +153,55 @@ def update_histogram(self, message: WorkflowResultUpdate) -> None:
...
-        self.info("Received histogram, saving into %s...", self.image_path_prefix)
-        for name, figure in self.figures.items():
-            figure.save(f'{self.image_path_prefix}-{name}.png')
+        image_file_name = f"{self.image_path_prefix}.png"
+        self.info("Received histogram(s), saving into %s...", image_file_name)
+        plot_tile = tiled((len(self.figures) // self.max_column) + 1, self.max_column)
+        for i_fig, figure in enumerate(self.figures.values()):
+            plot_tile[i_fig // self.max_column, i_fig % self.max_column] = figure
+
+        plot_tile.save(image_file_name)
```
nvaytet, line 200: I probably missed that discussion, but when using this from the command line, why are we saving png files instead of opening a qt window which would show the matplotlib figure? Or maybe we are doing that somewhere else and that should also be updated?

YooSunYoung: The default visualisation should do bare-minimum I thought.
Dumping a image file was the minimum since we can just use any viewers, without any dependencies of GUI libraries.
So I didn't really think about having QT window here.
Do you think we should...?

nvaytet: I guess what I meant was calling Matplotlib's `show` on a figure and it opens the window for you.

YooSunYoung: How should we do that...? 
I tried simply calling `plt.show()` but it doesn't do anything in the terminal though...

nvaytet: Maybe it's because you're in VScode? https://stackoverflow.com/a/65086012

YooSunYoung: Probably.
I think the default Visualisation should work in any environment,
so if it requires any sort of effort, I'd go for dumping a file.

Also, I'm not sure how to only-show the grid-view of all plots, not the individual ones.

nvaytet: Have you tried just from a terminal to make sure that this is indeed the issue? If it is, then I agree with you.

> Also, I'm not sure how to only-show the grid-view of all plots, not the individual ones.

Not sure I understood exactly what you meant here...

YooSunYoung: > Have you tried just from a terminal to make sure that this is indeed the issue? If it is, then I agree with you.

I just tried, and it didn't pop up the plot... 
it only works when I created plot directly using `pyplot`.
And whenever the window pops up, it blocks the process, and if I use `block=False` option, it just doesn't show up at all... 

> > Also, I'm not sure how to only-show the grid-view of all plots, not the individual ones.
> 
> Not sure I understood exactly what you meant here...

I just don't know how to update the plot display with the new one, without making a new window...
And if you plot sth with `pyplot` and then use `plt.show()`, it will show all the plots you have drawn before. But we only want one grid plots, without the individual ones.

nvaytet: > it only works when I created plot directly using pyplot.

Yeah, it seems there is something wrong with getting figures to show from the terminal. I am getting this error
```
AttributeError: Figure.show works only for figures managed by pyplot, normally created by pyplot.figure()
```
Is that also what you got?

I opened https://github.com/scipp/plopp/issues/342


For the second part, if we can resolve the usage of `fig.show()` above, then we should be able to just show the figrues we want instead of everything.

nvaytet: In any case, as all of the above requires more work, we can keep it as png images for now, and always add a new interactive handler later.

nvaytet (approved):

PR #195, opened 2024-05-15
Fix random generator,  missing numpy array wrapping

### LINKED ISSUE #193: Things to fix
As I tried to test `esssans` workflow https://github.com/scipp/esssans/pull/134,

I found some bugs in the nexus helper and the random data generator.

- [x] dtype parsing from a message.

https://github.com/scipp/beamlime/blob/66c9efd9e1cc324b936452c75aff368f552c8ec3/src/beamlime/applications/_nexus_helpers.py#L8-L17

Numpy dtype doesn't match with the nexus dtype. So I suggest to update it to:
```python

def _dtype_to_nexus_dtype(dtype: type):
    if dtype == str:
        return 'string'
    if dtype == int:
        return 'int'
    if dtype == float:
        return 'float'
    raise ValueError(f"Unsupported dtype: {dtype}")


def _dtype(value):
    """Return the dtype of a numpy array or the type of a scalar value."""
    if hasattr(value, 'dtype'):
        return _dtype_to_nexus_dtype(value.dtype)
    else:
        return _dtype_to_nexus_dtype(type(value))
```

- [x] numpy array wrapping missing

https://github.com/scipp/beamlime/blob/66c9efd9e1cc324b936452c75aff368f552c8ec3/src/beamlime/applications/_random_data_providers.py#L51-L52

In the production, the message de-serializer will convert them into numpy array so it doesn't matter.

```python
reference_time=np.array([et_zero]),
reference_time_index=np.array([0]),
```

It's related to the dtype parsing above.

- [x] unit missing when we merge messages

When we initialize the dataset, we also need to insert units.
That's why I hard-coded the recipes per schema before.
The units are mentioned in the comment lines of the schemas but we can't get that information else where...
So we either get back the recipes, or we should insert some if-statements that handle the units.
Units are attributes in the nexus structure, that should be inserted like this:
> Note that the attributes are a list of dictionaries(each attribute).
```python
dataset['attributes'] = [dict(
                name='units',
                dtype='string',
                values='ns'
            )]
```
- [x] Merging decides to concatenate/stack/replace based on the field name

It was missing some fields to concatenate the values. I think it shouldn't depend on the names of the fields at all.
It should concatenate if it's possible, or replace it based on the values type.

https://github.com/scipp/beamlime/blob/66c9efd9e1cc324b936452c75aff368f552c8ec3/src/beamlime/applications/_nexus_helpers.py#L35-L43

- [x] Merging concatenate the initial data twice.

https://github.com/scipp/beamlime/blob/66c9efd9e1cc324b936452c75aff368f552c8ec3/src/beamlime/applications/_nexus_helpers.py#L39-L43

After we initialize the dataset, we don't know if we retrieved the existing dataset or we create the new dataset with the new `value` but here it concatenates the new `value` anyways.
We probably have to update how it initializes dataset. Also related to the previous point.

- [x] Field name mapping from schema to nexus.

https://github.com/scipp/beamlime/blob/66c9efd9e1cc324b936452c75aff368f552c8ec3/src/beamlime/applications/_nexus_helpers.py#L119-L120

Unfortunately, the name of the fields in the schema doesn't match with the nexus file...!
For example, `reference_time` should be `event_time_zero`.

It should be translated before we merge.

```python
    if kind == 'ev44':
        content['event_time_zero'] = content.pop('reference_time')
        content['event_index'] = content.pop('reference_time_index')
        content.pop('source_name')
        content['event_time_offset'] = content.pop('time_of_flight')
        _merge_ev44(store[path], content)
```

- [x] Typo in the random generator and too small time offsets.

This is not really a bug, but for generating realistic dataset, we should fix these lines:
https://github.com/scipp/beamlime/blob/66c9efd9e1cc324b936452c75aff368f552c8ec3/src/beamlime/applications/_random_data_providers.py#L44

should be:

```python
et_zero = ReferenceTimeZero(13620492*10**11)  # No reason
```
and

https://github.com/scipp/beamlime/blob/66c9efd9e1cc324b936452c75aff368f552c8ec3/src/beamlime/applications/_random_data_providers.py#L53

should be:

```python
time_of_flight=rng.random((cur_event_number,)) * 48_000_000 + 4_000_000,  # No reason
```

YooSunYoung:
Partially-fixes #193, numpy array wrapping missing

I also added/modified some unit-tests.

### DIFF
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -39,7 +39,7 @@ def _merge_ev44(group, message):
         dataset = _init_dataset(group, field, value)
         if field in ('time_of_flight', 'pixel_id'):
             dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], value)
+                (dataset['config']['values'], np.asarray(value))
             )
 
 
@@ -47,7 +47,9 @@ def _merge_f144(group, message):
     for field, value in message.items():
         dataset = _init_dataset(group, field, value)
         if field in ('timestamp', 'value'):
-            dataset['config']['values'] = np.stack((dataset['config']['values'], value))
+            dataset['config']['values'] = np.stack(
+                (dataset['config']['values'], np.asarray(value))
+            )
 
 
 def _merge_tdct(group, message):
@@ -55,7 +57,7 @@ def _merge_tdct(group, message):
         dataset = _init_dataset(group, field, value)
         if field in ('timestamps',):
             dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], value)
+                (dataset['config']['values'], np.asarray(value))
             )
 
 
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -39,7 +39,7 @@ def _merge_ev44(group, message):
         dataset = _init_dataset(group, field, value)
         if field in ('time_of_flight', 'pixel_id'):
             dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], np.asarray(value))
+                (dataset['config']['values'], value)
             )
 
 
@@ -47,9 +47,7 @@ def _merge_f144(group, message):
     for field, value in message.items():
         dataset = _init_dataset(group, field, value)
         if field in ('timestamp', 'value'):
-            dataset['config']['values'] = np.stack(
-                (dataset['config']['values'], np.asarray(value))
-            )
+            dataset['config']['values'] = np.stack((dataset['config']['values'], value))
 
 
 def _merge_tdct(group, message):
@@ -57,7 +55,7 @@ def _merge_tdct(group, message):
         dataset = _init_dataset(group, field, value)
         if field in ('timestamps',):
             dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], np.asarray(value))
+                (dataset['config']['values'], value)
             )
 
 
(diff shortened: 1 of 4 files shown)


src/beamlime/applications/_nexus_helpers.py
```diff
@@ -39,23 +39,25 @@ def _merge_ev44(group, message):
         dataset = _init_dataset(group, field, value)
         if field in ('time_of_flight', 'pixel_id'):
             dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], value)
+                (dataset['config']['values'], np.asarray(value))
```
jokasimr, line 42: When does the absence of `np.asarray` cause an issue here?

YooSunYoung: Hmm actually it doesn't matter here since concatenate will make it into numpy ndarray anyways.

src/beamlime/applications/_random_data_providers.py
```diff
@@ -72,8 +73,8 @@ def nxevent_data_ev44_generator(
     for i, (start, end) in enumerate(zip(event_index[:-1], event_index[1:])):
         yield EV44(
             source_name=source_name,
-            reference_time=event_time_zero[i : i + 1],
-            reference_time_index=[start],
-            time_of_flight=event_time_offset[start:end],
-            pixel_id=None if event_id is None else event_id[start:end],
+            reference_time=np.asarray(event_time_zero[i : i + 1]),
+            reference_time_index=np.asarray([start]),
```
jokasimr, line 77: This should just be `np.asarray([0])`. The `reference_time_index` is always `[0]`.

jokasimr (commented):
Added one small comment about `reference_time_index` that needs to be addressed otherwise everything looks good to me!

src/beamlime/applications/_random_data_providers.py
YooSunYoung: Good point! Thank you!

jokasimr (approved):

PR #196, opened 2024-05-21
Fix nexus helper

### LINKED ISSUE #193: Things to fix

YooSunYoung:
Fixes #193


YooSunYoung:
We reviewed this offline with @jokasimr 
and the conclusion was to separate 
- `field-name-mapping` f.x. `pixel_id` to `event_id`
- `initilization of a dataset`
- `merge (concatenating)` values

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+
+    module: str
+    config: dict[str, Any]
+    attributes: list[Mapping[str, Any]]
+
+
+NexusPath = tuple[str | None, ...]
+
+
+class DataModuleParent(TypedDict):
```
jokasimr, line 28: I think this should be called `NexusGroup`

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+    since the module place holder is one of the children of it.
+    """
+
+    children: list[Mapping]
+    """A module place holder or datasets."""
+    name: str
+    """Name of the group."""
+
+
+DataModuleStore = dict[NexusPath, DataModuleParent]
```
jokasimr, line 41: I think this should be called `ModuleStore` or `NexusGroupStore`.
`DataModule` is more convoluted than simply `Module`, but it doesn't really communicate more than just `Module`. `NexusGroupStore` is longer, but explains a bit more what the type is.

If we want to let the name indicate that it's a store for the streams then  maybe `StreamStore` or `StreamedGroupStore` are other good options.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
-        if field in ('timestamps',):
-            dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], value)
+    - event_time_zero
+    - event_time_offset
+    - event_index
+    - event_id
+
+    This function doesn't remove a place holder from the ``children``.
+    It is expected to be removed outside of this function.
```
jokasimr, line 99: I don't understand what this means? What place holder should be removed?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+
+    """
+    children = group.setdefault("children", [])
```
jokasimr, line 102: This can be removed if we make sure to always initialize the nexus group with a `children` list

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+            dtype="int",
+            initial_values=np.asarray([]),
+            unit="ns",
+        )
+    )
+    # event_time_offset
+    children.append(
+        create_dataset(
+            name="event_time_offset",
+            dtype="int",
```
jokasimr, line 116: we should specify the right `dtype` here as specified in `nexus-streaming-types`, that is, set `int32` or `int64`.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+    # event_index
+    children.append(
+        create_dataset(
+            name="event_index",
+            dtype="int",
+            initial_values=np.asarray([]),
+        )
+    )
+    # event_id
+    if "monitor" in group["name"]:
```
jokasimr, line 130: Isn't it better to look at `data_part` and only set `event_id` if `pixel_id` is not `None`?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+
+    """
+    # event_time_zero - reference_time
+    event_time_zero_dataset = find_nexus_structure(group, ("event_time_zero",))
+    original_event_time_zero = _get_array_values(event_time_zero_dataset)
+    new_event_time_zero = data_piece["reference_time"]
+    _set_values(
+        event_time_zero_dataset,
+        np.concatenate((original_event_time_zero, new_event_time_zero)),
+    )
```
jokasimr, line 169: This looks verbose to me, how about rewriting it like this?
```python
ds = find_nexus_structure(group, ("event_time_zero",))
ds["config"]["values"] = np.concatenate((ds["config"]["values"], data_piece["reference_time"])
```

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+    event_index_dataset = find_nexus_structure(group, ("event_index",))
+    original_event_index = _get_array_values(event_index_dataset)
+    event_index_offset = len(original_event_time_offset)
+    new_event_index = data_piece["reference_time_index"] + event_index_offset
+    _set_values(
+        event_index_dataset, np.concatenate((original_event_index, new_event_index))
+    )
+    # event_id - pixel_id
+    if (
+        "pixel_id" in data_piece and data_piece["pixel_id"] is not None
```
jokasimr, line 188: (minor) `data_piece.get("pixel_id")` is shorthand for the check in the if statement

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
+        unexpectedly-not-receiving data.
+    merge_func:
+        A function that merges the message into the store.
+
+    Side Effects
+    ------------
+    The ``data_module_store`` is updated with the merged data.
+
+    """
+    for path in path_matching_func(structure, data_piece):
```
jokasimr, line 277: I don't understand this. Should the message be merged in multiple places?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,365 @@
...
-    if (None,) in store:
-        return store[(None,)]
+    if (None,) in data_module_store:
+        return data_module_store[(None,)]
 
-    new = structure.copy()
-    if 'children' in structure:
+    new = {**structure}  # Faster than `dict()`
+    # and avoid static type-error of `copy`
+    # since `structure` is `Mapping`
```
jokasimr, line 350: (minor) I think `{**structure}` is absolutely fine as a replacement for `.copy`. However, maybe we should just change the argument type of `structure` to `dict` anyway. It is a dict, after all. It's up to you!

In any case, the comments explaining this are not necessary.

jokasimr (commented):
This is starting to look really good! Left some comments.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: https://github.com/scipp/beamlime/blob/81dfc0b6c75f5522769d811f677cf8f34b63c711/src/beamlime/applications/_nexus_helpers.py#L286-L289

Here you drop the place holder that looks like 

```python
{
      "module": "ev44",
      "config": {
        "topic": "loki_detector",
        "source": "loki_1"
      }
}
```

I'll either add more explanation or remove it : D...

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Hmm... yeah. 
Does it mean that we use `int64` if they are `long` and `int32` if they are just `int` in the schema?
For example:

https://github.com/ess-dmsc/streaming-data-types/blob/13dd32d1aa7f8eb47cbb1393918e6929f9890f17/schemas/ev44_events.fbs#L5C1-L20C2

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: I'll just remove it since there is a comment in ``_merge_message_into_data_module_store``.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: I thought `initialize` function shouldn't depend on the data it accepts.
But I looked at other schemas and there are quite a few optional fields as well... so I'll do that.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Some of them are used in other places so I'll reduce them as much as possible.

src/beamlime/applications/_nexus_helpers.py
jokasimr: I'm still not sure I understand. The `NexusGroup` should not have a place holder. The place holder is just something that exists in the nexus template.

src/beamlime/applications/_nexus_helpers.py
jokasimr: Yes I think that makes sense.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: But I would keep the helper to set/get the values. 
I think we should avoid using literal `string` keys as much as possible.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Yes... unfortunately....

src/beamlime/applications/_nexus_helpers.py
jokasimr: Hm good point. Not sure what's best. Maybe we can use the first message in the initialization to determine the optional fields, but not populate the fields with data.

src/beamlime/applications/_nexus_helpers.py
jokasimr: This is news to me, can you give an example just so I understand how it's supposed to work?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,378 @@
...
-        if value is None:  # Could be monitor without pixel_id
-            continue
-        dataset = _init_dataset(group, field, value)
-        if field in ('time_of_flight', 'pixel_id'):
-            dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], value)
+ModuleNameType = Literal["ev44"]  # "f144", "tdct" will be added in the future
+"""Name of the module that is supported by beamlime."""
+
+DeserializedMessage = Mapping
```
jokasimr, line 75: I don't think we need to specify that it is a deserialized message. It's enough to say that it is a message, or maybe `StreamMessage` if we want to specify that it is part of a module strem.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,378 @@
...
+"""Deserialized message from one of the schemas bound to :attr:`~ModuleNameType`."""
+
+
+def _get_array_values(dataset: Mapping) -> np.ndarray:
+    return dataset["config"]["values"]
+
+
+def _set_values(dataset: Mapping, values: Any) -> None:
+    dataset["config"]["values"] = values
```
jokasimr, line 84: I don't think we need this getter and setter. Let's just explicitly write 
```python
dataset["config"]["values"]
```
and
```python
dataset["config"]["values"] = values
```
in the merge methods.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,378 @@
...
+) -> NexusGroupDict:
+    # TODO: We can use typeguard here later.
+    return find_nexus_structure(structure, path[:-1])
 
-def path_to_message_group(structure: Mapping, message: Mapping) -> str:
-    '''Finds the group where the message should be merged'''
-    kind, content = message
+
+def find_ev44_matching_paths(
+    structure: NexusDictStructure, data_piece: DeserializedMessage
```
jokasimr, line 241: (minor) Maybe `NexusTemplate` or `NexusStructure` instead of `NexusDictStructure`, the editor can tell us it is a dictionary already

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,378 @@
...
+        A function that merges the message into the store.
+
+    Side Effects
+    ------------
+    The ``nexus_group_store`` is updated with the merged data.
+
+    """
+    for path in path_matching_func(structure, data_piece):
+        if path not in nexus_group_store:
+            parent = find_parent(structure, path)
```
jokasimr, line 294: (minor) Why make `find_parent` a separate function if it is only used here? In my opinion that just creates unnecessary indirection. The variable name already mentions that the value is the 'parent'.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -1,155 +1,360 @@
...
+    - event_id
 
-def _merge_tdct(group, message):
-    for field, value in message.items():
-        dataset = _init_dataset(group, field, value)
-        if field in ('timestamps',):
-            dataset['config']['values'] = np.concatenate(
-                (dataset['config']['values'], value)
+    """
+    children = group.setdefault("children", [])
```
jokasimr, line 99: Like I mentioned earlier I think it's better to not create the `children` here if we are already doing it elsewhere.
Doing it in two places creates confusion about whose responsibility it is to initialize it. It's better that we explicitly do not create it here, and if it isn't created elsewhere then this merge routine will just fail.

YooSunYoung: Hmm yeah. It won't create a children if it exists but might be still confusing.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: It was just for the types... I'm trying to pass the mypy as much as possible...

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: How would you know if it's a dictionary...?

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: I fixed it...!

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: I find it confusing because there are three types of messages:
- deserialized `flat_buffer`
- kafka message
- message for a beamlime message router

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Ah you mean by the type? hmm okay we can keep the name simple.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: I thought it's worth mention that it's deserialized flat buffer.

src/beamlime/applications/_nexus_helpers.py
jokasimr: Why don't mypy let it pass without the function?

src/beamlime/applications/_nexus_helpers.py
jokasimr: Okay then let's keep it as it is

jokasimr (approved):

PR #197, opened 2024-05-22
Add line breaks in a title for long names.

YooSunYoung:
Follow up fixes for #136 

We use the domain type names as titles for default visualization and they are expected to be long.

test

So I just added a line break if it has `[]`.
test


YooSunYoung:
I'll just merge it since it's trivial.

PR #205, opened 2024-05-22
Fix rest ruff errors.

### LINKED ISSUE #200: Make ruff happy
## PT*
- PT001 Use `@pytest.fixture()` over `@pytest.fixture`
- PT011 `pytest.raises(ValueError)` is too broad, set the `match` parameter or use a more specific exception
- PT019 Fixture `_datetime` without value is injected as parameter, use `@pytest.mark.usefixtures` instead
- PT006 Wrong type passed to first argument of `@pytest.mark.parametrize`; expected `tuple`
- PT013 Incorrect import of `pytest`; use `import pytest` instead
- PT007 Wrong values type in `@pytest.mark.parametrize` expected `list` of `tuple`
- PT003 `scope='function'` is implied in `@pytest.fixture()`
- PT018 Assertion should be broken down into multiple parts

## RUF*
- RUF015 Prefer `next(...)` over single element slice
- RUF001 String contains ambiguous `➕` (HEAVY PLUS SIGN). Did you mean `+` (PLUS SIGN)? -> ignore
- RUF013 PEP 484 prohibits implicit `Optional`
- RUF005 Consider `(*group_path, group_path[-1] + "_events")` instead of concatenation
- RUF007 Prefer `itertools.pairwise()` over `zip()` when iterating over successive pairs
- RUF009 Do not perform function call `TMI` in dataclass defaults

## B*
- B905 `zip()` without an explicit `strict=` parameter
- B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling

## C*
- C419 Unnecessary list comprehension
- C408 Unnecessary `list` call (rewrite as a literal)
- C408 Unnecessary `dict` call (rewrite as a literal)
- C416 Unnecessary `list` comprehension (rewrite using `list()`)
- C403 Unnecessary `list` comprehension (rewrite as a `set` comprehension)

## D*
- DTZ003 `datetime.datetime.utcnow()` used
- 

## E*
-  E401 Multiple imports on one line
-  E714 Test for object identity should be `is not`

## F*
- F811 Redefinition of unused `FakeListener` from cell 4, line 16
- F401 `pandas` imported but unused

## G*
- G004 Logging statement uses f-string

## S*
- S603 `subprocess` call: check for execution of untrusted input -> It's in a benchmark helper. Gonna ignore this.
- S311 Standard pseudo-random generators are not suitable for cryptographic purposes -> It's in a example notebook for benchmark testing. Will ignore this.
- S324 Probable use of insecure hash functions in `hashlib`: `md5`

## PYI*
- Use `typing.NamedTuple` instead of `collections.namedtuple`
-  PYI041 Use `float` instead of `int | float`

## PERF*
- PERF203 `try`-`except` within a loop incurs performance overhead. -> ignoring this since it doesn't make sense to move try-except to outer range.
- PERF401 Use a list comprehension to create a transformed list

## Else
- W291 Trailing whitespace

## UP rules
- UP035 Import from `collections.abc` instead: `Generator`
- UP007 Use `X | Y` for type annotations
- UP014 Convert `ProcessCpuAffinity` from `NamedTuple` functional to class syntax
- UP028 Replace `yield` over `for` loop with `yield from`
- UP035 Import from `collections.abc` instead: `Generator`, `Mapping`
- UP034 Avoid extraneous parentheses
- UP006 Use `type` instead of `Type` for type annotation
- UP006 Use `dict` instead of `Dict` for type annotation

YooSunYoung:
Related to #200 

**Please don't merge anything until all the reviews are done related to ruff.**

## D*
- DTZ003 `datetime.datetime.utcnow()` used

## E*
-  E401 Multiple imports on one line
-  E714 Test for object identity should be `is not`

## F*
- F811 Redefinition of unused `FakeListener` from cell 4, line 16
- F401 `pandas` imported but unused

## G*
- G004 Logging statement uses f-string

## S*
- S603 `subprocess` call: check for execution of untrusted input -> It's in a benchmark helper. Gonna ignore this.
- S311 Standard pseudo-random generators are not suitable for cryptographic purposes -> It's in a example notebook for benchmark testing. Will ignore this.
- S324 Probable use of insecure hash functions in `hashlib`: `md5`

## PYI*
- Use `typing.NamedTuple` instead of `collections.namedtuple`
-  PYI041 Use `float` instead of `int | float`

## PERF*
- PERF203 `try`-`except` within a loop incurs performance overhead. -> ignoring this since it doesn't make sense to move try-except to outer range.
- PERF401 Use a list comprehension to create a transformed list

## Else
- W291 Trailing whitespace


src/beamlime/constructors/inspectors.py
```diff
@@ -127,9 +128,12 @@ def __init__(self, dependency_type: Any, default_value: Product) -> None:
 
     @staticmethod
     def extract_dependency_type(dependency_type: Any) -> None:
+        from types import UnionType
         from typing import Union, get_args, get_origin
 
-        if get_origin(dependency_type) == Union:
+        if get_origin(dependency_type) == Union or isinstance(
+            dependency_type, UnionType
+        ):
```
YooSunYoung, line 136: This is a bug I found thanks to ruff : D

tests/constructors/provider_test.py
```diff
@@ -119,11 +119,19 @@ def local_function(): ...
...
-def func_with_union_return() -> Union[int, float, None]:
+def func_with_explicit_union_arg(_: Union[float, str, None]) -> int | str:
+    return 0
+
+
+def func_with_union_return() -> int | str | None:
+    return None
+
+
+def func_with_explicit_union_return() -> Union[int, str, None]:
```
jl-wynen, line 134: Does this really make a difference compared to the pipe syntax? 

In any case, you should suppress the relevant `UP` rule here.

YooSunYoung: Yes, if you make Union with `Union`, the `get_origin` returns `Union` but if you make it with the pipe syntax, `get_origin` returns `UnionType`. 

```python
>>> from typing import get_origin, Union 
>>> get_origin(int | str)
<class 'types.UnionType'>
>>> get_origin(Union[int, str])
typing.Union
```

What do you mean by the `UP` rule though?

YooSunYoung: PYI041 Use float instead of int | float
This one you mean?

jl-wynen: No, I meant this one: https://docs.astral.sh/ruff/rules/non-pep604-annotation/ It is not enabled in our setup but I used the pyupgrade rules to upgrade the code in other repos. Suppressing this here ensures that we don't accidentally change the function in the future.

YooSunYoung: Apparently it's not allowed to suppress unused rule!
`RUF100 Unused `noqa` directive (non-enabled: `UP007`)`
I'll just enable `UP` and fix it there...! Thanks : D

jl-wynen: You would also have to suppress `RUF100` in the same line.

You are welcome to enable `UP`. I didn't do that in the template because it does not always lead to desirable code. E.g., https://docs.astral.sh/ruff/rules/non-pep604-isinstance/ reduces performance. So I would at least not enable auto fixes for all `UP` rules.

jl-wynen (approved):

tests/constructors/provider_test.py
YooSunYoung: Done in https://github.com/scipp/beamlime/blob/fba4b3ddf17a235ae73adc8f696d37785e6a2c8a/tests/constructors/factory_test.py#L66-L77
and
https://github.com/scipp/beamlime/blob/fba4b3ddf17a235ae73adc8f696d37785e6a2c8a/tests/constructors/factory_test.py#L120-L133

jl-wynen (approved):

PR #211, opened 2024-06-24
feat: add support for f144 and tdct message types

### LINKED ISSUE #171: Handle ``f144(log)`` and ``tdct(chopper)``.

jokasimr:
Fixes #171 

~Draft because needs tests~

There are quite a lot of changes, this is because there were some assumptions in the old code that did not hold for the new streams.

The old code assumed a module is the single child of a group, and all data sets in the group are defined by the content of the stream. This is not the case for the `tdct` stream, it only defines a single data set.
Now the code is more agnostic to if a stream defines one or several data sets.

### DIFF
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -326,7 +326,8 @@ def _merge_message_into_nexus_store(
     data:
         The content of the message.
     find_insert_paths:
-        A function that returns the paths that match the message.
+        A function that returns the paths in the structure where the
+        message should be merged.
     initialize:
         A function that initializes the dataset / group.
         *Initialize is done only when the relevant message arrives.*
@@ -348,6 +349,7 @@ def _merge_message_into_nexus_store(
 
 
 def merge_message_into_nexus_store(
+    *,
     structure: NexusStructure,
     nexus_store: NexusStore,
     data: DeserializedMessage,
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -429,7 +429,7 @@ def combine_nexus_store_and_structure(
             if len(path) == 1:
                 # Don't replace existing values, they were added in the previous step
                 try:
-                    find_nexus_structure(structure, path)
+                    find_nexus_structure(new, path)
                 except KeyError:
                     new["children"].append(value)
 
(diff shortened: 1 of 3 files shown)


jokasimr:
Looked now at the bifrost files being written and it seems that the `tdct` modules do actually become nexus groups in the files, not individual datasets, despite them not being the single child of a group.
Therefore there seems to be something about how the templates are to be interpreted that I've misunderstood.
I'll talk to Jonas from ECDC today to try and clear this up.

If this is the case it will not require large changes, it might event simplify some things. But it will need changes.

jokasimr:
Okay now I spoke to ECDC and it seems what we're doing here is correct.
There's documentation for this stuff here: https://gitlab.esss.lu.se/ecdc/ess-dmsc/kafka-to-nexus/-/tree/main/documentation?ref_type=heads

Good to know!

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -238,51 +319,38 @@ def _merge_message_into_nexus_group_store(
...
-        # Merge data piece
-        merge_func(nexus_group_store[path], data_piece)
+    for path in find_insert_paths(structure, data):
+        if path not in nexus_store:
+            nexus_store[path] = initialize(structure, path, data)
+        merge_message(cast(DatasetOrGroup, nexus_store[path]), data)
 
 
-def merge_message_into_nexus_group_store(
-    *,
```
YooSunYoung, line 282: We typically make functions keyword-only if it has more than 1-2 arguments.
I think this one especially should be keyword-only, since `structure` and `nexus_store` and `data` can all be dictionaries.

tests/applications/nexus_helpers_test.py
```diff
@@ -255,3 +247,152 @@ def test_nxevent_data_ev44_generator_yields_frame_by_frame() -> None:
...
+    assert len(result['children']) == 2
+    log = find_nexus_structure(result, ('the_log_name',))
+    assert len(log['children']) == 2
+    tdct = find_nexus_structure(
+        result,
+        (
+            'chopper_3',
+            'top_dead_center',
+        ),
+    )
```
YooSunYoung, line 398: This one will be overwritten by the `tdct` variable of the `for` loop and not used for any tests.... right...?

YooSunYoung:
I just had a couple of comments but it looks good to me overall : D

YooSunYoung (approved):

PR #216, opened 2024-07-25
Proposed new workflow interface

SimonHeybrock:
This is incomplete, but this is what we think could work.

To do (not in Beamlime):
- Add alternative provider that can load event data directly from group, instead of group nested in entry/instrument/detector


YooSunYoung:
I'll just close it since it was included in #219

PR #218, opened 2024-08-14
Add plugin nightly test.

YooSunYoung:
**The CI will be failing until #219 is merged**

It's similar to `nightly` but `nightly` tests are for upstream software that `beamlime` uses, but this one is for the downstream software that uses `beamlime`.

So downstream softwares that implement `beamlime` plugins
should add themselves in the `requirements/plugin.in` 
so that `beamlime` can test all the plugins every night.


tests/applications/test_stateless_workflow.py
YooSunYoung, line 1: Without the plugin packages installed, it'll only test the `dummy` plugin that is included in `beamlime`.

YooSunYoung:
I'll merge it since the key part is approved already in #219

PR #220, opened 2024-08-16
New interface adapt

YooSunYoung:
Instead of handling nested dictionary of `nexus_structure`, (most of static information will be in the skeleton file)
now we just need to collect the `module` placeholders (and their parents) from the `nexus_structure` at the beginning of the experiment(from the `run_start` message)
and then keep them in a dictionary (`_store` in the assemble handler)


SimonHeybrock (commented):
Some questions, maybe easiest if you walk me through the changes?

src/beamlime/applications/daemons.py
```diff
@@ -46,16 +58,14 @@ class ChopperDataReceived:
...
+NexusStructurePath = NewType("NexusStructurePath", str)
 
 EventDataSourcePath = NewType("EventDataSourcePath", str)
 
 
-def read_nexus_template_file(path: NexusTemplatePath) -> NexusTemplate:
+def read_nexus_template_file(path: NexusStructurePath) -> NexusGroup:
     with open(path) as f:
-        return NexusTemplate(json.load(f))
+        return NexusGroup(json.load(f))
```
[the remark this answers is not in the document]
YooSunYoung: No, we need to collect the module place holders...?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -243,50 +229,51 @@ def _merge_f144(group: NexusGroup, data: DeserializedMessage) -> None:
...
-) -> Iterable[NexusPath]:
-    name = data["name"]
-    for path, node in iter_nexus_structure(structure):
-        if node.get("name") == name:
-            for _, child in iter_nexus_structure(node):
-                if child.get("module") == "tdct":
-                    yield (*path, 'top_dead_center')
+    modules: ModuleRegistry, data: DeserializedMessage
+) -> Generator[tuple[str, NexusGroup]]:
+    raise NotImplementedError("TDCT is not supported yet.")
```
jokasimr, line 234: Why is the support for TDCT messages removed?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -243,50 +229,51 @@ def _merge_f144(group: NexusGroup, data: DeserializedMessage) -> None:
...
+    dtype = module["config"].get("dtype", "uint64")
+    group['children'] = [
+        create_dataset(
+            name=data["name"],
+            dtype=dtype,
+            initial_values=np.asarray([], dtype=dtype),
+            unit=module["config"].get("unit", "ns"),
+        )
+    ]
+    return group
```
jokasimr, line 250: I'm pretty sure that `tdct` is only a dataset and not a group (unlike `ev44` and `f144`). That's why it was implemented the way it was.

See https://manual.nexusformat.org/classes/base_classes/NXdisk_chopper.html

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -377,63 +370,63 @@ def merge_message_into_nexus_store(
...
     then the data of the rest of detectors will be lost.
     """
-    merge = partial(_merge_message_into_nexus_store, structure, nexus_store, data)
+    merge = partial(_merge_message_into_store, store=store, data=data, modules=modules)
     if module_name == "ev44":
-        merge(
-            _find_ev44,
-            _initialize_ev44,
-            _merge_ev44,
+        return merge(
```
jokasimr, line 375: `merge` returns nothing so it seems odd to return its return value

jokasimr (commented):
I have not participated in the discussions about the new interface. But what I understand the new interface is mainly:
1. We don't read static data from the nexus template, instead we read static data from the nexus file.
2. Non-static data is not accumulated in a nexus structure.

From only those two points it seems we could just remove `combine_nexus_store_and_structure` function in `_nexus_helpers.py` and that would be enough.

src/beamlime/applications/_nexus_helpers.py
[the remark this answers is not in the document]
jokasimr: Steps we needed to do before:
1. Identify datasets/groups in the nexus template that are associated with streams
2. Initialize datasets/groups where streamed data is accumulated
3. When new messages arrive, dispatch to the associated merge-function to accumulate the message
4. When complete nexus structure is requested, return nexus structure with stream placeholders replaced by datasets/groups with data from streams.

Only step 4 is removed by the interface change, but most of the logic is associated with step 2 and 3.

YooSunYoung:
Simon and I had an offline discussion and we decided to take back the assumption that `topic name` and `source` can have multiple matches.
So, I'll apply that changes and reopen the PR.

YooSunYoung:
I'll close it as #231 will replace the most important part of this PR.

PR #226, opened 2024-09-18
Result registry to retrieve results in a notebook.

YooSunYoung:
Splitting up #220 

Just like the widget, data reduction handler has a result registry so that user can access to the workflow results easily.

It'll keep updating as it is running.


src/beamlime/applications/handlers.py
```diff
@@ -157,8 +159,11 @@ def from_args(
 class DataReductionHandler(HandlerInterface):
     """Data reduction handler to process the raw data."""
 
-    def __init__(self, workflow: StatelessWorkflow) -> None:
+    def __init__(
+        self, workflow: StatelessWorkflow, reuslt_registry: ResultRegistry | None = None
+    ) -> None:
         self.workflow = workflow
+        self.result_registry = {} if reuslt_registry is None else reuslt_registry
```
MridulS, line 166: Do we need to worry about excessive memory usage if we keep this registry around?

YooSunYoung: Not really, they'll be usually histograms, so a few hundreds of numbers.

MridulS (approved):

PR #228, opened 2024-09-20
Parse run start message earlier to collect module information

YooSunYoung:
We have to match the message by `schema`(`module` type in the template), `topic` and `source_name` (`source` in the json template) so I made the combination of them as a key (`StreamModuleKey`).

Values have handles to the useful objects so that we can initialize/extend dataset easily from the information.

With this change, the merging message no longer have to depend on the whole structure. (This change will be followed)

### DIFF
--- a/tests/applications/streaming_module_test.py
+++ b/tests/applications/streaming_module_test.py
@@ -1,0 +1,115 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+from beamlime.applications._nexus_helpers import (
+    StreamModuleKey,
+    StreamModuleValue,
+    collect_streaming_modules,
+)
+
+
+def test_collect_streaming_modules_event_data(ymir: dict) -> None:
+    modules = collect_streaming_modules(ymir)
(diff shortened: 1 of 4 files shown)


tests/applications/streaming_module_test.py
```diff
@@ -0,0 +1,115 @@
...
+                ],
+                "attributes": [{"name": "NX_class", "values": "NXevent_data"}],
+            },
+        )
+        for i in range(2)
+    }
+    ev44_modules = {
+        key: value for key, value in modules.items() if key.module_type == "ev44"
+    }
+    assert ev44_modules == expected_ev44_modules
```
MridulS, line 42: Should we also test if there are errors here? `collect_streaming_modules` doesn't do any error handling.

YooSunYoung: Sorry for late reply.

I checked the error testing/handling possibility but it seems like there should be proper validation step that confirms all the assumption that ``collect_streaming_modules`.
We should also have an integration test between `json` structure and `beamlime`.
It's a little out of the scope in this PR so I did it in #230 ...!

jokasimr (approved):
Looks good to me!

PR #232, opened 2024-09-26
LIve data reduction workflow protocol

### LINKED ISSUE #131: Define workflow interface
Ideas from preliminary discussion:

```python
class MyWorkflow(beamlime.Workflow):
    def __init__(self, run_start, params: dict[str, ParamValue]):
        self.pipeline = sciline.Pipeline()

    # Later, do not include in v1?
    def get_params(self) -> dict[str, ParamSpec]:
        """Return parameters for the workflow that can be configured by a user."""
        return {'wavelength-binning': ParamSpec(float, 0.0, 10.0, description='Wavelength binning', default=0.1)}

    def __call__(self, merged_chunk) -> sc.DataGroup:
        """
        Return plottable data
        
        For now return merged data. If workflow wants merge, it has to do it itself.
        We can later extract this, since it is somewhat generic (or implement helpers
        in this base class). Options are:
        - Show average over last X seconds.
        - Show total since start.
        """ 
        self.pipeline[RawData[SampleRun]] = merged_chunk
        return self.pipeline.compute(IofQ[SampleRun])
```

 - We might want to make this a `Protocol` so it can be implemented in packages that do not depend on Beamlime.
 - See #134, should this be done before calling the workflow? Should `Workflow` define an interface for customizing this? What if the workflow wants to perform additional setup steps based on static input that it wants to avoid reperforming?
 - Consider the problem of finding the correct merge-mode in #135. Should the workflow interface require from the workflow inplementer to specify this?

YooSunYoung:
Fixes #131 

We can use ``StreamProcessor`` in the live data reduction workflows.

This is not stateless anymore so it needs a constructor(__init__) to initialize the first state and a caller (__call__) to update the new chunk.

https://github.com/scipp/essreduce/blob/aef4a3c1094dd15f1849aedc74e9960e1b922de3/src/ess/reduce/streaming.py#L131-L208


src/beamlime/workflow_protocols.py
```diff
@@ -67,4 +67,27 @@ def provide_stateless_workflow(workflow: Workflow) -> StatelessWorkflow:
...
+        ...
+
+    def __call__(
+        self, nxevent_data: dict[str, JSONGroup], nxlog: dict[str, JSONGroup]
+    ) -> WorkflowResult:
+        """Call the workflow and return the computed results that can be visualized.
+
+        ``nxevent_data`` should contain empty events even if no events was received
+        since the last call.
+        ``nxlog`` should only contain keys that we received since the last call.
```
SimonHeybrock, line 91: Can we use the standard docstring format? https://scipp.github.io/development/coding-conventions.html#docstrings

YooSunYoung: @SimonHeybrock 
Done...!

SimonHeybrock (approved):

Issue #227, opened 2024-09-19
Integration test between nexus_structure json files and beamlime.


YooSunYoung:
In the beamlime, we collect streaming module configuration and we assume certain fields to exist.

We need to add integration between the latest json files and the beamlime nexus helpers.

SimonHeybrock:
We are not using the JSON files anymore.

PR #234, opened 2024-09-26
More strict nexus structure validation and tests.

YooSunYoung:
Related to #227 

This is for integration tests between json template and beamlime.

### DIFF
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -303,7 +303,6 @@ def _is_monitor(group: NexusGroup) -> bool:
 
 def _initialize_ev44(module_spec: StreamModuleValue) -> NexusGroup:
     parent = module_spec.parent
-    _validate_ev44_module_spec(module_spec)
     group: NexusGroup = cast(NexusGroup, parent.copy())
     group['children'] = [
         create_dataset(
@@ -390,7 +389,6 @@ def _merge_ev44(group: NexusGroup, data: DeserializedMessage) -> None:
 
 def _initialize_f144(module_spec: StreamModuleValue) -> NexusGroup:
     parent = module_spec.parent
-    _validate_f144_module_spec(module_spec)
     group: NexusGroup = cast(NexusGroup, parent.copy())
     group["children"] = [
         create_dataset(
--- a/src/beamlime/applications/_nexus_helpers.py
+++ b/src/beamlime/applications/_nexus_helpers.py
@@ -182,6 +182,48 @@ def _validate_module_keys(
         )
 
 
+def _validate_f144_module_spec(
+    module_spec: StreamModuleValue,
+) -> None:
+    """Validate the f144 module."""
+    if len(module_spec.parent["children"]) != 1:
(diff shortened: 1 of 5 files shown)


tests/applications/ymir_detectors.json
YooSunYoung, line 1: You don't have to pay attention to this file too much since it'll be updated with the latest version of file along with compatible nexus file.

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -182,6 +182,48 @@ def _validate_module_keys(
...
+    module_spec: StreamModuleValue,
+) -> None:
+    """Validate the ev44 module."""
+    if len(module_spec.parent["children"]) != 1:
+        raise InvalidNexusStructureError(
+            "Group containing ev44 module should have exactly one child"
+        )
+
+
+def _validate_ev44_module_values(
```
jokasimr, line 218: Wouldn't it be good to merge these functions into one?

Something like 

```
def _validate_module_values(
    key_value_dict: dict[StreamModuleKey, StreamModuleValue],
) -> None:
    for key, value in key_value_dict.items():
        if key.module_type == "f144":
            _validate_f144_module_spec(value)
        if key.module_type == "ev44":
            _validate_ev44_module_spec(value)
        # etc
```

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -222,7 +264,10 @@ def collect_streaming_modules(
     )
```
jokasimr, line 264: Why first create a tuple of pairs and then make it into a dict?
Why not just make the dict directly?

src/beamlime/applications/_nexus_helpers.py
```diff
@@ -258,8 +303,7 @@ def _is_monitor(group: NexusGroup) -> bool:
 
 def _initialize_ev44(module_spec: StreamModuleValue) -> NexusGroup:
     parent = module_spec.parent
-    if len(parent['children']) != 1:
-        raise ValueError('Group containing ev44 module should have exactly one child')
+    _validate_ev44_module_spec(module_spec)
```
jokasimr, line 306: It seems excessive to validate the spec every time we are using it. Can't we make sure to validate the specs when they are constructed, then we don't have to sprinkle validation code here and there?

jokasimr (commented):
Looks good overall. Left a couple of comments/questions.

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: Yeah... that's true. It's already checking in the `collect_streaming_modules`.

src/beamlime/applications/_nexus_helpers.py
jokasimr: Aha okay 👍

src/beamlime/applications/_nexus_helpers.py
YooSunYoung: @jokasimr  Done...!

jokasimr (approved):

PR #242, opened 2024-10-04
Do not depend on json template for mocking streaming.

YooSunYoung:
@SimonHeybrock  

This update should allow you to run beamlime only with a file without any json structure.

```bash
beamlime \ 
    --nexus-file-path 60392-2022-02-28_2215.nxs \
    --workflow ess-loki-monitor \
    --log-level DEBUG \
    --merge-every-nth 20 \
    --image-path-prefix loki-test \
    --num-frames 40 \
    #--fill-dummy-data  # uncomment this if you want to use actual data from the file.

```

### DIFF
--- a/src/beamlime/applications/handlers.py
+++ b/src/beamlime/applications/handlers.py
@@ -105,12 +105,12 @@ class DataAssembler(HandlerInterface):
         self.debug("Data piece merged for %s", message.content.key)
         if self._should_send_message():
             nxevent_data = {
-                nexus_path_as_string(self.streaming_modules[key].path): JSONGroup(value)
+                nexus_path_as_string(self.streaming_modules[key].path): value
                 for key, value in self._nexus_store.items()
                 if key.module_type == "ev44"
             }
             nxlog = {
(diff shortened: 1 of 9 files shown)


src/beamlime/applications/handlers.py
```diff
@@ -105,7 +105,7 @@ def merge_data_piece(self, message: DataPieceReceived) -> DataReady | None:
         self.debug("Data piece merged for %s", message.content.key)
         if self._should_send_message():
             nxevent_data = {
-                nexus_path_as_string(module_spec.path): JSONGroup(value)
```
YooSunYoung, line 108: There was a bug that I found as I tested it with a file : D....!! wheeee

YooSunYoung:
Feel free to make it better and proceed while I'm on vacation : D ...!

YooSunYoung:
For testing, there are already existing tests that I had updated, but we still need to test ``fill-dummy-data``.
I opened an issue #243 since it was too complicated to do it at once.

SimonHeybrock (approved):
Worked for testing with the new SANS workflows.

PR #248, opened 2024-10-18
Add application for continuous operation and raw detector view displays

### LINKED ISSUE #246: Continuous operation
During a meeting with scientists, it became clear that in addition to the start/stop mode, we will also need a "continuous" operation mode. The main purpose of this is to display a live raw detector view. We are only interested in the `ev44` messages for displaying raw counts on the the detectors.

Notes:
- We will require a template file so we can load and create the detector view, but this can either be static or be determined from an older/fixed run-start message.
- Subsequent run-start and run-stop messages should be ignored, we should just keep passing event data to the data reduction workflow.

I think the simplest and most pragmatic solution is to simple not handle run-start and run-stop at all in this mode. We can just give a path to a NeXus template file by hand. This could even by handled as part of setting up the reduction workflow, i.e., outside Beamlime itself. Mapping from topic/source names to detector names can also be configured by hand.

- [ ] Add new application that can be run as a script, see https://github.com/scipp/beamlime/blob/51853589b5510c70ac52527fddb45a5c1212e88b/pyproject.toml#L41-L42.
- [ ] Instead of https://github.com/scipp/beamlime/blob/51853589b5510c70ac52527fddb45a5c1212e88b/src/beamlime/applications/handlers.py#L74, define a new handler that only requires `DataPieceReceived` (no `RunStart`) @SimonHeybrock

YooSunYoung:
Fixes #246.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dynamic = ["version"]
 
 [project.scripts]
 beamlime = "beamlime.__main__:main"
-show-detector = "beamlime.executables.show_detector:main"
+beamlime-raw-detectors = "beamlime.executables.show_detector:main"
 
 [project.urls]
 "Bug Tracker" = "https://github.com/scipp/beamlime/issues"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ dynamic = ["version"]
 
 [project.scripts]
 beamlime = "beamlime.__main__:main"
+show-detector = "beamlime.executables.show_detector:main"
 
 [project.urls]
 "Bug Tracker" = "https://github.com/scipp/beamlime/issues"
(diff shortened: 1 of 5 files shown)


YooSunYoung:
@SimonHeybrock 
Now the changes are merged here!
Will you let me know when it works from your side...??

SimonHeybrock:
@YooSunYoung How do I have to run this now?

SimonHeybrock (approved):
Works for me.

YooSunYoung:
@SimonHeybrock  I don't have a proper role to merge it bypassing the CI checks for some reason...? 
The test fails since the new implementation in essreduce is not published yet.

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dynamic = ["version"]
 
 [project.scripts]
 beamlime = "beamlime.__main__:main"
+show-detector = "beamlime.executables.show_detector:main"
```
SimonHeybrock, line 43:
```suggestion
beamlime-raw-detectors = "beamlime.executables.show_detector:main"
```
To make it easier to discover, e.g., by tab-completion?

PR #251, opened 2024-10-24
Show detector patch

YooSunYoung:
@SimonHeybrock I added some more arguments for easier command line testings.

As I tried it with a fake listener, I had this error.

I was using this file: /ess/data/coda/2024/268227/raw/268227_00029504.hdf

Do you have any idea why it happens and how to fix it?

```
 File "/Users/sunyoungyoo/ESS/beamlime/src/beamlime/applications/handlers.py", line 144, in handle
    det.add_counts(np.concatenate(buffer))
  File "/Users/sunyoungyoo/ESS/essreduce/src/ess/reduce/live/raw.py", line 161, in add_counts
    counts = self.bincount(data)
  File "/Users/sunyoungyoo/ESS/essreduce/src/ess/reduce/live/raw.py", line 90, in bincount
    out.values = np.bincount(offset, minlength=self._size).reshape(self.data.shape)
ValueError: 'list' argument must have no negative elements
```

However, with a workaround(I just filtered all negative offset in the bincount call) 
``show-detector`` command worked with a fake listener ...!

The command I used is:
```bash
show-detector --fake-listener \
  --static-file-path loki_valid.h5 \
  --nexus-file-path loki_valid.h5 \
  --image-path-prefix test \
  --num-frames 100

```

### DIFF
--- a/src/beamlime/applications/handlers.py
+++ b/src/beamlime/applications/handlers.py
@@ -291,6 +291,8 @@ def random_image_path() -> ImagePath:
 
 MaxPlotColumn = NewType("MaxPlotColumn", int)
 DefaultMaxPlotColumn = MaxPlotColumn(3)
+NormalizeHistogramFlag = NewType("NormalizeHistogramFlag", bool)
+DefaultNormalizeHistogramFlag = NormalizeHistogramFlag(False)
 
 
 class PlotStreamer(HandlerInterface):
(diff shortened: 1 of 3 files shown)


src/beamlime/applications/handlers.py
```diff
@@ -125,25 +136,42 @@ def handle(self, message: DataPieceReceived) -> WorkflowResultUpdate | None:
...
+        if self._pulse % self.n_pulses == 0:
             results = {}
             for name, det in self._detectors.items():
                 buffer = self._buffer[name]
                 if len(buffer):
                     det.add_counts(np.concatenate(buffer))
                     buffer.clear()
                 for window in (1,):
-                    key = f'{name.split("/")[-1]} window={window*npulse}'
+                    key = f'{name.split("/")[-1]} window={window*self.n_pulses}'
```
SimonHeybrock, line 147: Can you undo all changes to `__init__` and `handle`? It will run into big conflicts with what I am doing.

src/beamlime/applications/handlers.py
```diff
@@ -427,9 +457,11 @@ def __init__(
         logger: BeamlimeLogger,
         image_path_prefix: ImagePath,
         max_column: MaxPlotColumn = DefaultMaxPlotColumn,
+        normalize: NormalizeHistogramFlag = DefaultNormalizeHistogramFlag,
```
SimonHeybrock, line 460: Same here, as mentioned yesterday on Slack, please undo all changes to plotting, as I already have a ton of changes here.

PR #252, opened 2024-10-28
Dependencies update.

YooSunYoung:
We had to update `essreduce` to use live data reduction stuff.


SimonHeybrock:
Shoulnd't we put a minimum essreduce into `pyproject.toml`?

SimonHeybrock (approved):

PR #254, opened 2024-10-30
Avoid recreating figure and axes with every update

SimonHeybrock:
This save some time in every iteration, improving the update latency and frequency.

The same was previously implemented in `PlotStreamer` using Plopp, but for now we are avoiding that due to performance (most likely since it uses `pcolormesh` instead of `imshow`).

### DIFF
--- a/src/beamlime/applications/handlers.py
+++ b/src/beamlime/applications/handlers.py
@@ -449,7 +449,7 @@ def _plot_1d(
     **kwargs,
 ):
     x = sc.midpoints(da.coords[da.dims[0]]).values
-    ax.errorbar(x, da.values, yerr=np.sqrt(da.variances), fmt='o', **kwargs)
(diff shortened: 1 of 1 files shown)


src/beamlime/applications/handlers.py
```diff
@@ -530,7 +531,7 @@ def _plot_2d(
         aspect = 'equal'
 
     values = da.values
-    ax.imshow(values, **kwargs, aspect=aspect)
+    image = ax.imshow(values, **kwargs, aspect=aspect)
```
nvaytet, line 534: Are you missing a `origin='lower'` to have the intuitive image layout (instead of the web layout)?

nvaytet: Ah yes! I missed that :-)

nvaytet (approved):
Not sure why the CI is failing, but changes look good.

PR #262, opened 2024-11-11
Move config to beamlime.config.raw_detectors for creating raw views

MridulS:
This PR moves the config of detectors to `beamlime.config.raw_detectors` and cleans up the `handlers.py` a bit. For now these are just python files but we can make them into TOML/YAML/.... in the future. We can also move them to instrument packages and register them via entry-points. I chose `beamlime.config.raw_detectors` to be explicit that this is the configuration used during raw detector view.

### DIFF
--- a/src/beamlime/config/raw_detectors/__init__.py
+++ b/src/beamlime/config/raw_detectors/__init__.py
@@ -1,11 +1,14 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+
 """
-This config is used during Live Data Reduction detector view.
+This config is used to setup live raw detector views.
 Currently the instrument specific config is stored in python files, but
 they can be moved to a separate file format in the future.
 """
 
-from .dream import _dream
-from .loki import _loki
-from .nmx import _nmx
+from .dream import dream_detectors_config
+from .loki import loki_detectors_config
+from .nmx import nmx_detectors_config
 
-__all__ = ["_dream", "_loki", "_nmx"]
+__all__ = [dream_detectors_config, loki_detectors_config, nmx_detectors_config]
--- a/src/beamlime/config/raw_detectors/__init__.py
+++ b/src/beamlime/config/raw_detectors/__init__.py
@@ -1,0 +1,11 @@
+"""
+This config is used during Live Data Reduction detector view.
+Currently the instrument specific config is stored in python files, but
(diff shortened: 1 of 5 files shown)


src/beamlime/config/raw_detectors/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+This config is used during Live Data Reduction detector view.
+Currently the instrument specific config is stored in python files, but
+they can be moved to a separate file format in the future.
+"""
+
+from .dream import _dream
+from .loki import _loki
+from .nmx import _nmx
+
+__all__ = ["_dream", "_loki", "_nmx"]
```
SimonHeybrock, line 11: Should we consider it public now and remove the leading `_`?

src/beamlime/config/raw_detectors/dream.py
```diff
@@ -0,0 +1,49 @@
+# Order in 'resolution' matters so plots have X as horizontal axis and Y as vertical.
```
SimonHeybrock, line 1: License headers missing?

src/beamlime/config/raw_detectors/nmx.py
```diff
@@ -0,0 +1,25 @@
...
+        'Panel 2': {
+            'detector_name': 'detector_panel_2',
+            'gridspec': (0, 2),
+            'projection': raw.LogicalView(),
+        },
+    },
+}
+# These banks currently have out-of-range event_id values, so we ignore them.
+del _nmx['detectors']['Panel 1']
+del _nmx['detectors']['Panel 2']
```
SimonHeybrock, line 25: I think this was fixed now, both in CODA files as well as `ess.reduce.live.raw` now discarding out-of-range IDs. I think it is safe to remove these lines.

src/beamlime/config/raw_detectors/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
+This config is used during Live Data Reduction detector view.
```
SimonHeybrock, line 2: Not data-reduction actually, how about something like
```suggestion
This config is used to setup live raw detector views.
```

src/beamlime/config/raw_detectors/__init__.py
MridulS: I would then argue to make it something like `dream_detectors_config`, so we don't litter around a python object `dream` in the python namespace.

SimonHeybrock (approved):

Issue #250, opened 2024-10-23
Timer daemon.


YooSunYoung:
From the discussion with @SimonHeybrock 

Instead of having a loop or an if statement to check a timer,
we can have a daemon that publishes `Trigger` event-message every `N` seconds.
By subscribing to the message, event handlers can be triggered every `N` seconds so that it doesn't have to check the timing itself.

SimonHeybrock:
Outdated, since refactoring the entire architecture. If anything, this is something `StreamProcessor` needs to do something to address https://github.com/scipp/beamlime/blob/6f431bdc41c8e47379bf0a2f01675ed1d8e6665b/src/beamlime/core/processor.py#L44-L46. But this might not be necessary since it is not clear we will ever run into such situations that are relevant.

PR #265, opened 2024-11-13
Time-based update and window size for raw-detector view

SimonHeybrock:
This switches to a time based solution, which avoids linking the number of detectors to the update rate, and provides a more natural way of configuration.

This also adds config options instead of hard-coding the updates frequencies.

This probably makes #250 superfluous.

### DIFF
--- a/src/beamlime/applications/raw_count_handlers.py
+++ b/src/beamlime/applications/raw_count_handlers.py
@@ -110,6 +110,8 @@ class RawCountHandler(HandlerInterface):
             self.info("Updating views at %s", reference_time)
             while reference_time >= self._next_update:
                 # If there were no pulses for a while we need to skip several updates.
+                # Note that we do not simply set _next_update based on reference_time
+                # to avoid drifts.
                 self._next_update += self._update_every
             results = {}
             for name, det in self._views.items():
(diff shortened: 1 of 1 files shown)


src/beamlime/applications/raw_count_handlers.py
```diff
@@ -47,7 +47,14 @@ class RawCountHandler(HandlerInterface):
...
     """
 
-    def __init__(self, *, logger: BeamlimeLogger, nexus_file: NexusFilePath) -> None:
+    def __init__(
+        self,
+        *,
+        logger: BeamlimeLogger,
+        nexus_file: NexusFilePath,
+        update_every: float,
+        window_length: float,
```
MridulS, line 56: Can we get some docs for this in Parameters?

src/beamlime/applications/raw_count_handlers.py
```diff
@@ -91,20 +101,30 @@ def handle(self, message: DataPieceReceived) -> WorkflowResultUpdate | None:
...
+        reference_time = sc.datetime(int(data_piece.reference_time), unit='ns')
+        if self._next_update is None:
+            self._next_update = reference_time.copy()
+        if reference_time >= self._next_update:
+            self.info("Updating views at %s", reference_time)
+            while reference_time >= self._next_update:
+                # If there were no pulses for a while we need to skip several updates.
+                # Note that we do not simply set _next_update based on reference_time
+                # to avoid drifts.
+                self._next_update += self._update_every
```
MridulS, line 115:
```suggestion
            self._next_update += ((reference_time - self._next_update) // self._update_every + 1) * self._update_every
```
this should avoid the loop?

MridulS (approved):

PR #266, opened 2024-11-13
Remove hidden default values and move them to instrument config files

SimonHeybrock:
Also fix a changed dimension name to match latest ess.reduce.live.raw


src/beamlime/applications/raw_count_handlers.py
```diff
@@ -85,9 +85,9 @@ def __init__(
                 nexus_file,
                 detector_name=detector['detector_name'],
                 window=self._window_length,
-                projection=detector.get('projection', 'xy_plane'),
+                projection=detector['projection'],
```
MridulS, line 88: I asume this means that `projection` is a required value in all configs? Should we somehow mark them in config schema ?

SimonHeybrock: Yes it is required, and we should eventually develop a schema. For now this is going to be refactored a couple more times.

MridulS (approved):

PR #268, opened 2024-11-14
Add a first version of a frontend, serving plot results as a website

SimonHeybrock:
**Note: The ZMQ code is replaced by FastAPI+WebSockets in #272, please ignore that part of code, or review and merge #272 first**

A new script was added, reinstall so it becomes available:
```sh
pip install --no-deps -e .
```
There is a whole bunch of changes here:
- Add a frontend server. This is meant as a demo for testing.
- Extract plotting code to make it reusable in the frontend.
- Serialization of Scipp data for sending it to the frontend.

The backend (connecting to Kafka and doing data processing) needs to be run separately from the frontend webserver.
```sh
beamlime-frontend-matplotlib
```
Run backend:
```sh
filename='<your recent DREAM, Loki, or NMX file>'
beamlime-raw-detectors \
        --nexus-file-path $filename \
        --static-file-path $filename \
        --log-level WARNING \
        --num-frames 100000 \
        --data-feeding-speed 0.01 \
        --fake-listener \
        --update-every 0.1 \
        --window-length 1.0
```
Then navigate to `http://127.0.0.1:5042` unless it is opened automatically.

Next steps will include:
- Add Control Daemon in the backend, taking commands from the frontend via a ZMQ socket.
- Add a different frontend that is more interactive, supporting zooming, etc.
- Probably the ZMQ code needs to be made more resilient.


src/beamlime/plotting/plot_matplotlib.py
```diff
@@ -0,0 +1,181 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
SimonHeybrock, line 1: No changes in this file, just extracted from `raw_count_handlers.py`.

src/beamlime/frontend/zmq_client.py
```diff
@@ -0,0 +1,167 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
nvaytet, line 1: I'm guessing most of the Flask, ZMQ code was written by copilot?

I didn't look so much into the changes, as I understand close to none of it.
However, this is at an experimental stage, I don't see a reason not to merge this.

src/beamlime/frontend/matplotlib_flask.py
```diff
@@ -0,0 +1,112 @@
...
+
+import asyncio
+import os
+import threading
+from collections import deque
+from io import BytesIO
+
+import msgpack
+from flask import Flask, Response, render_template_string, request
+from matplotlib.figure import Figure
```
MridulS, line 15: We need to force matplotlib to pick a non interactive backend with `matplotlib.use('agg')` https://github.com/matplotlib/matplotlib/issues/14304/#issuecomment-545717061

pyproject.toml
```diff
@@ -30,17 +30,20 @@ requires-python = ">=3.10"
...
+  "msgpack",
+  "rich",
+  "zmq",
 ]
 
 dynamic = ["version"]
 
 [project.scripts]
 beamlime = "beamlime.__main__:main"
 beamlime-raw-detectors = "beamlime.executables.show_detector:main"
```
MridulS, line 45: once https://github.com/scipp/beamlime/pull/272 is merged the project script section should be updated.

src/beamlime/frontend/matplotlib_flask.py
SimonHeybrock: Fixed.

YooSunYoung:
I just saw this PR,
so you did try writing async daemon without the message router I guess.
The ``ZMQClient`` could be written simpler using a message router.

But is this PR still valid or are we going to try implementing another solution?

SimonHeybrock:
> I just saw this PR, so you did try writing async daemon without the message router I guess. The `ZMQClient` could be written simpler using a message router.
> 
> But is this PR still valid or are we going to try implementing another solution?

ZMQ was replaced already (see next PR). But even that is just an experiment, and as you saw from the discussion I would like to consider moving away from that approach as well.

PR #272, opened 2024-11-18
Refactor to use FastAPI+WebSocket

SimonHeybrock:


src/beamlime/frontend/matplotlib_flask.py
```diff
@@ -24,26 +24,41 @@
...
-                        data_buffer.append(arrays)
-                    except (msgpack.UnpackException, KeyError) as e:
-                        app.logger.error("Error processing data: %s", e)
-                await asyncio.sleep(0.1)
-
-    asyncio.run(run_client())
+                try:
+                    data = await websocket.recv()
+                    if data:
+                        print('new data')
```
MridulS, line 37: Should this be a log?

src/beamlime/frontend/matplotlib_flask.py
```diff
@@ -93,6 +108,7 @@ def plot_png():
         ax = fig.subplots()
         ax.text(0.5, 0.5, 'Waiting for data...', ha='center', va='center')
     else:
+        print('plotting')
```
MridulS, line 111: Log?

src/beamlime/executables/raw_detectors_fastapi.py
```diff
@@ -0,0 +1,58 @@
...
+factory = make_factory()
+manager = factory[WebSocketManager]
+
+
+@app.websocket("/ws")
+async def websocket_endpoint(websocket: WebSocket):
+    await manager.connect(websocket)
+    try:
+        while True:
+            _ = await websocket.receive_text()
```
MridulS, line 25: What is happening here? What data is the websocket trying to recieve?

MridulS:
Note: I can run this locally on my machine once I set matplotlib to use the non interactive backend.

src/beamlime/executables/raw_detectors_fastapi.py
SimonHeybrock: I think the whole bit is unused now, I was just doing some initial testing. I'll see if it can be removed.

PR #279, opened 2024-12-16
Clean V1

YooSunYoung:
I left the old version here: https://github.com/scipp/beamlime/releases/tag/legacy


SimonHeybrock:
We can leave them for now, until we know where they will be needed, I don't think they will cause conflicts with the new content?

Is the docs build expected to fail?

YooSunYoung:
> Is the docs build expected to fail?

No, I just pushed a fix. It should pass now.

SimonHeybrock (approved):
Thanks!

PR #285, opened 2025-01-09
Improve setup to facilitate integration testing and future service setup

SimonHeybrock:
Not much of a test yet, but I hope this provides the foundation.

### DIFF
--- a/src/beamlime/core/config_subscriber.py
+++ b/src/beamlime/core/config_subscriber.py
@@ -5,16 +5,15 @@ import logging
 import threading
 from typing import Any
 
-from confluent_kafka import Consumer
-
 from ..kafka.message_adapter import KafkaMessage
+from ..kafka.source import KafkaConsumer
 
 
 class ConfigSubscriber:
     def __init__(
         self,
         *,
-        consumer: Consumer,
+        consumer: KafkaConsumer,
         config: dict[str, Any] | None = None,
         logger: logging.Logger | None = None,
     ):
@@ -50,8 +49,9 @@ class ConfigSubscriber:
     def _run_loop(self):
         try:
             while self._running:
-                msg = self._consumer.poll(0.1)
-                self._process_message(msg)
+                messages = self._consumer.consume(num_messages=100, timeout=0.1)
+                for msg in messages:
+                    self._process_message(msg)
         except RuntimeError as e:
             self._logger.exception("Error ConfigSubscriber loop failed: %s", e)
             self.stop()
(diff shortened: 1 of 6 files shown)


src/beamlime/core/config_subscriber.py
```diff
@@ -32,22 +33,25 @@ def start(self):
...
+            return
+        if message.error():
+            self._logger.error('Consumer error: %s', message.error())
+            return
+        key = message.key().decode('utf-8')
+        value = json.loads(message.value().decode('utf-8'))
+        self._logger.info(
+            'Updating config: %s = %s at %s', key, value, message.timestamp()
+        )
+        self._config[key] = value
```
YooSunYoung, line 47: I think we should be careful of using `key` in this case.
`key` is for associating the messages with a certain partition, so it's mainly for the broker not the consumer.
https://www.confluent.io/learn/kafka-message-key/

src/beamlime/core/config_subscriber.py
```diff
@@ -32,22 +33,25 @@ def start(self):
...
-                    self._logger.error('Consumer error: %s', msg.error())
-                    continue
-
-                key = msg.key().decode('utf-8')
-                value = json.loads(msg.value().decode('utf-8'))
-                self._logger.info(
-                    'Updating config: %s = %s at %s', key, value, msg.timestamp()
-                )
-                self._config[key] = value
+                messages = self._consumer.consume(num_messages=100, timeout=0.1)
```
YooSunYoung, line 52: Why does the config consumer need to consume so many messages at once?

src/beamlime/core/config_subscriber.py
SimonHeybrock: Moved to new issue #286, since this is unrelated to this change (just moved code).

src/beamlime/core/config_subscriber.py
SimonHeybrock: To enable better batching and minimize the number of back and forth with the broker? True, this usually does not happen as few values are updated, but it does when loading initial config or catching up after restart.

Do you think it is harmful?

tests/services/data_service_test.py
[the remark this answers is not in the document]
SimonHeybrock: I don't know?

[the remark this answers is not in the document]
YooSunYoung: We can just have it in mind as an alternative to write a fake object in case it becomes too complicated later.

YooSunYoung (approved):

tests/services/data_service_test.py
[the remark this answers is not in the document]
SimonHeybrock: Yes, feel free to refactor when writing more tests, in case a mocking approach is more convenient.

PR #294, opened 2025-01-20
Add detector plots in dashboard

### LINKED ISSUE #293: Update `dashboard.py` to also make 2-D detector plots.

SimonHeybrock:
For now, this is *not* using any of the plot settings in the instrument config files, nor is it trying to look pretty.

Fixes #293.


src/beamlime/services/dashboard.py
```diff
@@ -201,17 +194,49 @@ def create_monitor_plot(key: str, data: sc.DataArray) -> go.Figure:
...
+        fig = go.Figure()
+        x_dim, y_dim = data.dims
+        fig.add_heatmap(
+            z=[[]],
+            x=[],  # Will be filled with coordinate values
+            y=[],  # Will be filled with coordinate values
+            colorscale='Viridis',
+        )
+
+        def maybe_unit(dim: str) -> str:
```
nvaytet, line 211: Use the `maybe_unit` in the monitor plot above as well?

src/beamlime/services/dashboard.py
```diff
@@ -201,17 +194,49 @@ def create_monitor_plot(key: str, data: sc.DataArray) -> go.Figure:
...
+            num = len(messages)
+            messages = compact_messages(messages)
             self._logger.info(
-                "Got %d messages, showing most recent %d", num, len(monitor_messages)
+                "Got %d messages, showing most recent %d", num, len(messages)
             )
+            monitor_messages = [
+                msg for msg in messages if 'beam_monitor' in msg.key.topic
+            ]
+            detector_messages = [msg for msg in messages if 'detector' in msg.key.topic]
```
nvaytet, line 239: Can the looping over messages become expensive at any point?
I.e. is it worth refactoring to have a single `for msg in messages` and splitting between monitor and detector inside the loop?

src/beamlime/services/dashboard.py
```diff
@@ -222,6 +247,24 @@ def update_plots(self, n: int | None):
...
+                    if dim not in data.coords:
+                        data.coords[dim] = sc.arange(dim, data.sizes[dim], unit=None)
+                if key not in self._detector_plots:
+                    self._detector_plots[key] = self.create_detector_plot(key, data)
+                fig = self._detector_plots[key]
+                if len(data.dims) == 1:
+                    fig.data[0].x = data.coords[data.dim].values
+                    fig.data[0].y = data.values
+                else:  # 2D
+                    x_dim, y_dim = data.dims
```
nvaytet, line 263: In plopp, the order is usually `y_dim, x_dim`. I don't know if it matters here though...

src/beamlime/services/dashboard.py
SimonHeybrock: There are typically O(10) messages.

src/beamlime/services/dashboard.py
SimonHeybrock: Good catch, was indeed wrong!

src/beamlime/services/dashboard.py
SimonHeybrock: Monitors always have a time axis with a unit.

nvaytet (approved):

PR #295, opened 2025-01-21
Add stripped geometry files in a Pooch

SimonHeybrock:
This allows for running the `detector_data` service without having to provide a local file.

As explained in the updated docstring, a more dynamic mechanism will likely need to be implemented in the future. We could even imagine having a service that can provide the path via Kafka.

### DIFF
--- a/scripts/setup-kafka-topics.sh
+++ b/scripts/setup-kafka-topics.sh
@@ -42,17 +42,7 @@ kafka-topics --create --bootstrap-server kafka:29092 \
   --config retention.bytes=-1
 
 kafka-topics --create --bootstrap-server kafka:29092 \
-  --topic ${BEAMLIME_INSTRUMENT}_beam_monitor_cumulative \
-  --config cleanup.policy=delete \
-  --config delete.retention.ms=60000 \
-  --config max.message.bytes=327680 \
-  --config retention.bytes=1073741824 \
-  --config retention.ms=30000 \
-  --config segment.bytes=10485760 \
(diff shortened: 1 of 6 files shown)


scripts/setup-kafka-topics.sh
SimonHeybrock, line 1: Changes in this file unrelated, this was missed in #292.

src/beamlime/handlers/detector_data_handler.py
```diff
@@ -126,3 +129,68 @@ def get(self) -> sc.DataArray:
...
         self._det.clear_counts()
+
+
+# Note: Currently no need for a geometry file for NMX since the view is purely logical.
+# DetectorHandlerFactory will fall back to use the detector_number configured in the
+# detector view config.
+_registry = {
+    'geometry-dream-2025-01-01.nxs': 'md5:91aceb884943c76c0c21400ee74ad9b6',
+    'geometry-loki-2025-01-01.nxs': 'md5:e2e48c30ad02dcfa940b7c26885216f2',
+}
```
YooSunYoung, line 140: Why do we have the date in the file name instead of using the `version` ...?

SimonHeybrock: Good question! My though was that we actually need multiple at the same time, depending on what date we use for runs. Will not happen in production but in dev/testing. In other words: The old files never become invalid. Instead they will always remain valid for a certain date range. Thus the pooch version mechanism is not what we want here.

PR #298, opened 2025-01-24
Handle multiple messages at a time

SimonHeybrock:
The reason for this change of mechanism is to allow for calling expensive preprocessing and accumulation more eagerly. We cannot afford to do this with every message. The previous approach of delaying until an update is produced had the downsides of introducing (1) additional latency and (2) spending more idle time waiting for messages and thus reducing the maximum achievable throughput.

### DIFF
--- a/src/beamlime/core/handler.py
+++ b/src/beamlime/core/handler.py
@@ -182,6 +182,15 @@ class PeriodicAccumulatingHandler(Handler[T, U]):
         )
 
     def handle(self, message: Message[T]) -> list[Message[V]]:
+        self._preprocess(message)
+        if message.timestamp < self._next_update:
+            return []
+        data = self._preprocessor.get()
+        for accumulator in self._accumulators.values():
+            accumulator.add(timestamp=message.timestamp, data=data)
+        return self._produce_update(message.key, message.timestamp)
+
+    def _preprocess(self, message: Message[T]) -> None:
         if self.start_time() > self._last_clear:
             self._preprocessor.clear()
             for accumulator in self._accumulators.values():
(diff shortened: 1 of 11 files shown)


src/beamlime/core/handler.py
```diff
@@ -181,7 +179,35 @@ def __init__(
...
+        # delay will lead to more messages to be consumed in the next iteration, driving
+        # down the number of calls to this method until equilibrium is reached. This can
+        # be demonstrated using a simple model process with an accumulate call that has
+        # a constant + linear-per-message cost.
+        data = self._preprocessor.get()
+        for accumulator in self._accumulators.values():
+            accumulator.add(timestamp=messages[-1].timestamp, data=data)
+        if messages[-1].timestamp < self._next_update:
+            return []
+        return self._produce_update(messages[-1].key, messages[-1].timestamp)
```
jokasimr, line 208: In the old version this function took a `message` as argument, now it takes a message key and a message timestamp. Why the change?

jokasimr: What is the timestamp of a message conceptually? I saw the timestamp from the latest message in the collection is propagated to the output message, is the timestamp the latest point in time the message contains data from?

jokasimr (approved):
Looks good. I got some questions to make sure I understood it correctly.

src/beamlime/core/handler.py
```diff
@@ -181,7 +179,35 @@ def __init__(
             convert=lambda raw: int(sc.scalar(**raw).to(unit='ns', copy=False).value),
         )
 
-    def handle(self, message: Message[T]) -> list[Message[V]]:
+    def handle(self, messages: list[Message[T]]) -> list[Message[V]]:
+        # Note an issue here: We preprocess all messages, with a range of timestamps.
+        # If one of the accumulators is a sliding-window accumulator, the cutoff time
```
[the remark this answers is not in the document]
SimonHeybrock: There are preprocessors that merge messages. For example, we have one that computes a time-of-arrival histogram from all messages. It is not clear if this is a good approach, but pushing more work into the accumulators is also problematic.

src/beamlime/core/handler.py
SimonHeybrock: For input data it is the time of the pulse that came before the neutrons in the message. For the output it is as you say. We need need a more precise mechanism in the future.

PR #303, opened 2025-01-31
Add ROI histograms

### LINKED ISSUE #300: ROI selection service
Users need to select a single pixel or regions on a detector (or projection thereof) and obtain a result for this region. The result could be:
- Time-of-arrival histogram (simple case).
- Reduced data, such as I(Q) for the detector region (more complex case).

There are roughly two options:
1. Include the ROI selection in multiple places, such as the data-reduction service and in a TOF histogram service.
2. Create a dedicated ROI selection service that can restream the selected subset of data.

The latter is way more flexible, since it can allow for, e.g., having two data reductions done, one for the full data, one for the ROI (or even support multiple ROIs), etc. The downside is that for ROIs containing a substantial fraction of the total counts the restreamed data volume would approach the raw data volume.

On the other hand, data reduction groups incoming data by pixel id anyway, making ROI selection itself very easy and efficient. It therefore seems wasteful to not use this. Should/could the grouping be performed separately, independently of the data reduction and then restreamed? Or should TOF histograms be produced as a "side effect" of data reduction? The latter would put limits on how quickly they could update.

I am not sure we have enough data to make a final decision on this right now.

In terms of implementation, a quick benchmark indicates that `numpy.isin` with boolean indexing is generally fast enough (unless we reach very high even rates). Break-even with `scipp.group` with boolean indexing is reached only at higher event numbers, i.e., unless we reuse the grouping for data reduction a NumPy-based approach seems like a good choice. *Update: After scipp/scipp#3643 is released, `scipp.group` should be more suitable.*

SimonHeybrock:
This change is mainly about the `detector_data` service, even though I have made a small dashboard addition so it can be tested/demonstrated. This is not meant to be usable, in particular there will/should be per-view controls. The config mechanism will also need updating for that.

Depends on a release of scipp/essreduce#173.

image

Fixes #300 (in first approximation).


src/beamlime/handlers/accumulators.py
```diff
@@ -117,34 +130,118 @@ def _cleanup(self) -> None:
...
+            self.clear()  # Clear accumulated data when ROI changes
+
+    def _add_weights(self, data: sc.DataArray) -> None:
+        constituents = data.bins.constituents
+        content = constituents['data']
+        content.coords['time_of_arrival'] = content.data
+        content.data = sc.ones(
+            dims=content.dims, shape=content.shape, dtype='float32', unit='counts'
+        )
+        data.data = sc.bins(**constituents)
```
nvaytet, line 228: Can the `_bins_no_validate` be used here for extra performance gain?

SimonHeybrock: Yes, thanks, fixed now!

nvaytet (approved):

PR #304, opened 2025-02-03
Add support for selecting a TOA range in detector plots

### LINKED ISSUE #305: Add TOA-range selector to dashboard and use it in `accumulators.PixelIDMerger`.

SimonHeybrock:
Fixes #305.


src/beamlime/handlers/detector_data_handler.py
```diff
@@ -135,9 +136,29 @@ class DetectorCounts(Accumulator[sc.DataArray, sc.DataArray]):
...
+        return (
+            sc.scalar(value['low'], unit=value['unit']).to(unit='ns'),
+            sc.scalar(value['high'], unit=value['unit']).to(unit='ns'),
+        )
+
+    def apply_toa_range(self, data: sc.DataArray) -> sc.DataArray:
+        if (toa_range := self._toa_range()) is None:
+            return data
+        low, high = toa_range
+        return data.bins.assign_coords(toa=data.bins.data).bins['toa', low:high]
```
jokasimr, line 157: I need some explanation here :smile:

* What is `data`? (What shape shape and what fields is it expected to have?)
* Why do we need to assign coords?
* Why is `toa=data.bins.data`?

src/beamlime/services/dashboard.py
```diff
@@ -157,6 +157,23 @@ def _setup_layout(self) -> None:
                 value=[45, 55],
                 marks={i: str(i) for i in range(0, 101, 20)},
             ),
+            html.Label('Time-of-arrival range (us)'),
+            dcc.Checklist(
+                id='toa-checkbox',
+                options=[
+                    {'label': 'Filter by time-of-arrival [μs]', 'value': 'enabled'}
+                ],
+                value=[],
```
jokasimr, line 166: (minor) checklist is slightly odd for what seems to be a "toggle" on/off, maybe consider https://dash.plotly.com/dash-daq/booleanswitch

jokasimr (approved):

src/beamlime/handlers/detector_data_handler.py
```diff
@@ -135,9 +136,29 @@ class DetectorCounts(Accumulator[sc.DataArray, sc.DataArray]):
...
+        )
+        self._current_toa_range = None
+
+    def _convert_toa_range(self, value: dict[str, Any] | None) -> None:
+        self.clear()
+        if value is None:
+            return None
+        return (
+            sc.scalar(value['low'], unit=value['unit']).to(unit='ns'),
+            sc.scalar(value['high'], unit=value['unit']).to(unit='ns'),
```
nvaytet, line 150: Will the unit always be 'ns', or should we instead just store floats and use the units of `data.bins.coords['toa']` in `apply_toa_range` below?

src/beamlime/services/dashboard.py
```diff
@@ -157,6 +157,23 @@ def _setup_layout(self) -> None:
                 value=[45, 55],
                 marks={i: str(i) for i in range(0, 101, 20)},
             ),
+            html.Label('Time-of-arrival range (us)'),
+            dcc.Checklist(
+                id='toa-checkbox',
+                options=[
+                    {'label': 'Filter by time-of-arrival [μs]', 'value': 'enabled'}
```
nvaytet, line 164: It seems it's microseconds here and nanoseconds above?

src/beamlime/handlers/detector_data_handler.py
SimonHeybrock: The bounds come via a message from Kafka. It is essential to send to unit with this (or you will waste hours debugging). Not sure what is the benefit from your suggestion, recreating the `low` and `high` variables in every single call below? It is just making it a bit slower?

src/beamlime/services/dashboard.py
SimonHeybrock: Where?

src/beamlime/services/dashboard.py
SimonHeybrock: That appears to be from a different library (`dash_daq`)?

src/beamlime/services/dashboard.py
nvaytet: The text in the `html.Label` is `'Time-of-arrival range (us)'`, and the unit label in the square brackets is `Filter by time-of-arrival [μs]`?

Unless this is not the same time as in the `_convert_toa_range` function?

SimonHeybrock: Not sure I understand, `us` is just another spelling of microsecond?

nvaytet: What I meant was that on L149-150 in `detector_data_handler.py` you have converted them to nanoseconds?
By 'above' I meant further up the page, not in the same file, sorry.

SimonHeybrock: What is wrong with converting microseconds to nanoseconds? The dashboard uses microseconds, since nanoseconds are a bit unwieldy for a user to think about.

PR #309, opened 2025-02-10
Support weighting plot pixels by number of contributing detector pixels

SimonHeybrock:
Requires scipp/essreduce#182. Weighting can be essential to remove detector irregularities as well as Moire effects. There is a checkbox in the dashboard to toggle this.

Example with uniform noise across all pixels. No weights:
image

Weights:
image


src/beamlime/config/raw_detectors/dream.py
```diff
@@ -2,41 +2,40 @@
...
 
 import scipp as sc
 from ess.reduce.live import raw
 
-_res_scale = 12
+_res_scale = 8
+pixel_noise = sc.scalar(4.0, unit='mm')
 
 dream_detectors_config = {
-    'dashboard': {'nrow': 3, 'ncol': 2},
```
SimonHeybrock, line 14: This and `gridspec` below were used in the prototype last autumn but are no longer in use.

src/beamlime/config/raw_detectors/dream.py
```diff
@@ -53,9 +52,8 @@
                 },
                 transpose=('wire', 'module', 'segment', 'counter', 'strip'),
                 select={'wire': 0},
-                flatten={'z_id': ('module', 'segment', 'counter')},
+                flatten={'mod/seg/cntr': ('module', 'segment', 'counter')},
```
SimonHeybrock, line 55: Unrelated: `z` was misleading, since it was "up", whereas `z` in the instrument coordinates is actually along the beam (`y` would be "up").

MridulS (approved):

PR #318, opened 2025-02-14
Document adding new instruments

SimonHeybrock:
@MridulS Do you want to try this for Bifrost so we can see if the docs are complete?

### DIFF
--- a/docs/user-guide/adding-new-instruments.md
+++ b/docs/user-guide/adding-new-instruments.md
@@ -1,0 +1,101 @@
+# Adding New Instruments
+
+This guide explains how to add support for a new instrument in Beamlime.
+
+## Required Steps
+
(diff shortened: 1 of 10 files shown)


MridulS (approved):
Was able to follow along the docs just fine :)

SimonHeybrock:
> Was able to follow along the docs just fine :)

Did you manage to get everything to run, including the dashboard with updating plots?

MridulS:
Yeah I was able to run with the dashboard now, the dashboard doesn't look that great as there are a lots of detectors.

SimonHeybrock:
> Yeah I was able to run with the dashboard now, the dashboard doesn't look that great as there are a lots of detectors.

Yes, maybe we have to look into merging messages so we can make a single plot from multiple detectors.

PR #319, opened 2025-02-14
Add tests for detector_data and monitor_data service equivalents

SimonHeybrock:
While this is not actually running the service scripts, the setup logic is (I think) as close as possible without actually using Kafka.

~Passing tests will require a release of scipp/essreduce#190.~ ... or so I thought.

### DIFF
--- a/src/beamlime/config/raw_detectors/bifrost.py
+++ b/src/beamlime/config/raw_detectors/bifrost.py
@@ -1,35 +1,24 @@
-import numpy as np
 import scipp as sc
 from ess.reduce.live import raw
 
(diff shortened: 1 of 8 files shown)


MridulS (approved):

SimonHeybrock:
@MridulS I had to fix something in the Bifrost detector number setup, I *think* it was not entirely correct. Could you please have a close look at the last commit.

MridulS:
> @MridulS I had to fix something in the Bifrost detector number setup, I think it was not entirely correct. Could you please have a close look at the last commit.

Yeah the shape of the inner Variable was wrong, it was (100, 3) instead of (3, 100).

PR #320, opened 2025-02-14
 Add BIFROST detector configs 

MridulS:
Ignore the older commits


src/beamlime/config/raw_detectors/bifrost.py
```diff
@@ -0,0 +1,279 @@
...
+pixel_noise = sc.scalar(4.0, unit='mm')
+
+detectors_config = {
+    'detectors': {
+        'Channel (1, 1)': {
+            'detector_name': '123_channel_1_1_triplet',
+            'resolution': {'y': 30 * _res_scale, 'x': 20 * _res_scale},
+            'projection': 'xy_plane',
+            'pixel_noise': pixel_noise,
+        },
```
MridulS, line 13: Is there no need for the projection? I thought I need to use this to apply the transformations and offsets in the nexus file. The LogicalView doesn't run through that information right?

src/beamlime/services/fake_detectors.py
```diff
@@ -50,6 +50,53 @@
...
         f'detector_panel_{i}': (i * 1280**2 + 1, (i + 1) * 1280**2) for i in range(3)
     },
+    'bifrost': {
+        '123_channel_1_1_triplet': (1, 300),
+        '127_channel_1_2_triplet': (301, 600),
+        '131_channel_1_3_triplet': (601, 900),
+        '135_channel_1_4_triplet': (901, 1200),
+        '139_channel_1_5_triplet': (1201, 1500),
+        '144_channel_2_1_triplet': (1501, 1800),
+        '148_channel_2_2_triplet': (1801, 2100),
```
SimonHeybrock, line 60: Seems like there is a pattern, can we avoid spelling it out and use a loop or comprehension, to facilitate future changes?

PR #327, opened 2025-02-24
Bifrost: Merge detectors for raw-detector view and spectrum view

SimonHeybrock:
Bifrost has 45 detectors in stream&files, three He3 tubes each. This is not useful for visualization and data reduction. Therefore, this PR adds a mechanism to merge messages from various sources. Then, we add the following:

- 2-D logical view of the raw detectors.
- "Data reduction", producing a spectrum view (pixel vs. toa), with a reduced pixel number.


src/beamlime/handlers/detector_data_handler.py
```diff
@@ -232,7 +233,7 @@ def _cleanup(self) -> None:
 _registry = {
     'geometry-dream-2025-01-01.nxs': 'md5:91aceb884943c76c0c21400ee74ad9b6',
     'geometry-loki-2025-01-01.nxs': 'md5:8d0e103276934a20ba26bb525e53924a',
-    'geometry-bifrost-2025-02-14.nxs': 'md5:b581f1560aefd4bec56c86395f553516',
+    'geometry-bifrost-2025-01-01.nxs': 'md5:ae3caa99dd56de9495b9321eea4e4fef',
```
SimonHeybrock, line 236: Couldn't get the previous file to work, something about missing `depends_on`. I think just replacing it is fine?

src/beamlime/services/dashboard.py
```diff
@@ -379,14 +380,31 @@ def maybe_unit(dim: str) -> str:
...
+            ratio = long / short
+            max_size = 900
+            if ratio > 3:
+                if y_size < x_size:
+                    fig.update_layout(width=max_size, height=max_size // 3, **opts)
+                else:
+                    fig.update_layout(width=max_size // 3, height=max_size, **opts)
+            else:
+                scale = max_size / long
+                fig.update_layout(width=x_size * scale, height=y_size * scale, **opts)
```
SimonHeybrock, line 407: All the plot sizing is a hack, and I am sure this made it less readable in other cases. Someone needs a better idea, sometime.

pyproject.toml
```diff
@@ -45,6 +45,7 @@ dynamic = ["version"]
 [project.optional-dependencies]
 test = [
     "pytest",
+    "essspectroscopy @ git+https://github.com/scipp/essspectroscopy.git@main",
```
SimonHeybrock, line 48: Not sure it should be added to the main deps. Only needed when running the Bifrost detector & reduction services.

MridulS: We should probably have instrument specific `optional-dependencies` extras.

SimonHeybrock: Added.

src/beamlime/services/data_reduction.py
MridulS, line 1: Is this missing the accompanying `data_reduction.yaml` file? It doesn't know which topic to look at?

MridulS (approved):

PR #332, opened 2025-02-26
Use ':' to separate device name and signal name

SimonHeybrock:
See https://gitlab.esss.lu.se/ecdc/interface-specs/-/blob/main/nicos-beamlime.md?ref_type=heads


src/beamlime/kafka/helpers.py
```diff
@@ -32,3 +32,15 @@ def detector_topic(instrument: str) -> str:
...
+def source_name_separator() -> str:
+    """
+    Return the separator between device name and signal name.
+
+    This is used to construct the source name from the device name and signal name
+    The source_name is used in various Kafka messages.
+
+    ':' is used as the separator in the ECDC naming convention at ESS.
+    """
+    return ':'
```
MridulS, line 46: Do we need to have a separate function here?

SimonHeybrock: You mean, we should just hard-code it where used?

MridulS: Yeah, seems like an overkill and too much indirection unless this needs to be controlled or changed in the future a lot.

SimonHeybrock: But it avoids 5 hours debugging because a wrong separator made it into some random file?

SimonHeybrock: But maybe it would be better to have a helper function building the source name, instead of one for the separator?

MridulS: > But maybe it would be better to have a helper function building the source name, instead of one for the separator?

Yeah I do like that better, but anyway we can iterate on that later.

PR #338, opened 2025-03-03
Add pytest-benchmark to test deps

MridulS:
The nightly test is failing as there is `--benchmark-skip` in the pytest config but pytest-benchmark needs to be installed to use that flag.


SimonHeybrock (commented):
It is already here: https://github.com/scipp/beamlime/blob/33bf26968543f966f60a83e3e9eb79807e949022/requirements/basetest.in#L7

Not sure why that is not working on nightly. In any case, should be removed there is it is now pyproject.toml?

MridulS:
> Not sure why that is not working on nightly

Nightly deps are read from pyproject.toml while constructing nightly.in, so we should be careful while manually adding deps in requirements/*

SimonHeybrock (approved):

PR #340, opened 2025-03-04
Turn off conda builds and releases

MridulS:


conda/meta.yaml
```diff
@@ -23,7 +23,7 @@ requirements:
 
   {# Conda does not allow spaces between package name and version, so remove them #}
   {% for package in dependencies %}
-    - {% if package == "graphviz" %}python-graphviz{% else %}{{ package|replace(" ", "") }}{% endif %}
+    - {% if package == "confluent_kafka" %}python-confluent-kafka{% else %}{{ package|replace(" ", "") }}{% endif %}
```
YooSunYoung, line 26: If multiple packages are causing problem, should we go back to using a registry instead...?

YooSunYoung (approved):

PR #344, opened 2025-03-07
Add message adapters for f144 "log data"

SimonHeybrock:
This will be needed for data reduction, e.g., detector angles and sample angles.


SimonHeybrock:
Will be replaced by a larger PR.

Issue #345, opened 2025-03-07
Access logdata attributes in HDF5 template


SimonHeybrock:
Information in logdata attributes can be found in the JSON included in the run-start messages. We will however want to process logs even when no run was started. It was suggested that we should read this information from the templates instead.

JSON templates as tracked at https://gitlab.esss.lu.se/ecdc/nexus-json-templates, but we should probably work with the HDF5 versions at `/ess/nexus/{inst}/{inst}.hdf`.

YooSunYoung:
It looks like `topic` and `source_name` are auto-completed like: https://github.com/scipp/beamlime/blob/d744cad888c22d082c0a474c075dca3cf3fa2c10/src/beamlime/services/timeseries.py#L100-L115

It's more about shape and units, no...?

SimonHeybrock:
Johannes may be right that the HDF5 does not have enough info. Will need to have a closer look when we get around to implementing this.

SimonHeybrock:
Brief note: We currently hard-code these attributes in Beamlime (we only have a single case for testing). Depending on how wide-spread the use of logs in data reduction will become, we can stick with that "solution" or more to a more sophisticated approach.

SimonHeybrock:
> Johannes may be right that the HDF5 does not have enough info. Will need to have a closer look when we get around to implementing this.

See #388, topic and source are available as attributes.

SimonHeybrock:
See #936.

PR #346, opened 2025-03-10
Fake service for logdata; timeseries handler

SimonHeybrock:
Logdata will be needed for various data reduction. This would typically be fed into a `StreamProcessor` as *context* (see additions in scipp/essreduce#210), but that will be done in follow-up work.

Here, we add:
- Message adapters for f144 logdata messages.
- Fake service for log data. This will be extended in the future to make multiple logs.
- Handler with accumulators for logdata. This would be used for live data reduction.
- Instead of a more complicated live reduction, a simple service for publishing timeseries of logdata.


src/beamlime/services/timeseries.py
```diff
@@ -0,0 +1,106 @@
...
+
+def make_timeseries_service_builder(
+    *, instrument: str, log_level: int = logging.INFO
+) -> DataServiceBuilder:
+    adapter = ChainedAdapter(first=KafkaToF144Adapter(), second=F144ToLogDataAdapter())
+    handler_factory_cls = partial(
+        LogdataHandlerFactory,
+        instrument=instrument,
+        # This is hard-coded for now. In practice this needs to be read from a template
+        # file.
```
SimonHeybrock, line 45: See #345.

src/beamlime/handlers/timeseries_handler.py
```diff
@@ -0,0 +1,88 @@
...
+    DataArray.
+    """
+
+    def __init__(
+        self,
+        *,
+        instrument: str,
+        logger: logging.Logger | None = None,
+        config: Config,
+        attribute_registry: dict[str, dict[str, any]],
```
YooSunYoung, line 35: I'm a bit concerned that plain dictionaries are used in many places.
If it doesn't have to accept arbitrary fields, I think it'd better be a dataclass or pydantic models.

src/beamlime/handlers/to_nx_log.py
```diff
@@ -0,0 +1,87 @@
...
+from beamlime.handlers.accumulators import LogData
+
+
+class ToNXlog(Accumulator[LogData, sc.DataArray]):
+    """
+    Preprocessor for log data.
+
+    Accumulates LogData objects and returns a single DataArray as it would be read from
+    an NXlog in a NeXus file. The DataArray grows as data is added and is not cleared
+    until explicitly requested.
```
YooSunYoung, line 17: Typically in which cases and by whome will it be clear...?

src/beamlime/handlers/to_nx_log.py
```diff
@@ -0,0 +1,87 @@
...
+                [self._timeseries, self._timeseries], dim='time'
+            )
+
+    def add(self, timestamp: int, data: LogData) -> None:
+        self._ensure_capacity(data.value)
+        self._timeseries.coords['time'].values[self._end] = (
+            self._start_value + data.time
+        )
+        self._timeseries.data.values[self._end] = data.value
+        self._end += 1
```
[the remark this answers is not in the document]
SimonHeybrock: Not sure what we would do this those in live data reduction? Which part exactly in the code are you worried about?

src/beamlime/handlers/timeseries_handler.py
SimonHeybrock: At least the outer one should actually be a `Mapping`, I think. My thought was that attributes might be read from Kafka as well, similar to the config. Are you asking about the inner one?

src/beamlime/handlers/timeseries_handler.py
```diff
@@ -0,0 +1,88 @@
...
+        self._instrument = instrument
+        self._attribute_registry = attribute_registry
+
+    def make_handler(self, key: MessageKey) -> Handler[LogData, sc.DataArray] | None:
+        source_name = key.source_name
+        attrs = self._attribute_registry.get(source_name)
+        if attrs is None:
+            self._logger.warning(
+                "No attributes found for source name '%s'. Messages will be dropped.",
+                source_name,
```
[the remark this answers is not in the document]
SimonHeybrock: You mean we should not display this as warnings but hide it in debug, since it is quite common?

src/beamlime/handlers/to_nx_log.py
SimonHeybrock: The main handler implementation we are using (`PeriodicAccumulatingHandler`) will clear accumulators when the user presses a button.

src/beamlime/handlers/timeseries_handler.py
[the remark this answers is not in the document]
YooSunYoung: Yes, exactly.

src/beamlime/handlers/to_nx_log.py
YooSunYoung: Ah so by the handlers. Okay.

src/beamlime/handlers/timeseries_handler.py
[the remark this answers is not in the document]
SimonHeybrock: For now I consider it useful.

src/beamlime/handlers/to_nx_log.py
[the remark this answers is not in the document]
YooSunYoung: Hmm... I don't remember... maybe I misread something....

YooSunYoung (approved):

src/beamlime/handlers/timeseries_handler.py
SimonHeybrock: Ideally I'd like to forward the dicts to ScippNexus. Currently I don't think that is so straightforward, but given that the `attrs` in HDF5 are dict-like I am not convinced it should be changed here. Obviously using pydantic for validation or when we send this via Kafka makes sense.

PR #348, opened 2025-03-17
Feed context into workflow in data reduction

SimonHeybrock:
First working demo of live reduction with "slow" log data in addition to "fast" event data.


src/beamlime/config/raw_detectors/bifrost.py
```diff
@@ -75,6 +77,18 @@ def make_spectrum_view(data: DetectorData[SampleRun]) -> SpectrumView:
...
 
+def make_counts_per_angle(
+    data: DetectorData[SampleRun], rotation: DetectorRotation
+) -> CountsPerAngle:
+    edges = sc.linspace('angle', 0, 91, num=46, unit='')
+    da = sc.DataArray(
+        sc.zeros(dims=['angle'], shape=[45], unit='counts'), coords={'angle': edges}
+    )
+    counts = sc.values(data.sum().data)
+    da['angle', rotation.data[-1]] += counts
```
nvaytet, line 88: What happens if the rotation given on input is outside of the 0-91 range? Can that ever happen?

src/beamlime/config/raw_detectors/bifrost.py
```diff
@@ -75,6 +77,18 @@ def make_spectrum_view(data: DetectorData[SampleRun]) -> SpectrumView:
     )
 
 
+def make_counts_per_angle(
+    data: DetectorData[SampleRun], rotation: DetectorRotation
+) -> CountsPerAngle:
+    edges = sc.linspace('angle', 0, 91, num=46, unit='')
```
nvaytet, line 83: Should angles have degree unit?

src/beamlime/config/raw_detectors/bifrost.py
nvaytet, line 88: What happens if the rotation given on input is outside of the 0-91 range? Can that ever happen?

src/beamlime/handlers/data_reduction_handler.py
```diff
@@ -89,19 +110,26 @@ def make_handler(
...
+        self, stream_processors: list[StreamProcessor], key: Key, is_context: bool
+    ) -> None:
         self._stream_processors = stream_processors
         self._key = key
+        self._is_context = is_context
 
     def add(self, timestamp: int, data: sc.DataArray) -> None:
-        for stream_processor in self._stream_processors:
-            stream_processor.accumulate({self._key: data})
+        if self._is_context:
```
nvaytet, line 127: Would there be more options in the future? (apart from `set_context` and `accumulate`?).
Is it worth making two different types of `MultiplexingProxy` that would do different things in their `add` method, instead of using the `is_context` boolean?

SimonHeybrock: I hope there will not be more. The proxy mechanism may change in the future. For example, I am currently working on supporting defining workflows and workflow params after service launch, which is currently not possible (but required, unless we want users to restart the service just to change reduction parameters). So this code is being touched/changed anyway.

nvaytet (approved):

PR #350, opened 2025-03-17
Define f144 attrs in instrument config modules

SimonHeybrock:
Could be part of #348, but #349 is in the middle and easier review independently.

We might want to read this from JSON or HDF template files (see #345) but defining it in the config is much easier for now. We may actually need very few logs, or they may rarely change. We may learn more about requirements and the template files as we progress to more complex use cases. I feel it can be an advantage to postpone a template-file-based implementation until then.

### DIFF
--- a/src/beamlime/handlers/data_reduction_handler.py
+++ b/src/beamlime/handlers/data_reduction_handler.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Mapping
 from typing import Any
 
 import scipp as sc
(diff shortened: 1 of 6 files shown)


src/beamlime/handlers/data_reduction_handler.py
```diff
@@ -38,21 +37,19 @@ class ReductionHandlerFactory(
...
         self._config = config
-        self._instrument = instrument
-        self._attrs_registry = attrs_registry
-        instrument_config = get_config(instrument)
-        self._processors = instrument_config.make_stream_processors()
-        self._source_to_key = instrument_config.source_to_key
+        self._instrument_config = instrument_config
+        self._processors = self._instrument_config.make_stream_processors()
+        self._source_to_key = self._instrument_config.source_to_key
+        self._attrs_registry = self._instrument_config.f144_attribute_registry
```
SimonHeybrock, line 49: It is likely that I will turn `instrument_config` into a class in the future. I will try to figure out the details when trying to implement workflow parameter setup and restarts, which can likely be partially encapsulated in such a class.

nvaytet (approved):

PR #352, opened 2025-03-24
Replace config subscriber thread with handler-based config update mechanism

SimonHeybrock:
While working on the mechanism to configure data reduction workflows at service runtime, it became clear that the current mechanism for listening to config (aka command) messages was insufficient and cumbersome. Previously, it worked as follows:

- `ConfigSubscriber` would consume from a config/command topic. It was running in a separate thread.
- All data messages where processed in the main thread, using a handler mechanism.
- The config subscriber was passed to handlers on handler creation, allowing handlers to get the latest config values.

For live data reduction, we needed a mechanism to "take action" on certain config/command messages. The `ConfigSubscriber` could not do this. It therefore made more sense to redo the mechanism from scratch, handling config/command messages with the handler mechanism, just like data messages. Key changes are as follows:

- `ConfigSubscriber` is removed and replaced by `ConfigHandler`.
- Config messages are fed into the stream processor, routing adapters take care of the different message decoding.
- The `ConfigHandler` is created ahead of time, data message handlers are created by the handler factory as before.


src/beamlime/core/processor.py
```diff
@@ -32,12 +32,12 @@ def __init__(
         logger: logging.Logger | None = None,
         source: MessageSource[Tin],
         sink: MessageSink[Tout],
-        handler_factory: HandlerFactory[Tin, Tout],
+        handler_registry: HandlerRegistry[Tin, Tout],
```
SimonHeybrock, line 35: We now have to pass the registry, since we want to be able to insert a custom handler for config messages, i.e., the processor cannot create the registry itself.

src/beamlime/kafka/scipp_da00_compat.py
SimonHeybrock, line 1: Changes in this file are unrelated. The `timeseries` helper service previously only worked with `--sink=png`, now it also works with `--sink=kafka`.

src/beamlime/services/fake_detectors.py
```diff
@@ -173,16 +167,18 @@ def serialize_detector_events_to_ev44(
...
 
 def run_service(*, instrument: str, log_level: int = logging.INFO) -> NoReturn:
-    service_name = f'{instrument}_fake_producer'
     kafka_config = load_config(namespace=config_names.kafka_upstream)
-    source = FakeDetectorSource(instrument=instrument)
     serializer = serialize_detector_events_to_ev44
-    processor = StreamProcessor(
-        source=source,
-        sink=KafkaSink(kafka_config=kafka_config, serializer=serializer),
+    builder = DataServiceBuilder(
```
SimonHeybrock, line 172: All the `fake_*` services were previously not using `DataServiceBuilder`. Now they can, by adding a new method to the builder. This is in a sense unrelated to the main change.

src/beamlime/kafka/message_adapter.py
```diff
@@ -185,6 +194,18 @@ def adapt(
...
+class BeamlimeCommandsAdapter(MessageAdapter[KafkaMessage, Message[Any]]):
+    """
+    Adapts a Kafka message to a Beamlime command message.
+    """
+
+    def adapt(self, message: KafkaMessage) -> Message[Any]:
+        key = MessageKey(topic=message.topic(), source_name='')
+        legacy_key = message.key().decode('utf-8')  # See 286
+        value = message.value()
+        return Message(key=key, timestamp=0, value={'key': legacy_key, 'value': value})
```
YooSunYoung, line 206: Kafka messages have timestamp (timestamp-type and actual-timestamp) so we can probably just use it, unless `0` has a specific meaning in this context.

https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Message.timestamp

src/beamlime/kafka/source.py
```diff
@@ -14,6 +14,29 @@ def close(self) -> None:
...
+    data topics vs. config/command topics.
+    """
+
+    def __init__(self, consumers):
+        self._consumers = consumers
+
+    def consume(self, num_messages: int, timeout: float) -> list[KafkaMessage]:
+        messages = []
+        for consumer in self._consumers:
+            messages.extend(consumer.consume(num_messages, timeout))
```
YooSunYoung, line 32: I think they shouldn't have same timeout for all different topics, unless it's very small `~0`.
If `num_message` and `timeout` are not optimized, some topics that have many messages might fall behind.

YooSunYoung (approved):
I have a minor comment and a bit of concerns about consuming messages, with potential blocking.
But we can just revisit this if we ever bump into this issue.

src/beamlime/kafka/message_adapter.py
SimonHeybrock: Ah, is that why I got a tuple of two ints as a timestamp? I tried using it first but could not make sense of it.

PR #353, opened 2025-03-24
Allow runtime "configuration" of reduction handlers

SimonHeybrock:
This adds a mechanism for (re)defining the stream processors (wrapping workflows) used to reduce detector data at runtime. In this first iteration this only supports predefined workflows with no parameters, but in principle the mechanism can be extended for this.

Note: Will move to a (hopefully) better command/config mechanism in a follow-up. Details of `ConfigHandler` and the action-registration mechanism will thus change.


src/beamlime/handlers/data_reduction_handler.py
```diff
@@ -55,100 +46,27 @@ def make_handler(
...
             preprocessor = ToNXevent_data()
         self._logger.info(
-            "Preprocessor %s is used for source name %s",
-            preprocessor.__class__.__name__,
-            key.source_name,
+            "%s using preprocessor %s", key.source_name, preprocessor.__class__.__name__
         )
+        self._logger.info("%s using accumulator %s", key.source_name, accumulator)
 
-        if (processor := self._processors.get(key.source_name)) is not None:
```
SimonHeybrock, line 76: Most of here and below moved to `workflow_manager.py` with little or no change.

src/beamlime/kafka/scipp_da00_compat.py
```diff
@@ -14,7 +14,13 @@ def scipp_to_da00(
             _to_da00_variable(signal_name, sc.values(da.data)),
             _to_da00_variable('errors', sc.stddevs(da.data)),
         ]
-    variables.extend([_to_da00_variable(name, var) for name, var in da.coords.items()])
+    variables.extend(
+        [
+            _to_da00_variable(name, var)
+            for name, var in da.coords.items()
+            if var.shape == var.values.shape  # vector3 etc. not supported currently
```
YooSunYoung, line 21: Should we somehow log if some coordinates are dropped...? Or do we not care?

src/beamlime/handlers/config_handler.py
```diff
@@ -31,10 +32,28 @@ def message_key(instrument: str) -> MessageKey:
...
         self._store = config
+        self._actions: dict[str, list[callable]] = {}
 
     def get(self, key: str, default=None):
         return self._store.get(key, default)
 
+    def register_action(
+        self,
+        key: str,
+        callback: callable,
```
YooSunYoung, line 43: `callback` sounds like something to be done after the actual action though.
How about just `action`...?

And I think `callable` is a function, not a type. (at least my linter complains)

src/beamlime/handlers/config_handler.py
```diff
@@ -57,6 +77,17 @@ def handle(self, messages: list[Message[bytes]]) -> list[Message[None]]:
...
+        # Delay action calls until all messages are processed to reduce triggering
+        # multiple calls for the same key in case of multiple messages with same key.
+        for key, value in updated.items():
+            for action in self._actions.get(key, []):
+                try:
+                    action(value)
+                except KeyError:  # noqa: PERF203
+                    self._logger.exception(
+                        'Error processing config action for %s:', key
+                    )
```
YooSunYoung, line 92: Not so important, but do we want to call actions only when it changes...? We could quickly check if the values are actually changed and then add it to `updated` dictionary.

src/beamlime/kafka/scipp_da00_compat.py
SimonHeybrock: We don't care for now, but we have to agree with ECDC on more details on the da00 results, i.e., what should be included.

src/beamlime/handlers/config_handler.py
SimonHeybrock: Yes, and I should specify the signature. Ok if I address this in the follow-up #354, where the signature changes?

src/beamlime/handlers/config_handler.py
YooSunYoung: It's fine if it's expected to be handled even if it has same value.

YooSunYoung (approved):

PR #354, opened 2025-03-31
Implement more fine-grained config mechanism

### LINKED ISSUE #288: Control message schema
Since we are considering using NICOS as a frontend we should consider using the same message schemas used for controlling just-bin-it.

*Edit: Latest info at* https://gitlab.esss.lu.se/ecdc/interface-specs/-/blob/main/nicos-beamlime.md

From Jonas:

Hello! Regarding how we communicate from NICOS to just-bin-it and filewriter right now, we are using byte string json like:
```python
producer.produce("some_topic", json.dumps(json_command).encode())
```
And for heartbeat/status we can use:
https://github.com/ess-dmsc/streaming-data-types/blob/master/schemas/x5f2_status.fbs

SimonHeybrock:
- Config keys now include source name and service name.
- Service name is used to filter messages. We can thus, e.g., clear the live detector view without clearing the data reduction (note the dashboard only implements a wildcard clear for now).
- Source name is used to configure individual handlers. Main driver for this is (for now) data reduction, where previously the source name was encoded in the message value. This was insufficient since due to Kafka topic compaction we would thus only see a subset of config messages.

Fixes #288.

### DIFF
--- a/src/beamlime/core/handler.py
+++ b/src/beamlime/core/handler.py
@@ -18,6 +18,16 @@ class Config(Protocol):
 
 
 class ConfigRegistry(Protocol):
+    """
+    Registry for configuration for different sources.
+
+    This is used by handlers to obtain configuration specific to a source. This protocol
+    is first and foremostly implemented by :py:class:`ConfigHandler`, which is used to
+    handle configuration messages, providing a mechanism to update the configuration
+    dynamically. The configuration is then used by the handlers to configure themselves
+    based on the source name of the messages they are processing.
+    """
+
     def get_config(self, source_name: str) -> Config:
         pass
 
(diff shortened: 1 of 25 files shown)


src/beamlime/core/handler.py
```diff
@@ -50,28 +79,18 @@ def __init__(
...
-    @staticmethod
-    def from_handler(
-        handler_cls: type[Handler[Tin, Tout]],
-    ) -> Callable[[logging.Logger | None, Config], CommonHandlerFactory[Tin, Tout]]:
-        def make(
-            *, logger: logging.Logger | None = None, config: Config
-        ) -> CommonHandlerFactory[Tin, Tout]:
-            return CommonHandlerFactory(
-                logger=logger, config=config, handler_cls=handler_cls
-            )
```
SimonHeybrock, line 72: Unrelated removal, was overlooked in an earlier change.

src/beamlime/kafka/message_adapter.py
```diff
@@ -30,15 +30,18 @@ def key(self) -> bytes:
     def value(self) -> bytes:
         pass
 
-    def timestamp(self) -> int:
+    def timestamp(self) -> tuple[int, int]:
```
SimonHeybrock, line 33: Should now match what we actually get from Kafka.

YooSunYoung (approved):
Now that we decided to use `service_name`, it'll be easier to split processes per service.
Maybe we should think about consumer group id or auto commit policy.

PR #357, opened 2025-04-02
Isolate from upstream topic source name

### LINKED ISSUE #355: Handle input topic and source names properly
Currently Beamlime assumes that source names are unique, but in really only the combination of source name and topic is unique. For example, detector data may have duplicate `source_name='caen'` with one detector per topic.

Internally Beamlime handles messages with a `MessageKey` that stores both, but code accesses the fields and thus will not work in general. Instead, we should probably process the topic and source_name combination into a unique derived source name early on in Beamlime (e.g., in the message adapters). The knowledge of topic names does not need to make it into the handlers, I think?

SimonHeybrock:
Summary of changes:
- Remove knowledge of `topic` and `source_name` from Beamlime internals. Instead, we now use `StreamId` internally.
- Add per-instrument config mapping from the former to the latter.
- As the above implies that topics are given by the mappings, the config files for topic names have been removed, to avoid having two sources of truth.
- Thanks to better separation of concerns (relating to `StreamMapping`), the service setup could be unified, resulting in a reduction of near-duplicate code.
- Added `--dev` flag, since we want to use a simpler topic setup for development. For example, DREAM and LOKI use one topic per detector in production.

Fixes #355.


src/beamlime/handlers/data_reduction_handler.py
```diff
@@ -69,24 +63,31 @@ def make_handler(
...
-            config = {}
-        else:
-            preprocessor = ToNXevent_data()
-            config = self._config_registry.get_config(key.source_name)
+        match key.kind:
+            case StreamKind.MONITOR_EVENTS | StreamKind.MONITOR_COUNTS:
+                # TODO The config for the preprocessor may need to be setup differently,
+                # e.g., to obtain the desired number of TOA bins. Or maybe we should
+                # not histogram in the preprocessor, i.e., not reuse the same
+                # preprocessor as in the monitor_data service.
```
SimonHeybrock, line 71: This comment is unrelated to the changes in this PR.

src/beamlime/kafka/consumer.py
```diff
@@ -56,28 +57,26 @@ def make_consumer_from_config(
...
+    topic = stream_kind_to_topic(instrument=instrument, kind=StreamKind.BEAMLIME_CONFIG)
     with make_consumer_from_config(
-        topics=['beamlime_commands'],
+        topics=[topic],
         config={**control_config, **kafka_downstream_config},
-        instrument=instrument,
         group='beamlime_commands',
     ) as consumer:
-        yield consumer
+        yield topic, consumer
```
SimonHeybrock, line 82: This is a bit awkward, but I could not come up with a better way right now: `DataServiceBuilder` needs to know which topic is the config topic, since it needs a separate consumer to subscribe to. Otherwise it only has a `RouteByTopicAdapter`, which has a list of *all* topics.

I don't think this will cause much trouble because it will not be used in a ton of places.

PR #360, opened 2025-04-07
Properly shutdown service on exception in message adapter

SimonHeybrock:
This also caused unit tests to hang. I don't know if this is also related to some of the hard-to-trace errors seen in deployment @MridulS.

### DIFF
--- a/src/beamlime/core/service.py
+++ b/src/beamlime/core/service.py
@@ -137,7 +137,7 @@ class Service(ServiceBase):
                 remaining = max(0.0, self._poll_interval - elapsed)
                 if remaining > 0:
                     time.sleep(remaining)
-        except (RuntimeError, ValueError):
+        except Exception:
             self._logger.exception("Error in service loop")
             self._running = False
             # Send a signal to the main thread to unblock it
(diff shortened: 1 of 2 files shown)


src/beamlime/core/service.py
```diff
@@ -137,15 +137,18 @@ def _run_loop(self) -> None:
                 remaining = max(0.0, self._poll_interval - elapsed)
                 if remaining > 0:
                     time.sleep(remaining)
-        except RuntimeError:
+        except (RuntimeError, ValueError):
             self._logger.exception("Error in service loop")
-            self.stop()
```
SimonHeybrock, line 142: Calling stop here didn't work, since it tried to stop itself.

src/beamlime/core/service.py
```diff
@@ -137,15 +137,18 @@ def _run_loop(self) -> None:
                 remaining = max(0.0, self._poll_interval - elapsed)
                 if remaining > 0:
                     time.sleep(remaining)
-        except RuntimeError:
+        except (RuntimeError, ValueError):
```
MridulS, line 140: Does it make sense to do a catch all here?
```suggestion
        except Exception:
```
We do want to stop the loop if we trigger *any* unexpected error.

MridulS (approved):

PR #361, opened 2025-04-07
Add more message adapter tests

SimonHeybrock:
Also fix some small bugs in Kafka adapters, which ignored the `stream_lut`.

### DIFF
--- a/tests/kafka/message_adapter_test.py
+++ b/tests/kafka/message_adapter_test.py
@@ -432,6 +432,7 @@ class TestAdaptingMessageSource:
             source=TestMessageSource(),
             adapter=KafkaToDa00Adapter(stream_kind=StreamKind.DETECTOR_EVENTS),
             logger=fake_logger,
+            raise_on_error=False,
         )
 
         messages = adapting_source.get_messages()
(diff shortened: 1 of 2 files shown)


tests/kafka/message_adapter_test.py
```diff
@@ -137,37 +293,418 @@ def message_with_schema(schema: str) -> KafkaMessage:
...
+            logger=self.fake_logger,
+        )
+
+        # The exception should be caught and logged, then re-raised
+        with pytest.raises(ValueError, match="Failed to deserialize corrupt ev44 data"):
+            source.get_messages()
+
+        assert len(self.fake_logger.exception_calls) == 1
+        assert (
+            "error adapting message" in self.fake_logger.exception_calls[0][0].lower()
```
YooSunYoung, line 574:
```suggestion
            "error adapting message" in self.fake_logger.exception_calls[-1][0].lower()
```

Should we rather use [-1] index because it should be the latest one?
Also it won't break even if we insert more log related tests above.

YooSunYoung: Ah sorry now I understood. The fake logger is set up before every method. Feel free to ignore this suggestion.

YooSunYoung (approved):

PR #364, opened 2025-04-07
Move instrument specific code/config into instrument files

SimonHeybrock:
There should be zero functional changes here. Just renaming and moving.


src/beamlime/config/instruments/bifrost.py
```diff
@@ -144,3 +171,29 @@ def _all() -> StreamProcessor:
...
+    }
+
+
+stream_mapping_dev = make_dev_stream_mapping(
+    'bifrost', detectors=list(detectors_config['fakes'])
+)
+stream_mapping = StreamMapping(
+    **make_common_stream_mapping_inputs(instrument='bifrost'),
+    detectors=_make_bifrost_detectors(),
+)
```
MridulS, line 199: Just a subjective nit pick, but maybe it would be nice to have something like 
```suggestion
stream_mapping['default'] = make_stream_mapping(instrument='bifrost',  detectors=_make_bifrost_detectors, env='default')
stream_mapping['dev'] = make_stream_mapping(instrument='bifrost',  detectors=detectors_config['fakes'], env='dev')
```
It would be easier to add more keys if we want to run in multiple different envs, and makes it look consistent.

It's fine either way!

MridulS (approved):

PR #371, opened 2025-05-02
Fix config action calls

SimonHeybrock:
The update tracking mechanism was broken, discarding updates for same key but different sources.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -99,7 +99,7 @@ requests==2.32.3
     # via pooch
 sciline==25.4.1
     # via essreduce
-scipp==25.4.0
+scipp==25.5.0
     # via
     #   -r base.in
     #   essreduce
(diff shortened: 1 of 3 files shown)


src/beamlime/handlers/config_handler.py
```diff
@@ -110,40 +111,48 @@ def handle(self, messages: list[Message[RawConfigItem]]) -> list[Message[None]]:
...
-                    for store in self._stores.values():
-                        store[key] = value
+                    self._global_store[config_key] = value
+                    for source_name, store in self._stores.items():
+                        updated[config_key][source_name] = update
+                        store[config_key] = value
                 else:
-                    self.get_config(source_name)[key] = value
+                    updated[config_key][source_name] = update
+                    self.get_config(source_name)[config_key] = value
```
YooSunYoung, line 139: This line can also be done after collecting all updated configurations.

It doesn't matter so much, but we might want to watch this configuration object to cache it or sth... in the future....?
But feel free to ignore this comment.

YooSunYoung (approved):
Seems reasonable to me.

PR #372, opened 2025-05-02
Fix flipped x and y in ROI tool

### LINKED ISSUE #367: ROI visualization on dashboard.
The ROI showed in the NMX panel was flipped.
There was one out of 4 sub-pannels turned off so we found out that ROI drawn on the raw detector view was upside down.
Maybe it's just configuration issue...? 
Just taking notes here to look into it later...

SimonHeybrock:
Fixes #367


nvaytet (approved):

nvaytet:
I didn't actually check if this fixed the problem in real life, but the changes look reasonable :wink:

PR #374, opened 2025-05-07
Workflow config

### LINKED ISSUE #368: Publish available workflows and parameters to Kafka
Currently the dashboard reads the available workflows by importing a Python module (specific to the instrument). This is a hack that works for our dashboard, but it will not for Nicos since we cannot or do not want to force Nicos to depend on (and keep updated versions of) the instrument packages.

A simpler variant of this is the regular config options, which are currently hard-coded. This is probably fine for Nicos, as long as it does not change frequently.

Instead, we should probably publish available config to a compacted Kafka topic. This could include basic options (such as TOA range, ...), but for the data-reduction service it should include a complete specification of the input parameters of a workflow, with sufficient detail to generate input widgets from it.

Inputs also need to be validated. We should have a think about whether this can/should happen client side, or if it has to be server side. Since some failures can be detected only when running the workflow it may be a combination of both?

SimonHeybrock:
Fixes #368.

Please review #371 first, this is the base of this PR.

The changes in this PR and the dashboard additions allow for controlling workflows like this:
image

Note that currently only Loki has configured workflows with parameters, and even those are meant only for testing. The plan is still to migrate concrete workflows to the instrument packages.


src/beamlime/config/models.py
```diff
@@ -179,3 +179,110 @@ def from_string(cls, key_str: str) -> ConfigKey:
...
+    """
+
+    name: str = Field(description="Name of the parameter.")
+    unit: str | None = Field(default=None, description="Unit of the parameter.")
+    description: str = Field(description="Description of the parameter.")
+    param_type: ParameterType = Field(description="Type of the parameter.")
+    default: T | None = Field(
+        default=None, description="Default value of the parameter."
+    )
+    options: list[Any] | None = Field(
```
MridulS, line 215:
```suggestion
    options: list[T] | None = Field(
```

src/beamlime/config/models.py
```diff
@@ -179,3 +179,110 @@ def from_string(cls, key_str: str) -> ConfigKey:
...
+
+    @model_validator(mode='after')
+    def validate_options(self) -> Parameter:
+        """Validate that options are provided for OPTIONS parameter type."""
+        if self.param_type == ParameterType.OPTIONS:
+            if self.options is None or not self.options:
+                raise ValueError(
+                    "Options must be provided for parameter type 'OPTIONS'"
+                )
+            if self.default not in set(self.options):
```
MridulS, line 228:
```suggestion
            if self.default not in self.options:
```
?

MridulS (commented):
I've tested this locally yet, but the overall structure looks okay to me.

src/beamlime/config/models.py
SimonHeybrock: I think I got linter complaints without the `set`.

MridulS (approved):

PR #375, opened 2025-05-09
Add tests for data reduction service

SimonHeybrock:
Adding tests highlighted a few problems. Therefore this also includes the addition of `Instrument` for wrapping instrument config as well as making `StreamProcessor` and instrument-specific instance instead of a singleton.

### DIFF
--- a/src/beamlime/config/instrument.py
+++ b/src/beamlime/config/instrument.py
@@ -2,7 +2,8 @@
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 from __future__ import annotations
 
-from collections.abc import Callable, Iterator, Mapping, Sequence
+from collections import UserDict
+from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field
 from typing import Any
 
@@ -12,7 +13,7 @@ from beamlime.config.models import Parameter
 from beamlime.handlers.stream_processor_factory import StreamProcessorFactory
 
 
-class InstrumentRegistry(Mapping[str, 'Instrument']):
+class InstrumentRegistry(UserDict[str, 'Instrument']):
     """
     Registry for instrument configurations.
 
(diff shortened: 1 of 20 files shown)


src/beamlime/config/instrument.py
```diff
@@ -0,0 +1,113 @@
...
+    def __contains__(self, name: str) -> bool:
+        """Check if an instrument is registered."""
+        return name in self._instruments
+
+
+instrument_registry = InstrumentRegistry()
+
+
+@dataclass(frozen=True, kw_only=True)
+class Instrument:
```
SimonHeybrock, line 59: There are some other required variables in the `instrument/<name>.py` files which should be moved into this class, but I will leave that for future work.

tests/services/data_reduction_test.py
```diff
@@ -0,0 +1,394 @@
...
+    assert len(sink.messages) == 3
+    assert sink.messages[2].value.values.sum() == 7000
+
+    # Stop workflow
+    app.publish_config_message(key=config_key, value=None)
+    app.publish_events(size=1000, time=10)
+    service.step()
+    app.publish_events(size=1000, time=20)
+    service.step()
+    assert len(sink.messages) == 3
```
MridulS, line 98:
```suggestion
    assert len(sink.messages) == 3
    service.stop()
```

to clean up properly or not required?

MridulS: And same for tests below.

src/beamlime/handlers/workflow_manager.py
```diff
@@ -67,10 +58,22 @@ def __init__(self, *, source_to_key: dict[str, Key]) -> None:
...
+        Get the workflow specifications for the available workflows.
+
+        Returns
+        -------
+        WorkflowSpecs
+            The workflow specifications.
+        """
+        return WorkflowSpecs(workflows=dict(self._processor_factory.items()))
+
     def set_worklow(self, source_name: str, processor: StreamProcessor | None) -> None:
```
MridulS, line 77: not in this PR, but copilot did complain about this typo too.

tests/services/data_reduction_test.py
SimonHeybrock: The service is never actually started. The tests call `step` which does one "iteration" of a running service. This removes complications such as having to wait until the service thread (made when starting) did the next iteration, etc.

src/beamlime/config/instrument.py
```diff
@@ -0,0 +1,113 @@
...
+        """Return an iterator over instrument names (implements Mapping)."""
+        return iter(self._instruments)
+
+    def __len__(self) -> int:
+        """Return the number of registered instruments (implements Mapping)."""
+        return len(self._instruments)
+
+    def __contains__(self, name: str) -> bool:
+        """Check if an instrument is registered."""
+        return name in self._instruments
```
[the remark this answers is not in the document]
SimonHeybrock: Thanks for the suggestion. For some reason `UserDict` is the one class I always forget about!

MridulS (approved):

PR #376, opened 2025-05-14
Release prep; tox and copier bump

MridulS:


SimonHeybrock:
Builds might fail because `essspectroscopy` is currently not compatible with the latest `essreduce`.

SimonHeybrock:
> Builds might fail because `essspectroscopy` is currently not compatible with the latest `essreduce`.

See #381.

PR #382, opened 2025-05-19
Some more resilience tests

SimonHeybrock:
Also fixed a bunch of typing/pylance issues.


src/beamlime/config/defaults/raw_data_consumer.yaml
```diff
@@ -1,5 +1,6 @@
 auto.offset.reset: latest
 enable.auto.commit: true
 fetch.min.bytes: 1
-session.timeout.ms: 6000
-heartbeat.interval.ms: 2000
\ No newline at end of file
+fetch.max.bytes: 200000000  # might need at least 10x this in RAM for processing
+session.timeout.ms: 30000
+heartbeat.interval.ms: 10000  # long so we do not timeout when processing large data
```
SimonHeybrock, line 6: I have not really tried these parameters, but the old 2 seconds could easily be exceeded by data reduction, especially when processing the initial messages when the workflow computes all static intermediate results.

PR #385, opened 2025-06-06
Add Dream workflow

SimonHeybrock:
This depends on unmerged and unreleased changes in ESSdiffraction. PR should thus fail until we can update.

### DIFF
--- a/src/beamlime/config/instruments/dream.py
+++ b/src/beamlime/config/instruments/dream.py
@@ -3,17 +3,34 @@
 
 from typing import NewType
 
-import sciline
 import scipp as sc
+from ess.reduce.nexus.types import (
+    DetectorData,
+    Filename,
+    NeXusData,
+    NeXusName,
+    SampleRun,
+)
 from ess.reduce.streaming import StreamProcessor
+from ess.reduce.time_of_flight.workflow import GenericTofWorkflow
+from scippnexus import NXdetector
 
 from beamlime.config import Instrument
 from beamlime.config.env import StreamingEnv
+from beamlime.handlers.detector_data_handler import get_nexus_geometry_filename
 from beamlime.kafka import InputStreamKey, StreamLUT, StreamMapping
 
 from ._ess import make_common_stream_mapping_inputs, make_dev_stream_mapping
 
-instrument = Instrument(name='dream')
+instrument = Instrument(
+    name='dream',
+    source_to_key={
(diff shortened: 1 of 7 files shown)


src/beamlime/config/instruments/dream.py
```diff
@@ -167,6 +120,91 @@ def _make_dream_detectors() -> StreamLUT:
...
+        )
+    )
+
+
+_reduction_workflow.insert(_total_counts)
+_reduction_workflow[powder.types.DspacingBins] = sc.linspace(
+    dim='dspacing',
+    start=0.1,
+    stop=5.0,
+    num=2000,
```
nvaytet, line 154: Do we really need that many bins?

src/beamlime/config/instruments/dream.py
```diff
@@ -167,6 +120,91 @@ def _make_dream_detectors() -> StreamLUT:
...
+    wf[NeXusName[NXdetector]] = source_name
+    wf[Filename[SampleRun]] = GeometryFile
+    return StreamProcessor(
+        wf,
+        dynamic_keys=(
+            NeXusData[NXdetector, SampleRun],
+            NeXusData[powder.types.CaveMonitor, SampleRun],
+        ),
+        target_keys=(
+            powder.types.FocussedDataDspacing[SampleRun],
```
nvaytet, line 198: Is it also possible to make an example where we load a Vanadium file and display a normalized result?
Maybe it's beyond the scope of this PR...

src/beamlime/dashboard/plots.py
```diff
@@ -29,7 +30,9 @@ def create_monitor_plot(key: str, data: sc.DataArray) -> go.Figure:
     return fig
 
 
-def create_detector_plot(key: str, data: sc.DataArray) -> go.Figure:
+def create_detector_plot(
+    key: str, data: sc.DataArray, *, logscale: bool = False
```
nvaytet, line 34: add `logscale` to parameters in docstring?

nvaytet (commented):
looks good

src/beamlime/config/instruments/dream.py
SimonHeybrock: Powder is high res, so we might.

src/beamlime/config/instruments/dream.py
SimonHeybrock: It is possible, PRs will be accepted ;)

PR #389, opened 2025-06-20
Dashboard rewrite using Panel+Holoviews

SimonHeybrock:
This is still lacked a ton of tests, but I would like to have feedback before implementing those.

I have a follow-up branch with even more additions (implementing data reduction dashboard) as well as some tweaks of what is added here (but no major structural changes). I let the reviewer decide if I should make an even larger PR. If not, some comments here might only get address in a follow up PR.


src/beamlime/kafka/message_adapter.py
```diff
@@ -343,6 +343,3 @@ def get_messages(self) -> Sequence[U]:
                 if self._raise_on_error:
                     raise
         return adapted
-
-    def close(self) -> None:
-        self._source.close()
```
SimonHeybrock, line 348: The `close` mechanism was removed, since it was redundant with the `close()` calls from the contextmanagers used to create the Kafka consumers.

src/beamlime/core/config_service.py
SimonHeybrock, line 1: Note that this `ConfigService` is used by the *old* dashboard, which we are not removing yet in this PR. There is also a new `ConfigService` in the new `dashboard` folder.

jl-wynen:
Is there an architecture diagram? You showed me something in a call a couple of days ago

SimonHeybrock:
> Is there an architecture diagram? You showed me something in a call a couple of days ago

Added in #391.

SimonHeybrock:
There were some refactors in the follow up. It also adds tests, so I'll close this in favor of a larger PR.

PR #393, opened 2025-06-23
Add Dream workflow

SimonHeybrock:
Extracted from #392. Therefore it also includes a preparatory split of the `models.py` file. No functional change there. Otherwise, this is mostly the same as #385.


tests/services/data_reduction_test.py
```diff
@@ -405,10 +410,10 @@ def configured_dummy_reduction() -> BeamlimeApp:
...
 
-    config_key = models.ConfigKey(
+    config_key = ConfigKey(
         source_name='panel_0', service_name="data_reduction", key="workflow_config"
     )
-    workflow_config = models.WorkflowConfig(
+    workflow_config = workflow_spec.WorkflowConfig(
         identifier=workflow_id,
         values={param.name: param.default for param in spec.parameters},
     )
```
nvaytet, line 419: [In the other PR](https://github.com/scipp/beamlime/pull/385/files#diff-5a2b9847a56956235eb06d4bca7dc2d81562ed929ffa1b7a040899241f205a54), there were more changes in these test file.
Are they missing here or were they intentionally not included / no longer needed?

PR #401, opened 2025-07-02
Remove legacy dashboard

SimonHeybrock:
Also updates docs, adding missing WSGI stuff.


src/beamlime/dashboard/monitors.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, *, instrument: str = 'dummy', dev: bool = False, log_level: i
             dev=dev,
             log_level=log_level,
             dashboard_name='monitors_dashboard',
+            port=5007,  # Default port for monitors dashboard
```
jl-wynen, line 29: Why is this hardcoded here? Shouldn't it be read from an env var?

src/beamlime/dashboard/reduction.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, *, instrument: str = 'dummy', dev: bool = False, log_level: i
             dev=dev,
             log_level=log_level,
             dashboard_name='reduction_dashboard',
+            port=5009,  # Default port for reduction dashboard
```
jl-wynen, line 26: Same as for the monitor dashboard.

jl-wynen (approved):

PR #412, opened 2025-07-18
Unit format

### LINKED ISSUE #399: > Or are you suggesting we should try to override how Holoviews shows units?
> Or are you suggesting we should try to override how Holoviews shows units?

Yes. If this is simple enough. Otherwise, go with parentheses.

_Originally posted by @jl-wynen in https://github.com/scipp/beamlime/pull/396#discussion_r2177677061_

jokasimr:
Fixes #399


pyproject.toml
```diff
@@ -52,6 +52,7 @@ test = [
     "param",
     "pytest",
     "pytest-benchmark",
+    "holoviews",
```
jl-wynen, line 55: This seems wrong. If we need holoviews for testing, then he have a dashboard test. So shouldn't that use the 'dashboard' extra?

src/beamlime/dashboard/__init__.py
```diff
@@ -1,2 +1,5 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+import holoviews as hv
+
+hv.Dimension.unit_format = ' [{unit}]'
```
jl-wynen, line 5: Since this is a global setting, I think it should be set in the startup routine, not on import. This is to ensure that it runs in a predictable manner and only when desired.

YooSunYoung:
Is the test fail related to this fix...? Otherwise it looks fine to have it in the Base class __init__.

jokasimr:
> Is the failing test related to this fix

I think it is not related to this fix but related to recent changes in essreduce

YooSunYoung (approved):

pyproject.toml
SimonHeybrock: I don't think there are tests that use Holoviews, that is why it was not in `test`.

@jl-wynen The `dashboard` extra is to avoid installing GUI bits in the backend. We could still have tests that need (a subset) of what is in that extra, in which case having it in `test` should be fine. No need to install all GUI deps for tests?

PR #428, opened 2025-09-01
Identifier overhaul

SimonHeybrock:
This PR makes several changes related to IDs for workflows, jobs, and result data. It replaces some legacy mechanism, and will allow for running multiple jobs as well as a better mechanism for identifying results on the frontend.

Note: The current static way in which the front-end ties specific plots to some expect result-data-keys is broken by this change. As a follow-up, I am planning to implement a mechanism that allows for a more dynamic plot creation, probably by having a registry of plot types linked to specific result-data-keys.


SimonHeybrock:
> There are merge conflicts now.

Fixed, thanks!

tests/core/job_test.py
```diff
@@ -1,17 +1,67 @@
...
+            start_time=100,
+            end_time=200,
+            data=sc.DataArray(sc.scalar(3.14)),
+            error_message=None,
+        )
+        assert result.stream_name == (
+            '{"workflow_id":{"instrument":"TEST","namespace":"data_reduction",'
+            '"name":"test_workflow","version":1},"job_id":{"source_name":"test_source",'
+            '"job_number":"' + str(job_number) + '"},"output_name":"output1"}'
+        )
```
MridulS, line 62:
```suggestion
@pytest.mark.parametrize("output_name, output_result", [(None, 'null'), ('output1', '"output1"')])
def test_stream_name(output_name, output_result):
    workflow_id = WorkflowId(
        instrument="TEST",
        namespace="data_reduction",
        name="test_workflow",
        version=1,
    )
    job_number = uuid.uuid4()
    job_id = JobId(source_name="test_source", job_number=job_number)
    result = JobResult(
        job_id=job_id,
        workflow_id=workflow_id,
        output_name=output_name,
        start_time=100,
        end_time=200,
        data=sc.DataArray(sc.scalar(3.14)),
        error_message=None,
    )
    assert result.stream_name == (
        '{"workflow_id":{"instrument":"TEST","namespace":"data_reduction",'
        '"name":"test_workflow","version":1},"job_id":{"source_name":"test_source",'
        f'"job_number":"{job_number}"},"output_name":{output_result}}'
    )
```

MridulS: What about something like this? (haven't tested the string bits at the end)

MridulS (approved):
Nothing major blocking here.

PR #431, opened 2025-09-03
Refactor frontend to job-based mechanism, begin adding generic plotting functionality

SimonHeybrock:
This continues from the backend refactor:

- Frontend becomes job-aware by adding a `JobService` and a widget that displays results of jobs.
- Since we can no longer link received data to a fixed workflow we have to change the plotting mechanism.
- Plots can now be dynamically created from job results. This avoids the hard-coded dependence on certain results of the old implementation (see `reduction.py`).
- Plots can be configured, using the same widget code used for configuring workflows (based on Pydantic models).
- Removed the `monitors.py` and `detectors.py` dashboards. For now `reduction.py` lets us start and plot all possible kinds of workflows and their data. Unless there are plotting performance issues we cannot solve we can stick with a unified dashboard, but this is still subject to future changes.

To do in future work (these are tangential to the actual change, so I am postponing to keep the PR size more manegeable):
- The workflow status list widget should be removed. We instead need to add stop buttons in the job list.
- Refactor existing and add better plotters. In this PR I am simply wrapping the old implementations, with insufficient validation, and ignore the configured plot options.
- Mechanism to persist plotting config, or auto-create plots if certain workflows are run.
- Layout of the plots. Currently I just naively add a new tab for every plot.

### DIFF
--- a/src/beamlime/dashboard/widgets/plot_creation_widget.py
+++ b/src/beamlime/dashboard/widgets/plot_creation_widget.py
@@ -320,7 +320,8 @@ class PlotCreationWidget:
         # Create and show configuration modal
         modal = ConfigurationModal(config=config, start_button_text="Create Plot")
 
-        # Add modal to container and show it
+        # Clear previous modals and add the new one
+        self._modal_container.clear()
         self._modal_container.append(modal.modal)
         modal.show()
 


src/beamlime/dashboard/workflow_config_service.py
```diff
@@ -48,9 +48,8 @@ class ConfigServiceAdapter(WorkflowConfigService):
     Adapter to make ConfigService compatible with WorkflowConfigService protocol.
     """
 
-    def __init__(self, config_service: ConfigService, source_names: list[str]):
+    def __init__(self, config_service: ConfigService):
         self._config_service = config_service
-        self._source_names = source_names
```
SimonHeybrock, line 53: This was unused, unrelated change.

src/beamlime/dashboard/widgets/plot_creation_widget.py
```diff
@@ -0,0 +1,375 @@
...
+                            'job_number': job_number.hex,
+                        }
+                        for output_name in sorted(output_names)
+                    ]
+                )
+
+        if job_output_data:
+            # Convert to DataFrame for Tabulator widget
+            df = pd.DataFrame(job_output_data)
+            self._job_output_table.value = df
```
nvaytet, line 233:
```suggestion
```

src/beamlime/dashboard/widgets/plot_creation_widget.py
```diff
@@ -0,0 +1,375 @@
...
+        self._update_dependent_widgets()
+
+    def _update_dependent_widgets(self) -> None:
+        """Update plot selector based on job and output selection."""
+        if self._selected_job is None:
+            # No job selected - disable everything
+            self._plot_selector.options = []
+            self._plot_selector.value = None
+            self._plot_selector.disabled = True
+            self._create_button.disabled = True
```
nvaytet, line 267: These 4 lines are repeated 4 times; move to a separate method?

src/beamlime/dashboard/widgets/plot_creation_widget.py
```diff
@@ -0,0 +1,375 @@
...
+            job_number=self._selected_job,
+            output_name=self._selected_output,
+            plot_spec=spec,
+            available_sources=available_sources,
+            plotting_controller=self._plotting_controller,
+            success_callback=self._on_plot_created,
+        )
+
+        # Create and show configuration modal
+        modal = ConfigurationModal(config=config, start_button_text="Create Plot")
```
nvaytet, line 330: Will you ever need to access the modal at a later stage? Should it be a class attribute?

src/beamlime/dashboard/workflow_controller.py
```diff
@@ -258,15 +260,7 @@ def remove_workflow_for_source(self, source_name: str) -> None:
...
-                key
-                for key in self._data_service.keys()
-                if (
-                    isinstance(key, DataKey)
-                    and key.service_name == 'data_reduction'
-                    and key.source_name == source_name
-                )
-            ]
+            keys_to_remove = list(self._data_service.keys())
             for key in keys_to_remove:
```
nvaytet, line 264:
```suggestion
            for key in self._data_service.keys():
```

src/beamlime/dashboard/widgets/plot_creation_widget.py
SimonHeybrock: My plan is to allow for re-opening the modal from a plot tab. But I think we would recreate it based on config, not reuse it. We don't want 20 of them floating about.

PR #432, opened 2025-09-04
Add JobController and JobControlWidget

SimonHeybrock:
This adds the ability to control jobs (currently: reset and stop). The workflow-based mechanism is removed.

The widget is a placeholder and many improvements will be needed in the future. Time will tell if and how this should be unified with the job *data* widget, which is used for creating plots.

Update: Some if the widget's functionality will be replaced by `JobStatusWidget`, which will replace job-specific controls.

Subtlety:
- Beamlime uses a compacted Kafka topic for "config" messages.
- Previously we used source name and namespace in the key, so we could always get the latest value for any given kind.
- This does not work any more since we can now have multiple workflows for a given source, for example.
- This should be refactored to, e.g., include the job number in the Kafka message key.
- I am not doing this here, since we anyway want to split up the "abused" config topic to be more specific. I hope we will then have a cleared picture of how to handle key and compaction.

Next (not included here):
- While we can stop jobs, the backend does not report job status yet (and the old workflow-status reporting is defunct). That is, we cannot see from the job lists if a job is still running/producing data.


src/beamlime/config/workflow_spec.py
```diff
@@ -179,6 +191,33 @@ class PersistentWorkflowConfigs(BaseModel):
         description="Persistent configurations indexed by workflow ID",
     )
 
+    @field_serializer('configs')
+    def serialize_configs(
+        self, configs: dict[WorkflowId, PersistentWorkflowConfig]
+    ) -> dict[str, Any]:
+        """Serialize configs by converting WorkflowId keys to strings."""
+        return {str(k): v.model_dump() for k, v in configs.items()}
```
SimonHeybrock, line 199: A while ago I changed `WorkflowId` from a string to a dataclass. I noticed now that this broke the workflow-config-persist mechanism, since Pydantic does not like the new ID type as a dict key in serialization. This fixes this.

src/beamlime/config/workflow_spec.py
```diff
@@ -29,6 +28,19 @@ class WorkflowId:
...
+        """Parse WorkflowId from string representation."""
+        parts = workflow_id_str.split('/')
+        if len(parts) != 4:
+            raise ValueError(f"Invalid WorkflowId string format: {workflow_id_str}")
+        return WorkflowId(
+            instrument=parts[0],
+            namespace=parts[1],
+            name=parts[2],
+            version=int(parts[3]),
+        )
```
YooSunYoung, line 42: Are these always used in the streaming? i.e. we won't save it to a file and load it later with potentially different version of beamlime?
Or when one of services start up with a new version of beamlime and pick up the latest message from the kafka broker...

src/beamlime/dashboard/widgets/job_control_widget.py
YooSunYoung, line 1: When I selected a workflow, the panel created a huge space before the job controller widget like this.

Do you know why....?

image
image

src/beamlime/dashboard/widgets/job_control_widget.py
```diff
@@ -0,0 +1,239 @@
...
+                self._job_select,
+                pn.Row(
+                    self._pause_btn,
+                    self._resume_btn,
+                    self._stop_btn,
+                    self._reset_btn,
+                ),
+            ),
+            # Status section
+            pn.Column(pn.pane.HTML("Status:"), self._status_text),
```
YooSunYoung, line 140: [Minor] I would like to have the status section at the top of the job controller but I guess we can fix that later when we have other bottons working too...

SimonHeybrock: This file is removed in a follow-up, see PR description my comments before your review.

src/beamlime/dashboard/widgets/job_control_widget.py
YooSunYoung: Ah you entirely removed it from this view I see...

YooSunYoung (approved):
Workflow and Job handling seems to be clear enough.
I don't have any objection.

PR #440, opened 2025-09-10
Status heartbeat

### LINKED ISSUE #436: Split config topic into command and status
- For status, we should consider using https://github.com/ess-dmsc/streaming-data-types/blob/master/schemas/x5f2_status.fbs. Me idea would be to include the status of every job in the `status_json`. See also [Nicos compatible status message](https://gitlab.esss.lu.se/ecdc/interface-specs/-/blob/477da77510561f1a6afee874f7c8bfe907183723/nicos-beamlime.md#heartbeat-message-x5f2).

SimonHeybrock:
Add job status heartbeat based on x5f2 messages. Status is used to populate a job status list widget. We also use this to replace the temporary job control widget from an earlier PR.

image

Fixes #436.

I think `JobController` will needs some more iterations, and I am postponing writing tests for that to a follow-up.


src/beamlime/core/job.py
```diff
@@ -320,40 +397,144 @@ def reset_job(self, job_id: JobId) -> None:
...
         """Push data into the active jobs and return status for each job."""
         self._advance_to_time(data.start_time, data.end_time)
-        job_statuses: list[JobStatus] = []
+        job_statuses: list[JobError] = []
         for job in self.active_jobs:
             status = job.add(data)
             job_statuses.append(status)
+            # Track warnings from job operations, or clear them on success
+            if status.has_error and status.error_message is not None:
+                self._job_warning_messages[job.job_id] = status.error_message
```
nvaytet, line 426: I'm sort of getting confused here as to what is an error and what is a warnings (also in the `compute_results` below)
If I followed correctly, `status` here is a `JobError`. But we're putting it in the `warning_messages`...

src/beamlime/core/job.py
```diff
@@ -320,40 +397,144 @@ def reset_job(self, job_id: JobId) -> None:
...
+            self._active_jobs.get(job_id)
+            or self._scheduled_jobs.get(job_id)
+            or self._stopped_jobs.get(job_id)
+        )
+        if job is None:
+            return None
+
+        # Determine current state based on job's location in manager
+        if job_id in self._active_jobs:
+            if job_id in self._finishing_jobs:
```
nvaytet, line 481: We're having to check in various places if the job is in `_finishing_jobs`, or in `_active_jobs`. Can we set the type of job on the job object itself? Would this make things easier?

Maybe it was a way to categorize the jobs so that we don't need to scan all the jobs if we are looking only for fnishing jobs?
Do we expect to have thousands of jobs? If there are only a few jobs running, maybe iterating over all jobs is fast enough?

src/beamlime/core/orchestrating_processor.py
```diff
@@ -135,12 +140,12 @@ def process(self) -> None:
...
         # Handle data messages with the workflow manager, accumulating data as needed.
-        job_statuses = self._job_manager.push_data(workflow_data)
+        job_errors = self._job_manager.push_data(workflow_data)
 
         # Log any errors from data processing
-        for status in job_statuses:
-            if status.has_error:
-                self._logger.error(self._job_manager.format_job_error(status))
+        for error in job_errors:
+            if error.has_error:
```
nvaytet, line 147: Maybe related to the comment above about confusion as to what is warning and what is error:
I find it strange that an `error` potentially does not have an error...

src/beamlime/dashboard/job_service.py
```diff
@@ -72,23 +157,63 @@ def data_updated(self, updated_keys: set[ResultKey]) -> None:
...
+        )
+
+        # Notify all subscribers
+        for callback in self._job_status_update_subscribers:
+            try:
+                callback()
+            except Exception as e:
+                self._logger.error("Error in job status update callback: %s", e)
+
+    def remove_job(self, job_id: JobId) -> None:
```
nvaytet, line 189: Does the order of steps in the function matter? If so, can you add a comment about it?

src/beamlime/dashboard/widgets/job_status_widget.py
```diff
@@ -0,0 +1,547 @@
...
+        """Remove a job widget."""
+        if job_key in self._status_widgets:
+            self._status_widgets.pop(job_key)
+            if job_key in self._widget_panels:
+                widget_panel = self._widget_panels.pop(job_key)
+                # Remove from the job list
+                try:
+                    self._job_list.remove(widget_panel)
+                except ValueError:
+                    # Panel might not be in the list anymore
```
nvaytet, line 505: What happens if Panel is not in the list anymore? Can you describe in which situation this arises?

src/beamlime/core/job.py
```diff
@@ -74,6 +75,31 @@ def stream_name(self) -> str:
...
 class JobStatus:
+    """Complete status information for a job."""
+
+    job_id: JobId
+    workflow_id: WorkflowId
+    state: JobState
+    error_message: str | None = None
+    warning_message: str | None = None
+    start_time: int | None = None
+    end_time: int | None = None
```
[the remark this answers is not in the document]
SimonHeybrock: Not sure, those things maybe be more suited for the Grafana/Graylog integration.

src/beamlime/core/job.py
SimonHeybrock: The name is legacy. Renaming to `JobReply` and adding comments about the warning/error distinction.

src/beamlime/dashboard/job_service.py
SimonHeybrock: It should not matter, since currently `JobManager` is designed for single-threaded use.

src/beamlime/dashboard/widgets/job_status_widget.py
SimonHeybrock: Adding a comment.

PR #467, opened 2025-09-22
[Bifrost] Detector ratemeter

### LINKED ISSUE #426: [Bifrost] Detector ratemeter
> Purpose is to monitor the detector counts as a function of other parameters. However, it needs to be configurable, so that we are able to only look at the 5.0 meV bank for instance, or even specific analysers. This gives us the option of selectively make simple scans for alignment and calibration, using the elastic intensity, and visualize them quickly on a 1D data plot

Notes:
- [ ] Get reply from Rasmus about what the "other parameters" are.
- [ ] Figure out a nice way of defining the workflow parameters (and thus widgets) to allow for bank/analyzer selection. If we keep https://github.com/scipp/beamlime/blob/8b0cf64934c869e03b72ba8886dd05a434e58e8e/src/beamlime/config/instruments/bifrost.py#L85-L91 then we can simply select via some slicing operations, either via an int input, or a dropdown menu (which might be more readable, since we can label banks in a human-readable manner).
- [ ] Is some sort of TOF filtering required, or is the assumption that the total intensity is close to the elastic intensity (since almost everything is elastic scattering)?

SimonHeybrock:
Fixes #426, even though details of how/which detector subsets need to be configurable will need to be adjusted based on feedback.

For now detector rate is published by the `spectrum_view` workflow, since that was simple. We can then make a correlation histogram like this:

image

Future work:

- Support correlation hist with array-valued timeseries (to make a per-bank rate, for example).
- Consider if we should avoid publishing the full timeseries on every update. Should frontend accumulate? This question goes way beyond this particular application, but will be very important to address, or else long-running experiments will run into trouble.


src/ess/livedata/dashboard/scipp_to_holoviews.py
```diff
@@ -66,6 +66,28 @@ def convert_curve_1d(data: sc.DataArray) -> hv.Curve:
...
+    hv.ErrorBars
+        A Holoviews ErrorBars object.
+    """
+    data = _ensure_coords(data)
+    dim = data.dim
+    coord = data.coords[dim]
+    kdims = [coord_to_dimension(coord)]
+    vdims = [create_value_dimension(data)]
+
+    return hv.ErrorBars(
```
jl-wynen, line 84: Is it possible to add markers for the means? The plot you posted only shows the error bars, not the mean values.

src/ess/livedata/handlers/to_nxlog.py
```diff
@@ -42,9 +45,13 @@ def _at_capacity(self) -> bool:
     def _ensure_capacity(self, data) -> None:
```
jl-wynen, line 45:
```suggestion
    def _ensure_capacity(self, data: sc.Variable) -> None:
```
Would have helped me understand the change.

src/ess/livedata/handlers/to_nxlog.py
```diff
@@ -59,19 +66,16 @@ def _ensure_capacity(self, data) -> None:
             )
 
     def add(self, timestamp: int, data: LogData) -> None:
-        self._ensure_capacity(data.value)
+        self._ensure_capacity(data)
         self._timeseries.coords['time'].values[self._end] = data.time
         self._timeseries.data.values[self._end] = data.value
+        if data.variances is not None and self._timeseries.data.variances is not None:
+            self._timeseries.data.variances[self._end] = data.variances
```
jl-wynen, line 73: Is it a problem that we silently ignore variances of `data` here if the timeseries has no variances? Or that we silently do not add variances if `data` has none but the timeseries does?

src/ess/livedata/config/instruments/bifrost.py
```diff
@@ -232,8 +257,11 @@ def _spectrum_view(params: BifrostWorkflowParams) -> StreamProcessorWorkflow:
     return StreamProcessorWorkflow(
         wf,
         dynamic_keys={'unified_detector': NeXusData[NXdetector, SampleRun]},
-        target_keys=(SpectrumView,),
-        accumulators=(SpectrumView,),
+        target_keys=(SpectrumView, AnalyzerCounts, BankCounts, DetectorCounts),
```
jl-wynen, line 260: Do we need all three extra counts? Or is it wasteful to compute and stream all 3? 

You could stream only `AnalyzerCounts` and sum in the frontend.

src/ess/livedata/dashboard/scipp_to_holoviews.py
SimonHeybrock: As far as I can tell this is done by composing with, e.g., `hv.Curve`, but the styling (color) is different and I did not want to bother with that right now.

src/ess/livedata/handlers/to_nxlog.py
SimonHeybrock: Highly unlikely, since `self._timeseries` was made from previous calls to this method, i.e., the "data" are just new additions to the same thing, based on a single stream of messages. The double-check is actually pointless but satisfies linters.

src/ess/livedata/dashboard/correlation_histogram.py
```diff
@@ -370,13 +384,19 @@ def send(self, data: dict[ResultKey, sc.DataArray]) -> None:
...
+        self._normalize: bool = normalize
 
     def __call__(
         self, data: sc.DataArray, coords: dict[str, sc.DataArray]
     ) -> sc.DataArray:
         dependent = data.copy(deep=False)
+        if self._normalize:
+            times = dependent.coords['time']
+            widths = (times[1:] - times[:-1]).to(dtype='float64', unit='s')
+            widths = sc.concat([widths, widths.median()], dim='time')
```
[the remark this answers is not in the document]
jl-wynen: Then you can still use my suggestion and save one scan over the array.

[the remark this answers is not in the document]
SimonHeybrock: How so? If for some reason the last width was much smaller or much larger than the median we will get artifacts in the plot?

PR #468, opened 2025-09-22
Add basic password auth to the reduction dashboard

MridulS:
This adds some minimal auth in front of the panel dashboard. Running the dashboard with `LIVEDATA_BASIC_AUTH_COOKIE_SECRET` and `LIVEDATA_BASIC_AUTH_PASSWORD` env variable should set a password (and login with any random username).

It will give a screen like this
Screenshot 2025-09-22 at 14 41 15

Before adding ESS SSO I want to test this out with DST infra.

### DIFF
--- a/src/ess/livedata/dashboard/reduction.py
+++ b/src/ess/livedata/dashboard/reduction.py
@@ -16,13 +16,23 @@ hv.extension('bokeh')
 class ReductionApp(DashboardBase):
     """Reduction dashboard application."""
 
-    def __init__(self, *, instrument: str = 'dummy', dev: bool = False, log_level: int):
+    def __init__(
+        self,
+        *,
+        instrument: str = 'dummy',
+        dev: bool = False,
+        log_level: int,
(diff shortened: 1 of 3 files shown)


src/ess/livedata/dashboard/reduction.py
```diff
@@ -39,7 +49,9 @@ def create_main_content(self) -> pn.viewable.Viewable:
 
 
 def get_arg_parser() -> argparse.ArgumentParser:
-    return Service.setup_arg_parser(description='ESSlivedata Dashboard')
+    return Service.setup_arg_parser(
+        description='ESSlivedata Dashboard', dashboard_auth=True
+    )
```
SimonHeybrock, line 54: Since auth is only for the dashboard, it might make more sense to add the options here before returning, instead of adding a flag to the to the generic `Service`, which is used by the backend workers as well?

MridulS:
Making this draft for now. Let's wait to see how we can setup auth that makes DST happy too.

Issue #446, opened 2025-09-10
Configure ROI via rectangle/polygon selection on other plot


SimonHeybrock:
I have a proof-of-concept on how to do this with Holoviews/Bokeh (basically it is builtin, but was non-trivial to setup for some reason).

The main task is to understand how we can feed the rectangle/polygon vertices into the `WorkflowConfig` for a ROI workflow. The dashboard auto-generates input widgets from a Pydantic model. This would need to allow for the special case of selecting (one of) the drawn rectangles/polygons in a widget, and serializing it.

Notes from discussion with @jl-wynen related to #454:

- Stream polygon to Kafka.
- Select "blue polygon" as aux source name.
- Aux source name needs to be included in `WorkflowConfig` (requires schema change), similar to how we would need to include the correlation histogram axis timeseries stream names.

SimonHeybrock:
More detailed plan:

- Stream using `da00`. Holoviews streams polygons as a dict of lists: `{'xs': [[3.311561378272804, 5.40840904991906, 4.609610067831504], [5.551051234339809, 7.1771776353990715, 7.762012653522664, 6.093092909804335]], 'ys': [[7.055555804424694, 7.1523954525635505, 5.389908256880734], [4.460244834969896, 4.905708149667909, 2.8139652170537075, 1.942405584026185]], 'fill_color': ['red', 'green']}`. We can zip those, then each polygon can turn into `da00` (with two variables, for x and y).
- The source name must somehow reference the original source name (detector name *and* view name), as well as the polygon color. I don't think we need to job number, unless we can (or want to) couple the ROI to the detector view (in terms of resets, etc.).
- Moving the polygon would have to clear the histogram. This is not in line with the behavior of other workflows, but it is unavoidable, I think.

SimonHeybrock:
Made a start in https://github.com/scipp/esslivedata/tree/roi-stream.

SimonHeybrock:
Here is what Claude thinks:

Key Interdependency Analysis

  Both issues involve streaming auxiliary data that influences workflow execution:

  #497 - Timeseries refactoring

  - Moves accumulation from backend to frontend
  - Creates frontend-internal service for accumulating logs
  - Backend sends only latest values
  - Frontend maintains historical data for plotting

  #446 - Streamed ROI

  - ROI polygons become a timeseries/log stream (as you mentioned in the Oct 3 comment)
  - Backend needs to consume ROI stream from Kafka
  - Frontend publishes ROI changes to Kafka
[...]

PR #475, opened 2025-09-25
Allow config of aux-source-names in configuration modal

SimonHeybrock:
This serves several purposes:

1. Display to the user which aux sources are used (if there is only one choice). For example, with this change monitors or instrument angles that are specified in `aux_source_names` of `WorkflowSpec` will be visible in the workflow config modal. Currently there is just a single choice for each, but in the future this should allow for letting users select, e.g., which monitor to use for normalization.
2. Get rid of the "correlation histogram table widget" added recently. This was a third table (in addition to the job table, and result table), which was quite confusing (and never intended to be permanent). The one or two timeseries needed can now be selected as "aux source names". The param model updates dynamically.
3. Prep work for #446: With the addition here we should be able to implement something where we can select between rectangles/polygons drawn on the detector.


src/ess/livedata/dashboard/widgets/configuration_widget.py
```diff
@@ -106,16 +109,91 @@ def _create_source_selector(self) -> pn.widgets.MultiChoice:
...
+                )
+
+            selector = pn.widgets.Select(
+                name=category,
+                options=options,
+                value=options[0],  # Default to first entry
+                sizing_mode='stretch_width',
+                # No left margin and no right margin for last item: Align with source
+                # selector above.
+                margin=(0, 0 if i == size - 1 else 10, 10, 0),
```
jl-wynen, line 129: Not part of this PR. But scattering style information throughout the codebase can make it difficult to adjust the layout later. Is it possible to extract this into a CSS file?

src/ess/livedata/dashboard/widgets/correlation_histogram_widget.py
```diff
@@ -19,83 +18,49 @@ def __init__(
...
-    def _on_config_button_click(self, event: Any) -> None:
-        """Handle configuration button clicks."""
-        selection = self._tabulator.selection
-        indices = self._tabulator.value.iloc[selection].index
-        config = self._correlation_histogram_controller.create_config(
-            [self._result_key.iloc[i] for i in indices]
-        )
+        num_timeseries = len(timeseries_keys)
+        self._config_1d_button.disabled = num_timeseries < 1
+        self._config_2d_button.disabled = num_timeseries < 2
```
jl-wynen, line 48: Do these work when more than 1 or 2 time series are selected, respectively? The code in `correlation_histogram.py` looks like it uses the first 1 or first 2 coords. This could be confusing, can you change it to require exactly 1 or 2 coords?

src/ess/livedata/dashboard/correlation_histogram.py
```diff
@@ -86,63 +89,71 @@ class CorrelationHistogramConfigurationAdapter(
...
-    ) -> None:
-        self._data_keys = data_keys
-        self._axis_keys = axis_keys
+    def __init__(self, controller: CorrelationHistogramController) -> None:
         self._controller = controller
+        self._selected_aux_sources: dict[str, str] | None = None
+        # All timeseries except the axis keys can be used as dependent variables. These
+        # will thus be shown by the widget in the "source selection" menu.
+        timeseries = self._controller.get_timeseries()
+        # The to make unique names for display. The might not be very readable, for
```
jl-wynen, line 98: 2 times 'the' without a noun. What should this say?

src/ess/livedata/dashboard/widgets/correlation_histogram_widget.py
SimonHeybrock: This code is about enabling the buttons: If there are 5 timeseries then both plots are possible. The user can then choose between those in 1 or 2 dropdown menus.

src/ess/livedata/dashboard/correlation_histogram.py
SimonHeybrock:
```suggestion
        # Try to make unique names for display. This might not be very readable, for
```

jl-wynen (approved):

PR #484, opened 2025-09-30
Dashboard workflow code cleanup

SimonHeybrock:
A bunch of cleanup. Some wrappers did not make sense any more after we introduced the generic `ConfigurationAdapter` mechanism (but it was overlooked that this could be cleaned up). See individual commits.

The motivation for this cleanup is that I would like to merge to bespoke "correlation histogram widget" with the main workflow selection dropdown. I hope this will be easier after this.


SimonHeybrock:
I'll merge this since the cleanup mainly moved or removed redundant code, the other changes were reviewed as part of #485.

PR #492, opened 2025-10-02
[Odin] Detector view with working ROI

SimonHeybrock:
- Use projection instead of logical view so we can get ROI to work.
- Bypass issue of detector being placed at the sample by hacking the file.
- Bypass issue from detector dims not initializing properly (ScippNexus cannot deduce what X and Y are, I think). Note that the axes may be swapped, I have no idea.
- Add cumulative ROI.

@MridulS @nvaytet Do you want to try


MridulS:
Ran this on the ODIN testing machines, seems to look fine. There was no data incoming so can't say anything about the actual processing but able to spin up a Job and create a plot.

MridulS (approved):

PR #510, opened 2025-10-15
Add detector plotter with ROI

### LINKED ISSUE #446: Configure ROI via rectangle/polygon selection on other plot

SimonHeybrock:
This brings back support for ROIs, but this time properly:

- User can add, move, and delete.
- Multiple ROIs supported.
- While not supported yet, we should be able to add support for, e.g., polygon ROIs without many changes (but this also lacks ROI filter support from `essreduce`).
- My plan is to extend this further, adding a "band" tool to the ROI plot which will act as a stream to select a TOA range for the detector plot. The mechanism added here should work for this without conceptual changes.

Fixes #446.

Follow-ups and known issues, see #511.

Plot config dialog:
image

Detector with ROI plot:
image

### DIFF
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -123,6 +123,7 @@ class Instrument:
         description: str = '',
         source_names: Sequence[str] | None = None,
         aux_sources: type[Any] | None = None,
+        outputs: type[Any] | None = None,
     ) -> Callable[[Callable[..., Workflow]], Callable[..., Workflow]]:
         """
         Decorator to register a factory function for creating StreamProcessors.
@@ -157,6 +158,11 @@ class Instrument:
             this will be used for validation and UI generation. The auxiliary source
             configuration is handled by the Job layer and is not passed to the workflow
             factory function.
+        outputs:
+            Optional Pydantic model class defining workflow outputs with metadata.
+            Field names should be simplified identifiers (e.g., 'i_of_d_two_theta')
+            that match keys returned by workflow.finalize(). Field metadata (title,
+            description) provides human-readable information for the UI.
 
         Returns
         -------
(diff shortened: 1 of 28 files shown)


jokasimr:
I've only just scanned the changes here. One quick question: Why is the maximum number of roi:s pre-defined by a setting? Seems it would be simpler to not clutter the user interface with that option. And might be simpler to not have to enforce it in the backend.

SimonHeybrock:
> I've only just scanned the changes here. One quick question: Why is the maximum number of roi:s pre-defined by a setting? Seems it would be simpler to not clutter the user interface with that option. And might be simpler to not have to enforce it in the backend.

- There has to be some maximum (because we have to setup subscriptions to DataService, limit compute and memory use, etc.).
- When you draw ROIs and you already have to maximum, the oldest ROI disappears. So if you only ever want one, and the draw new ones you can set this to 1. Other users might want multiple.
- You can run into cases were you drew a ROI with, e.g., 0 width. It can be tricky to remove that (because you will have trouble selecting it with a mouse click). Having a small maximum you can just draw some new ROIs and eventually the "bad" one will disappear.

jokasimr:
We could consider just having a fixed limit to the number of ROIs, for example 5. If the user only ever wants to see one ROI they can just delete the previous one before creating a new one.

SimonHeybrock:
> We could consider just having a fixed limit to the number of ROIs, for example 5. If the user only ever wants to see one ROI they can just delete the previous one before creating a new one.

We could, let us see if we get user complaints.

SimonHeybrock:
I'll merge this as it is holding up other work. Review still highly welcome!

Issue #515, opened 2025-10-20
Frontend UX


SimonHeybrock:
Current idea:
- Output-centric UI. Job (or even workflow) lists should not be center of attention.
- Need to combine multiple outputs. May need mechanisms to (1) dynamically layout or overlay plots and (2) define "templates" for plots. One early example of the latter may be the ROI plot, where we have a fixed `Image + Curve(s)` layout. In practice we might want to be a lot more specific, e.g., to define a layout for the 2D detector views.
- Do we need "synchronized" workflows? For example, all detector projections, or data reduction for diffraction vs. SANS banks?

Related: #430

PR #516, opened 2025-10-20
Workflow output metadata

### LINKED ISSUE #430: Title and description of reduction workflow outputs
### Current situation

Target keys of a Sciline workflow such as `ess.powder.types.FocussedDataDspacingTwoTheta[ess.reduce.nexus.types.SampleRun]` are formatted as strings and displayed in the UI. This is bad UX.

### Suggested solution

- Define a `title` and `description` for each output, which can be used for a brief and long description in the UI.
- Consider if this can be nicely done with Pydantic field `title` and `descriptions`, which we already use for workflow inputs.
- Consider if this eventually move into technique packages, since they would also benefit (note: similar to how I would like to considering moving input parameter models to upstream).

SimonHeybrock:
Fixes #430. This is a preparation for #515, but until then the output list on the dashboard becomes more readable, since it now includes the output title and description (but this widget is temporary and will be replaced, just to demonstrate that we have what we need!):

Screenshot 2025-10-20 at 11 47 40

This should **should** also improve plot labels, but our da00 converter does not support `scipp.DataArray.name` (it is dropped). This was fixed as part of #510, but as it is not essential right now it won't bother with extrating it from there.

### DIFF
--- a/src/ess/livedata/config/workflow_spec.py
+++ b/src/ess/livedata/config/workflow_spec.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import time
 import uuid
+from collections import defaultdict
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any, TypeVar
@@ -160,19 +161,9 @@ class WorkflowSpec(BaseModel):
         if outputs is None:
             return outputs
 
-        titles = []
+        title_counts: dict[str, list[str]] = defaultdict(list)
         for field_name, field_info in outputs.model_fields.items():
-            title = field_info.title
-            if title is None:
-                # If no title is specified, use the field name
-                title = field_name
-            titles.append((title, field_name))
-
-        # Check for duplicate titles
-        title_counts: dict[str, list[str]] = {}
-        for title, field_name in titles:
-            if title not in title_counts:
-                title_counts[title] = []
+            title = field_info.title if field_info.title is not None else field_name
             title_counts[title].append(field_name)
 
         duplicates = {


src/ess/livedata/dashboard/widgets/plot_creation_widget.py
SimonHeybrock: This widget is legacy, I think details can safely be ignored.

src/ess/livedata/core/job_manager.py
```diff
@@ -67,6 +74,53 @@ class JobFactory:
...
+            return result
+
+        # Get workflow spec to access output metadata
+        workflow_spec = self.get_workflow_spec(result.workflow_id)
+        if workflow_spec is None or workflow_spec.outputs is None:
+            return result
+
+        def set_name_from_metadata(value, output_name: str) -> None:
+            """Set human-readable name on a value based on output metadata."""
+            if hasattr(value, 'name'):
```
jl-wynen: Would it make more sense to check for `DataArray` explicitly? What if `value` is some other type that happens to have a `name` attribute?

SimonHeybrock:
> I think this mechanism is fine. But did you consider adding these metadata fields to the domain types directly? That could allow us ot improve output in Sciline as well.
> 
> I'm not sure what would be the best approach. Wrapping the values in an extra pydantic model or dataclass would make them annoying to use:

Yes... but maybe it would be the right thing to do? One level above bare Sciline workflows, with Pydantic for both all workflow inputs and all workflow outputs?

src/ess/livedata/core/job_manager.py
```diff
@@ -67,6 +74,53 @@ class JobFactory:
...
+                field_info = workflow_spec.outputs.model_fields.get(output_name)
+                if field_info is not None and field_info.title is not None:
+                    value.name = field_info.title
+
+        # If data is a DataGroup, enrich each DataArray
+        if isinstance(result.data, sc.DataGroup):
+            for output_name, value in result.data.items():
+                set_name_from_metadata(value, output_name)
+        # If data is a single DataArray and we have a single output
+        elif hasattr(result.data, 'name') and result.output_name is not None:
```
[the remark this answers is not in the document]
SimonHeybrock: Because the fact that we have a datagroup indicates that we have multiple outputs, i.e., they *must* have an output_name by definition.

jl-wynen:
> > I think this mechanism is fine. But did you consider adding these metadata fields to the domain types directly? That could allow us ot improve output in Sciline as well.
> > I'm not sure what would be the best approach. Wrapping the values in an extra pydantic model or dataclass would make them annoying to use:
> 
> Yes... but maybe it would be the right thing to do? One level above bare Sciline workflows, with Pydantic for both all workflow inputs and all workflow outputs?

Not sure I understand. Why a level above sciline? I meant to change the domain types in the ess* packages which would then be directly coupled to the workflows.

Issue #445, opened 2025-09-10
Cleanup/fix reply mechanism to request from frontend


SimonHeybrock:
There is still some legacy `WorkflowStatus` code floating about, as "replies" to `WorkflowConfig` messages (which trigger workflow starts). This are currently not handled by the frontend any more (which relies on periodic `JobStatus` since #440).

https://gitlab.esss.lu.se/ecdc/interface-specs/-/blob/477da77510561f1a6afee874f7c8bfe907183723/nicos-beamlime.md#command-acknowledgement describes a mechanism that would be understood by NICOS. We should consider implementing it this way, then start using it on the frontend. Here is the content of that file (note BeamLime references there is ESSlivedata, i.e., this project) - also note this is not the latest document version, but some things are presented better here:

# BeamLime <-> NICOS Interface

## Introduction

BeamLime will be performing all the data reduction and NICOS will be responsible for displaying the data 
and controlling the BeamLime services. All communication between BeamLime and NICOS will be done through Kafka.

### BeamLime

BeamLime consists of multiple services each providing a different plot of reduced data. 
The services are responsible for collecting data from the detectors and monitors, reducing it and sending it to NICOS 
or other clients for display.
Each type of plot, even if coming from the same device, could be a different service.

### NICOS

NICOS is a control system software that provides the user interface for the instruments at the ESS.
NICOS can display plots from different services and devices. In this case, NICOS will display the plots from BeamLime.
There is a central device on the NICOS server that will receive the data from BeamLime and distribute it to follower devices.
Configuration commands will be sent from the follower devices through the central device to the BeamLime server via Kafka.

## Source Names

The source names should be unique per instrument and follow the format `device_name:signal_name`.

For example:
- `endcap_backwards_detector:accumulated_counts`
- `cave_monitor_1:sliding_window_counts`

Since the `device_name:signal_name` can be parsed by NICOS, we can use it to send commands to the device.

## Command JSON

The command JSON is a dictionary with the following mandatory keys:
- `command`: The command to be executed, `config`, `start`, `stop`, `clear`
- `message_id`: A unique identifier for the message
- `device`: The device to be configured, for example `endcap_backwards_detector:accumulated_counts`, `cave_monitor_1:sliding_window_counts`. 
This should match the data source name.

### Table of valid config keys

TODO: Add the correct keys that are expected by BeamLime. Below

| Key | Type | Description                                          |
| --- | --- |------------------------------------------------------|
[...]

PR #521, opened 2025-10-31
Split topic and add ConfigStore

SimonHeybrock:
Preparation for #445, but mainly for support of HTTP instead of Kafka.

For now we only have an in-memory config store, i.e., config is not persisted across restarts. The plan is to either implement sth. file-based, or to bring back the Kafka mechanism with a distinct topic.

### DIFF
--- a/src/ess/livedata/dashboard/config_store.py
+++ b/src/ess/livedata/dashboard/config_store.py
@@ -11,6 +11,7 @@ ConfigStore is for persistent storage (e.g., files, local storage) while
 ConfigService is for transient runtime coordination.
 """
 
+from collections import UserDict
 from collections.abc import MutableMapping
 from typing import Any
 
@@ -20,7 +21,7 @@ from ess.livedata.config.workflow_spec import WorkflowId
 ConfigStore = MutableMapping[WorkflowId, dict[str, Any]]
 
 
-class InMemoryConfigStore(MutableMapping[WorkflowId, dict[str, Any]]):
+class InMemoryConfigStore(UserDict[WorkflowId, dict[str, Any]]):
     """
     In-memory implementation of ConfigStore with optional LRU eviction.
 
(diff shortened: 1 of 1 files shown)


tests/dashboard/workflow_controller_test.py
```diff
@@ -149,34 +152,36 @@ def test_start_workflow_sends_config_to_sources(
...
 
-        # Assert
-        persistent_configs = service.get_persistent_configs()
-        assert workflow_id in persistent_configs.configs
-
-        workflow_config = persistent_configs.configs[workflow_id]
-        assert workflow_config.source_names == source_names
-        assert workflow_config.config.identifier == workflow_id
-        assert workflow_config.config.params == {"threshold": 200.0, "mode": "fast"}
+        # Assert - check ConfigStore instead of service
```
SimonHeybrock:
```suggestion
        # Assert
```

src/ess/livedata/dashboard/config_store.py
```diff
@@ -0,0 +1,85 @@
...
+    configurations for multiple use cases with different cleanup strategies:
+
+    - Workflow configs: Could theoretically be cleaned up by removing non-existent
+      workflows, but this requires tracking workflow registry state.
+    - Plotter configs: More complex - for each workflow there can be multiple outputs,
+      and for each output multiple applicable plotters, each needing its own config.
+      Tracking existence is impractical.
+
+    LRU eviction provides a simple, uniform policy that works for both cases and can
+    be configured per store instance based on expected usage patterns.
```
jokasimr: This might be a simpler implementation that lets us override fewer methods.

```python
from collections import OrderedDict

class LimitedLengthDict(OrderedDict):
    def __init__(self, max_length):
        self.max_length = max_length
    def __setitem__(self, *args, **kwargs):
        while len(self) > self.max_length - 1:
            self.popitem(last=False)
        super().__setitem__(*args, **kwargs)
```

src/ess/livedata/dashboard/configuration_adapter.py
```diff
@@ -40,9 +83,20 @@ def initial_aux_source_names(self) -> dict[str, str]:
...
-        return {}
+        if not self._config_state:
+            return {}
+        if not self.aux_sources:
+            return {}
+        # Filter to only include valid field names
+        valid_fields = set(self.aux_sources.model_fields.keys())
+        return {
+            k: v
+            for k, v in self._config_state.aux_source_names.items()
```
jokasimr: Why might there be invalid names in `aux_source_names`?

src/ess/livedata/dashboard/config_store.py
SimonHeybrock: `OrderedDict` seems not necessary since we use Python>3.7, but I pushed an update using `UserDict`.

src/ess/livedata/dashboard/configuration_adapter.py
```diff
@@ -81,14 +135,34 @@ def source_names(self) -> list[str]:
...
     @property
-    @abstractmethod
     def initial_source_names(self) -> list[str]:
-        """Initially selected source names."""
+        """
+        Initially selected source names.
+
+        Default implementation filters persisted source names to only include
+        currently available sources. If no valid persisted sources remain,
+        defaults to all available sources.
```
jokasimr: I can understand we default to all names when there is no config object, but why also return all names when there is a config object that has an empty `source_names`?

src/ess/livedata/dashboard/config_store.py
jokasimr: Regular dict does not have the `last` argument in `popitem`.

src/ess/livedata/dashboard/configuration_adapter.py
```diff
@@ -5,13 +5,56 @@
...
+    """
+
+    source_names: list[str] = Field(
+        default_factory=list,
+        description="Selected source names for this workflow or plotter",
+    )
+    aux_source_names: dict[str, str] = Field(
+        default_factory=dict,
+        description=(
+            "Selected auxiliary source names as field name to stream name mapping"
```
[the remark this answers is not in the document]
SimonHeybrock: Source is the "main" data stream such as the detector. Aux sources are typically every other stream.

src/ess/livedata/dashboard/configuration_adapter.py
SimonHeybrock: Because `config_state` is typically restored from a config file and may thus be outdated.

PR #522, opened 2025-10-31
Unify Kafka interface between frontend and backend

SimonHeybrock:
## Context

Dashboard was using custom Kafka abstractions (ConfigService, MessageBridge, custom transport layers) that duplicated backend patterns and created unnecessary complexity. These mechanisms where added when we had widgets that directly controlled running workflows via continues "config" changes. Furthermore, some of it was used for loopback and consistency mechanisms. #521 removes one need for this (albeit maybe only temporarily), but low-latency widget linking between dashboards is no longer considered a requirement.

The main driver for this change is the addition of HTTP support as an alternative to Kafka: I have this working in the backend, but the bespoke code in dashboard would have required a lot more duplication.

## What Changed

- Replaced generic ConfigService with focused services using backend
  MessageSource/MessageSink directly
- WorkflowConfigService and CommandService: Serves as a layer separating the controllers from the transport. Added since it hopefully lets us swap in HTTP transport, but I expect small refactors here.
- Removed 4 abstraction layers: ControllerFactory, MessageBridge, custom transport,
  throttling/deduplication handlers
- Add batch interfaces in some controllers since we no longer have bespoke logic in the transport layer to deal, e.g.., with a "reset" command for all jobs. Naively this would have sent and flushed many individual messages.

## Why

  - Alignment: Dashboard now uses same message patterns as backend services
  - Interface segregation: Controllers depend only on what they need (WorkflowController
   → WorkflowConfigService protocol, JobController → single publish method)
  - Simplicity: Removes "god object" ConfigService and unnecessary indirection

### DIFF
--- a/scripts/README_screenshots.md
+++ b/scripts/README_screenshots.md
@@ -1,0 +1,122 @@
+# Widget Screenshot Generator
+
+## Overview
+
+The `generate_widget_screenshots.py` script generates PNG screenshots of Panel ConfigurationWidget instances for all workflow specifications in a given instrument.
+
+These screenshots are useful for:
+- Documentation
+- Stakeholder presentations
+- Visual verification of widget layouts
+- Comparing configurations across instruments
+
+## Requirements
+
+The script requires:
+- `playwright` Python package (already in dev dependencies)
+- Chromium browser for Playwright: `python -m playwright install chromium`
+- System dependencies for Chromium (already in devcontainer)
+
+## Usage
+
+### Basic Usage
+
+Generate screenshots for all workflows in the dummy instrument:
+
+```bash
+python scripts/generate_widget_screenshots.py dummy
+```
+
+This creates screenshots in `screenshots/dummy/` directory.
+
+### Custom Output Directory
+
+```bash
+python scripts/generate_widget_screenshots.py dream --output-dir docs/screenshots
+```
+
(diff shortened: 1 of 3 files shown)


src/ess/livedata/dashboard/dashboard.py
```diff
@@ -184,49 +143,55 @@ def _setup_data_infrastructure(self, instrument: str, dev: bool) -> None:
...
+            for kind in [
+                StreamKind.LIVEDATA_DATA,
+                StreamKind.LIVEDATA_STATUS,
+                StreamKind.LIVEDATA_RESPONSES,
+            ]
+        ]
         consumer = self._exit_stack.enter_context(
             kafka_consumer.make_consumer_from_config(
-                topics=[data_topic, status_topic],
+                topics=topics,
```
jokasimr: Looks like the 'responses' topic is new. Where was that communication handled before?

jokasimr (commented):
Nice diff!

I've only reviewed the code that was added, I've not read the removed files.

Overall this looks reasonable to me. I've left a couple of questions to make sure I understand the new system.

Is there any functionality that is removed by these changes?
I think you touched on it in the PR description, but it was not entirely clear to me.

jokasimr (approved):

PR #524, opened 2025-11-05
Implement file-backed config store for dashboard persistence

SimonHeybrock:
Replace InMemoryConfigStore with FileBackedConfigStore to persist dashboard UI state (workflow and plotter configurations) across sessions. ConfigStoreManager puts it all together.

This brings back the ability to restore UI state (workflow and plotter configs) after dashboard restarts. This was removed in #521 since it used a Kafka topic for multiple unrelated purposes. File-based seems simpler now and has the advantage that it carries across Kafka topic wipes, which happen quite frequently during local development.


src/ess/livedata/dashboard/config_store.py
```diff
@@ -68,3 +76,262 @@ def _evict_oldest(self) -> None:
...
+        super().__delitem__(key)
+        self._save_to_file()
+
+    def _evict_oldest(self) -> None:
+        """Remove oldest configs based on cleanup fraction."""
+        num_to_remove = max(1, int(len(self.data) * self._cleanup_fraction))
+        # Dict maintains insertion order in Python 3.7+
+        oldest_keys = list(self.data.keys())[:num_to_remove]
+        for key in oldest_keys:
+            del self.data[key]
```
jl-wynen: These operations are not atomic. Is there a risk that multiple threads will access the config at the same time?

src/ess/livedata/dashboard/config_store.py
```diff
@@ -68,3 +76,262 @@ def _evict_oldest(self) -> None:
...
+                        serialized_data,
+                        f,
+                        default_flow_style=False,
+                        sort_keys=False,  # Preserve insertion order for LRU
+                    )
+                finally:
+                    fcntl.flock(f.fileno(), fcntl.LOCK_UN)
+
+            # Atomic rename
+            temp_path.replace(self._file_path)
```
jl-wynen: This happens after the lock has been released. So it is no synced with access to the persistent file in `_load_from_file`. (File locks are advisory, the OS will not enforce them.) And it is also not synced with access to the temporary file if multiple writes happen concurrently.

src/ess/livedata/dashboard/config_store.py
```diff
@@ -68,3 +76,262 @@ def _evict_oldest(self) -> None:
...
+    Parameters
+    ----------
+    file_path:
+        Path to the YAML file for storing configurations. Parent directory
+        will be created if it doesn't exist.
+    max_configs:
+        Maximum number of configurations to keep. When exceeded, oldest
+        configurations are automatically evicted. If None, no limit is enforced.
+    cleanup_fraction:
+        Fraction of configs to remove when limit exceeded (default 0.2 = 20%).
```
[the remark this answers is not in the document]
SimonHeybrock: To avoid running the cleanup every single time?

src/ess/livedata/dashboard/config_store.py
```diff
@@ -68,3 +99,280 @@ def _evict_oldest(self) -> None:
...
+            logger.warning("Starting with empty config store")
+
+    def _save_to_file(self) -> None:
+        """Save configurations to YAML file using atomic write operation."""
+        try:
+            # Serialize WorkflowId keys to strings for YAML
+            serialized_data = {str(key): value for key, value in self.data.items()}
+
+            # Write atomically: write to temp file, then rename
+            temp_path = self._file_path.with_suffix('.tmp')
```
jl-wynen: This uses the same temp file for a store across threads. So make sure that the lock is held when calling this function! (It is currently, but code changes.) I don't know if `threading.Lock` has a way to prove that you are holding it. So unless you want to build such a system, you may have to just be very careful.

src/ess/livedata/dashboard/config_store.py
```diff
@@ -68,3 +99,280 @@ def _evict_oldest(self) -> None:
...
+                    return
+
+                # Deserialize WorkflowId keys from strings
+                for key_str, value in raw_data.items():
+                    try:
+                        workflow_id = WorkflowId.from_string(key_str)
+                        self.data[workflow_id] = value
+                    except (ValueError, KeyError) as e:
+                        logger.warning(
+                            "Skipping invalid config entry '%s': %s", key_str, e
```
jl-wynen: So the consuming code can handle a partial config where some keys are missing?

src/ess/livedata/dashboard/config_store.py
```diff
@@ -3,27 +3,55 @@
...
+import logging
+import os
+import threading
 from collections import UserDict
 from collections.abc import MutableMapping
-from typing import Any
+from pathlib import Path
+from typing import Any, Literal
+
+import platformdirs
```
jl-wynen: Please add it as an explicit requirement.

src/ess/livedata/dashboard/dashboard.py
```diff
@@ -66,15 +66,11 @@ def __init__(
...
-            max_configs=100, cleanup_fraction=0.2
-        )
-        self._plotter_config_store = InMemoryConfigStore(
-            max_configs=100, cleanup_fraction=0.2
-        )
+        # Configs are persisted to disk for cross-session state preservation.
+        # ConfigStoreManager centralizes config directory setup and store creation.
+        self._config_manager = ConfigStoreManager(instrument=instrument)
+        self._workflow_config_store = self._config_manager.get_store('workflow_configs')
+        self._plotter_config_store = self._config_manager.get_store('plotter_configs')
```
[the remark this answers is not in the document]
jl-wynen: So it's not a singleton. It's just one set of stores per consumer?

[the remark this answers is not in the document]
SimonHeybrock: There is nothing that "enforces" `DashboardBase` being a singleton. By "consumer" you mean...? Let us have a short chat to clarify!

Related to my `gunicorn` concerns yesterday, I think we can consider them null, see #528 (planning to remove `gunicorn`).

PR #526, opened 2025-11-05
Integration test framework

SimonHeybrock:
See title. This will enable automatic a ton of manual testing. For now I only add two basic example tests to demonstrate it works.

**Note: Currently you need to start Kafka manually. The test setup is deliberately not doing this (and this does not run in CI). The medium-term goal is to enable HTTP support for this purpose.**

### DIFF
--- a/src/ess/livedata/dashboard/plot_configuration_adapter.py
+++ b/src/ess/livedata/dashboard/plot_configuration_adapter.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+from __future__ import annotations
+
+from typing import Any
+
+import pydantic
+
+from ess.livedata.config.workflow_spec import JobNumber
+from ess.livedata.dashboard.configuration_adapter import ConfigurationAdapter
+from ess.livedata.dashboard.plotting import PlotterSpec
+from ess.livedata.dashboard.plotting_controller import PlottingController
+
+
+class PlotConfigurationAdapter(ConfigurationAdapter):
+    """Adapter for plot configuration modal."""
+
+    def __init__(
+        self,
+        job_number: JobNumber,
+        output_name: str | None,
+        plot_spec: PlotterSpec,
+        available_sources: list[str],
+        plotting_controller: PlottingController,
+        success_callback,
+    ):
+        config_state = plotting_controller.get_persistent_plotter_config(
+            job_number=job_number,
+            output_name=output_name,
+            plot_name=plot_spec.name,
+        )
+        super().__init__(config_state=config_state)
+        self._job_number = job_number
+        self._output_name = output_name
+        self._plot_spec = plot_spec
+        self._available_sources = available_sources
+        self._plotting_controller = plotting_controller
+        self._success_callback = success_callback
+
+    @property
+    def title(self) -> str:
+        return f"Configure {self._plot_spec.title}"
+
+    @property
+    def description(self) -> str:
+        return self._plot_spec.description
+
+    def model_class(self) -> type[pydantic.BaseModel] | None:
+        return self._plot_spec.params
+
+    @property
+    def source_names(self) -> list[str]:
+        return self._available_sources
+
+    def start_action(
+        self,
+        selected_sources: list[str],
+        parameter_values: Any,
+    ) -> None:
+        """Create the plot and call the success callback with the result."""
+        plot = self._plotting_controller.create_plot(
+            job_number=self._job_number,
+            source_names=selected_sources,
+            output_name=self._output_name,
+            plot_name=self._plot_spec.name,
+            params=parameter_values,
+        )
+        self._success_callback(plot, selected_sources)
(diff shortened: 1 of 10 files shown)


tests/integration/__init__.py
jl-wynen: I thought we decided to not make `tests` a package?

Since you have several support modules here, how about turning those into a package that gets installed by tox when running tests?

tests/integration/README.md
```diff
@@ -0,0 +1,172 @@
...
+    workflow_id = WorkflowId(
+        instrument='dummy',
+        namespace='monitor_data',
+        name='monitor_histogram',
+        version=1,
+    )
+
+    backend.workflow_controller.start_workflow(workflow_id, ['monitor1'], config)
+
+    # ✅ GOOD: Filter by workflow_id before checking
```
jl-wynen: If this is a good pattern, it should be provided by the framework. We shouldn't duplicate this complex code across tests.

tests/integration/service_process.py
```diff
@@ -0,0 +1,316 @@
...
+                raise RuntimeError(
+                    f"Service {self.service_module} failed to start.\n"
+                    f"Exit code: {self.process.returncode}\n"
+                    f"STDOUT: {stdout}\n"
+                    f"STDERR: {stderr}"
+                )
+
+            # Check if we've seen "Service started" in output
+            combined_output = ''.join(self._stdout_lines + self._stderr_lines)
+            if 'Service started' in combined_output:
```
jl-wynen: This seems brittle. Is it part of the contract for services to output this message? Do services have a heartbeat or some kind of alive 'endpoint'?

tests/integration/service_process.py
```diff
@@ -0,0 +1,316 @@
...
+
+        Parameters
+        ----------
+        timeout:
+            Maximum time to wait for each service to stop (seconds)
+        """
+        # Stop in reverse order
+        for name, service in reversed(list(self.services.items())):
+            logger.info("Stopping service group member: %s", name)
+            service.stop(timeout=timeout)
```
jl-wynen: If any service raises during stop or start, other services might not get stopped. I think you should use `ExitStack` or something similar here.

tests/integration/test_workflow_lifecycle.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    backend = integration_env.backend
+
+    # Define workflow parameters for monitor histogram workflow
+    workflow_id = WorkflowId(
+        instrument='dummy',
+        namespace='monitor_data',
+        name='monitor_histogram',
+        version=1,
+    )
```
jl-wynen: (Based on guidelines on test separation. This is not actually relevant with function-scope fixtures.) Both tests use the same id. As far as I can tell, workflows are not removed at the end of a test. Also, do you forbid using pytest-xdist?

tests/integration/__init__.py
SimonHeybrock: Right, everything should go into a subfolder! No need for package and tox I think?

tests/integration/service_process.py
```diff
@@ -0,0 +1,316 @@
...
+            # Convert Python naming to CLI flags (underscore to hyphen)
+            flag_name = key.replace('_', '-')
+
+            if isinstance(value, bool):
+                # Boolean flags: only add if True
+                if value:
+                    args.append(f'--{flag_name}')
+            else:
+                # Regular arguments
+                args.extend([f'--{flag_name}', str(value)])
```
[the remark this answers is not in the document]
jl-wynen: Yes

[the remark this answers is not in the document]
SimonHeybrock: Claude still disagrees:
```
No, shlex should not be used here. The current implementation
  is correct and safe.

  Here's why:

  The current code is safe because it uses a list, not shell=True

  In _build_command_args() (lines 102-115), arguments are built
  as a list:

  args.extend([f'--{flag_name}', str(value)])

  And passed to subprocess.Popen as a list (line 206):

  self.process = subprocess.Popen(  # noqa: S603
      args,  # <-- list, not string
      ...
  )

  Why shlex is NOT needed

  1. No shell interpretation: When subprocess.Popen receives a
  list (and shell=False, which is the default), Python passes
  arguments directly to the OS without invoking a shell. There's
  no shell parsing, so no escaping is needed.
  2. shlex is for shell=True: shlex.quote() is only needed when:
    - Using shell=True
    - Constructing a shell command string manually

  Since neither applies here, it's unnecessary.
  3. Already security-safe: The current approach is actually MORE
   secure than shell=True + shlex.quote() because it avoids shell
   interpretation entirely, preventing command injection attacks.

  What about special characters?

  Even if value contains spaces or special characters, the
  list-based approach handles them correctly:

  # Example: value = "hello world"
  args.extend(['--instrument', 'hello world'])
  # This correctly passes 'hello world' as a SINGLE argument to 
  the subprocess

  The current implementation at service_process.py:113 is correct
   and secure as-is.
```

SimonHeybrock:
> Almost there, just the `shlex` comment left.

See above. I also refactored the helper + tests since it occurred to be that a job-based interface makes much more sense than workflow-based. I think they are also easier to understand now.

PR #535, opened 2025-11-13
Add integration tests for workflow config persistence and incompatible config handling

### LINKED ISSUE #525: Ensure incompatible config does not prevent ConfigurationWidget from operating
Context:
- ConfigurationWidget uses ConfigurationAdapter to fill in defaults.
- ConfigurationAdapter uses stored config (see #524 for example) to validate a Pydantic model.
- If workflow params change in an incompatible way we should increment the workflow version to avoid problems.
- But that may not reflect reality, i.e., we may have incompatible config that cannot be parsed any more.

Solution:
- Ensure we catch exceptions from Pydantic model validation on config (use workflow defaults instead, as if there was no stored config).
- Consider if we can or should clear the bad config, or just rely on it being overwritten?

SimonHeybrock:
## Summary

This PR adds integration tests for workflow configuration persistence via the config store and enhances the `ConfigurationAdapter` to gracefully handle incompatible stored configurations.

Fixes #525 - this worked before, but now we do validation in the adapter instead of relying on the UI components.

### 1. Enhanced `DashboardBackend` to support configurable transport
- Added `transport` parameter accepting `'kafka'` or `'none'`
- Allows testing without Kafka when it's not needed
- Uses `NullTransport` for tests that only need the controller/config store

### 2. Added integration tests for workflow config persistence
**File:** `tests/integration/config_persistence_test.py`

### 3. Enhanced `ConfigurationAdapter` to detect incompatible params
**File:** `src/ess/livedata/dashboard/configuration_adapter.py`

Modified `initial_parameter_values` to:
- Check for field overlap between stored params and current model fields
- If no overlap detected (complete incompatibility) → return `{}` to trigger defaults
- If partial/full overlap → return stored params (Pydantic handles missing fields)

This ensures schema changes between workflow versions don't break the dashboard - it gracefully falls back to defaults when configs are completely incompatible.

## Motivation

These tests verify the flow where:
1. Starting a workflow via `WorkflowController` stores params in the config store
2. Recreating the adapter via the controller reads those params back correctly
3. Incompatible configs (from old workflow versions) don't break the dashboard

### DIFF
--- a/tests/integration/config_persistence_test.py
+++ b/tests/integration/config_persistence_test.py
@@ -97,7 +97,8 @@ def test_adapter_filters_removed_sources(
         version=1,
     )
 
-    # Start workflow with multiple sources (monitor3 not in monitor_histogram spec for dummy)
+    # Start workflow with multiple sources (monitor3 not in monitor_histogram spec
+    # for dummy)
     source_names = ['monitor1', 'monitor2', 'monitor3']
     backend_with_null_transport.workflow_controller.start_workflow(
         workflow_id=workflow_id,


tests/integration/backend.py
```diff
@@ -44,6 +49,7 @@ def __init__(
         instrument: str = 'dummy',
         dev: bool = True,
         log_level: int | str = logging.INFO,
+        transport: Literal['kafka', 'none'] = 'kafka',
```
nvaytet: Consider using
```suggestion
        transport: Literal['kafka', None] = 'kafka',
```
?

tests/integration/config_persistence_test.py
```diff
@@ -0,0 +1,222 @@
...
+    # Create adapter - it should filter to only sources available in spec
+    adapter = backend_with_null_transport.workflow_controller.create_workflow_adapter(
+        workflow_id
+    )
+
+    # Adapter should only return sources that are both persisted AND in the spec
+    initial_sources = adapter.initial_source_names
+    assert 'monitor1' in initial_sources
+    assert 'monitor2' in initial_sources
+    # monitor3 should be filtered out as it's not in the workflow spec
```
nvaytet: Can you explain where in the test monitor3 gets dropped? I don't see it.

Also, we should probably test that it really does get dropped, something like
`assert 'monitor3' not in initial_sources`

tests/integration/backend.py
SimonHeybrock: This is actually deliberate and good I think: `None` is often interpreted as "use default", but here `'none'` refers to a particular kind of transport.

nvaytet: Then how about calling it `'null'`?

tests/integration/config_persistence_test.py
nvaytet: It's still not explicit. I see `set(stored_config.source_names) == {'monitor1', 'monitor2', 'monitor3'}`.
I can't figure out where it's happening.
Is it in the `backend_with_null_transport.workflow_controller.create_workflow_adapter`?

PR #544, opened 2025-11-19
Plot orchestrator/grid integration

SimonHeybrock:
This continues from #543 (but is still waiting to be fully integrated with the actual data streams, when #540 is ready).

Key changes:
- All workflows now define concrete `outputs` models. This is necessary since we want to configure plots in the grid *before* we have data.
- Some outputs models define `default_factory`. This allows for filtering available plotters (before we have actual data). As I am not entirely sure yet if this mechanism will work well in practice I did not roll this out to all workflow definitions.
- Replace job-based plot config modal by a new one.
- Removed `PlotGridTab`, which was added when `PlotGrid` was first implemented. Replaced with the new `PlotGridTabs`, which manages multiple tabs and integrates with the `PlotOrchestrator` (instead of `JobService`).
- Ability to reconfigure a plot once created.

Follow-up work:
- Integrate with `JobOrchestrator`, then remove the old `PlotCreationWidget`.

Updated/added plot-config-modal:

image

Grid with placeholders (shown before data available):

image

The gear icon allows for reconfiguring the plot.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
+++ b/src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
@@ -460,3 +460,8 @@ class PlotGridTabs:
     def panel(self) -> pn.Column:
         """Get the Panel viewable object for this widget."""
         return self._widget
+
+    @property
+    def tabs(self) -> pn.Tabs:
+        """Get the Tabs widget containing grid tabs."""
+        return self._tabs
--- a/tests/dashboard/widgets/plot_grid_tabs_test.py
+++ b/tests/dashboard/widgets/plot_grid_tabs_test.py
@@ -103,12 +103,12 @@ class TestPlotGridTabsInitialization:
 
     def test_creates_panel_tabs_widget(self, plot_grid_tabs):
         """Test that widget creates a Panel Tabs object."""
-        assert isinstance(plot_grid_tabs.panel, pn.Tabs)
+        assert isinstance(plot_grid_tabs.tabs, pn.Tabs)
 
     def test_starts_with_one_tab_when_no_grids(self, plot_grid_tabs):
         """Test that widget starts with only one tab when no grids exist."""
         # Should have exactly one tab (the Manage tab)
-        assert len(plot_grid_tabs.panel) == 1
+        assert len(plot_grid_tabs.tabs) == 1
 
     def test_initializes_from_existing_grids(
         self, plot_orchestrator, workflow_registry, plotting_controller
@@ -126,7 +126,7 @@ class TestPlotGridTabsInitialization:
         )
 
         # Should have 3 tabs: Manage + 2 grids
-        assert len(widget.panel) == 3
+        assert len(widget.tabs) == 3
 
     def test_subscribes_to_lifecycle_events(self, plot_orchestrator, plot_grid_tabs):
         """Test that widget subscribes to orchestrator lifecycle events."""
@@ -134,7 +134,7 @@ class TestPlotGridTabsInitialization:
         plot_orchestrator.add_grid(title='New Grid', nrows=3, ncols=3)
 
         # Should now have 2 tabs: Manage + New Grid
-        assert len(plot_grid_tabs.panel) == 2
+        assert len(plot_grid_tabs.tabs) == 2
 
 
 class TestGridTabManagement:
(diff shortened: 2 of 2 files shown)


src/ess/livedata/dashboard/widgets/plot_creation_widget.py
SimonHeybrock: Refactor in this file was to make `PlotConfigurationAdapter` work with the job-less new approach. As this widget will be removed when the new functionality is functional details might not matter much.

src/ess/livedata/config/workflow_spec.py
```diff
@@ -155,9 +156,21 @@ class WorkflowSpec(BaseModel):
...
+            "human-readable names and explanations for the UI. "
+            "\n\n"
+            "IMPORTANT: Use default_factory to provide an empty DataArray template "
+            "with the correct structure (dims, coords, units). This enables the "
+            "dashboard to perform automatic plotter selection before data exists. "
+            "Example:\n"
+            "    field: sc.DataArray = Field(\n"
+            "        default_factory=lambda: sc.DataArray(\n"
+            "            sc.zeros(dims=['x'], shape=[0], unit='counts'),\n"
+            "            coords={'x': sc.arange('x', 0, unit='m')}\n"
```
jokasimr: To clarify this a bit, is the `default_factory` used to select what kind of plot to display? (That's how I understood "automatic plotter selection".) If that's all it's used for, would it be more straightforward to directly specify the plotter type here instead?

src/ess/livedata/dashboard/widgets/plot_widgets.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # Dimensions
+    CELL_MARGIN = 2
+    TOOL_BUTTON_SIZE = 28
+    TOOL_BUTTON_TOP_OFFSET = '5px'
+    CLOSE_BUTTON_RIGHT_OFFSET = '5px'
+    TOOL_BUTTON_Z_INDEX = '1000'
+
+    # Typography
+    TOOL_BUTTON_FONT_SIZE = '20px'
```
jokasimr: I'm not sure if this is the typical way of doing things when working with panel, but it seems to me defining styles in a CSS class and assign that class on the buttons we create might be easier. :shrug: 

There are also a ton of `! important` in the styles below, maybe that's necessary but it would be good to know why

src/ess/livedata/dashboard/widgets/plot_widgets.py
```diff
@@ -0,0 +1,163 @@
...
+        button_type='light',
+        sizing_mode='fixed',
+        margin=(ButtonStyles.CELL_MARGIN, ButtonStyles.CELL_MARGIN),
+        styles={
+            'position': 'absolute',
+            'top': ButtonStyles.TOOL_BUTTON_TOP_OFFSET,
+            'right': right_offset,
+            'z-index': ButtonStyles.TOOL_BUTTON_Z_INDEX,
+        },
+        stylesheets=create_tool_button_stylesheet(button_color, hover_color),
```
jokasimr: Similar to the comment above, there are a lot of styles here. Why do we need to specify the styles both here directly on the element and on `button` in the stylesheet?

src/ess/livedata/handlers/monitor_workflow_specs.py
```diff
@@ -26,6 +28,30 @@ class MonitorDataParams(pydantic.BaseModel):
     )
 
 
+class MonitorHistogramOutputs(WorkflowOutputsBase):
```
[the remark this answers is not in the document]
SimonHeybrock: Examples:
- Detector view makes a cumulative and a "delta" output.
- Reduction workflow computes I(Q) as well as I(Qx, Qy)

So could could call it an "optimization", but it is more like "not doubling or tripling the compute needs".

src/ess/livedata/config/workflow_spec.py
jokasimr: And what about the second question?

src/ess/livedata/dashboard/widgets/plot_grid.py
```diff
@@ -424,33 +364,71 @@ def _show_error(self, message: str) -> None:
...
+        # Remove any existing widget at this exact position
+        if geometry in self._occupied_cells:
+            del self._occupied_cells[geometry]
+
+        self._insert_widget_into_grid(
+            geometry.row, geometry.col, geometry.row_span, geometry.col_span, widget
+        )
+
+        # Track occupation
+        self._occupied_cells[geometry] = widget
```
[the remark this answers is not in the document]
SimonHeybrock: You're right that having two data structures carries sync risk. However, Panel's GridSpec doesn't expose position/span metadata for multi-cell widgets, so we can't derive this efficiently. The sync risk is mitigated by:

  1. Only two methods modify both structures, and they do so in direct sequence
  2. Comprehensive tests that verify the behavior they enable (overlap detection, cell availability)
  3. Any desync would cause immediate, obvious test failures

We could add metadata to widgets and make _occupied_cells a cached property, but that adds complexity for unclear benefit. The current approach is straightforward and well-tested.

src/ess/livedata/dashboard/widgets/plot_widgets.py
jokasimr: It looks like we can style the wrapper by using the `:host` class: https://panel.holoviz.org/how_to/styling/apply_css.html

The advantage is that then we can put all styles in the stylesheet, which is easier because they are more contained.

jokasimr (approved):

Issue #545, opened 2025-11-20
Persist configs in `PlotOrchestrator`


SimonHeybrock:
Implementing this should also solve #480.

PR #546, opened 2025-11-20
Integrate job with plot orchestrator to make new plot grids functional

### LINKED ISSUE #541: Complete move to `PlotGrid`
We want to remove the old plotting widget and fully move to a plot-grid based architecture. There are a number of outstanding tasks:

- [x] Add small widget to add new plot-grids (a new tab). This needs to allow configuring the tab title as well as the grid size (2x2, 3x2, 3x3, ...). We probably need to generate a unique ID in addition to the display title.
- [ ] Add mechanism to persist plot grids configs. This will work hand-in-hand with the new `JobOrchestrator` (#540), which will move the UI towards a workflow-based approach.
- [x] Add mechanism to replace plots in the grids via the controller, e.g., if a job-transition for a workflow happens via the `JobOrchestrator`.

The backbone to this would probably be a `PlotOrchestrator` (working title, time will tell if it should remain independent of `PlottingController`, but during the refactor we want the old mechanism to remain functional so adding a fresh class seems benefitial). Conceptually it would be something like:

```python
@dataclass
class PlotTab:
    name: UUID
    title: str
    plots: tuple[corner1, corner2, WorkflowId, OutputName, PlotterConfig]  # better: dataclass(es)

class PlotOrchestrator:
    _plots: list[PlotTab]
```

Just like `JobOrchestrator` stores `WorkflowConfig`, `PlotOrchestrator` would store plotter configs and plots tabs.

SimonHeybrock:
This combines the work of recent PRs, bringing together JobOrchestrator and PlotOrchestrator. Configured plots grids refer to specific workflows. The subscription of the plot-orchestrator to the job-orchestrator ensures that plots auto auto-created once we have data, and auto-replaced when switching to a new job for a workflow. Note the difference: this is not about "resetting" (clearing) a running workflow, but about actually submitting a newly configured workflow, e.g., with changes parameters.

- `PlottingController.create_plot` had to be split into two phases. The old method still exists for the legacy plotting, but will be removed in the near future.
- Fixes #541, except for the #545 which will be dealt with in a follow-up.

Here is how the entire mechanism works (working on a document for the docs): 
image


src/ess/livedata/dashboard/job_orchestrator.py
```diff
@@ -110,27 +109,20 @@ def __init__(
...
-        self._source_names = source_names
         self._workflow_registry = workflow_registry
         self._config_store = config_store
         self._logger = logging.getLogger(__name__)
 
         # Workflow state tracking
         self._workflows: dict[WorkflowId, WorkflowState] = {}
 
+        # Workflow subscription tracking
+        self._subscriptions: dict[SubscriptionId, Callable[[JobNumber], None]] = {}
```
nvaytet: Can you explain the difference between a subscription id and a workflow id?

tests/dashboard/plot_orchestrator_test.py
```diff
@@ -1067,13 +1143,8 @@ def test_update_config_to_different_workflow_id_unsubscribe_old_subscribe_new(
         )
         plot_orchestrator.update_plot_config(cell_id, new_config)
 
-        # Should no longer be subscribed to old workflow
-        assert len(fake_job_orchestrator._workflow_subscriptions[workflow_id]) == 0
```
nvaytet: Many asserts have been removed. In some tests there are none left.
Are we now just checking that nothing raises? Is it enough?

tests/dashboard/plot_orchestrator_test.py
```diff
@@ -1126,5 +1189,534 @@ def test_add_cell_immediately_remove_it_then_workflow_commits_no_error(
...
+        workflow_id,
+        workflow_spec,
+        job_orchestrator,
+        fake_data_service,
+        fake_stream_manager,
+    ):
+        """get_cell_state should return error when plot creation fails."""
+        # Create plotting controller configured to fail
+        failing_controller = FakePlottingController(stream_manager=fake_stream_manager)
+        failing_controller.configure_to_raise(ValueError("Plot creation failed"))
```
nvaytet: Is it worth parametrizing the test with different kinds of errors (RuntimeError, ValueError, etc.), or does it not matter?
(I was thinking in case some try/except block somewhere misses one of the errors fx)

tests/dashboard/plot_orchestrator_test.py
```diff
@@ -1126,5 +1189,534 @@ def test_add_cell_immediately_remove_it_then_workflow_commits_no_error(
...
+            job_id=JobId(source_name='new_source', job_number=job_number1),
+            output_name='new_output',
+        )
+        fake_data_service[result_key2] = sc.scalar(2.0)
+
+        # Since workflow is still running, plot should be immediately recreated
+        plot2, error = plot_orchestrator.get_cell_state(cell_id)
+        assert plot2 is not None
+        assert error is None
+        # Note: We can't reliably test for a different instance since
```
nvaytet: Is it possible to check something about the config or the result to check the cell_id is now pointing to the new data?

src/ess/livedata/dashboard/job_orchestrator.py
SimonHeybrock: Workflow ID identifies a workflow. Subscription ID will be used, e.g., to unsubscribe. Since we may have multiple plots for a workflow we cannot repurpose the workflow ID for this.

src/ess/livedata/dashboard/job_orchestrator.py
```diff
@@ -424,3 +392,111 @@ def get_active_config(self, workflow_id: WorkflowId) -> dict[SourceName, JobConf
...
+                    callback_invoked = True
+                except Exception:
+                    self._logger.exception(
+                        'Error in immediate callback for subscription %s',
+                        subscription_id,
+                    )
+
+        return subscription_id, callback_invoked
+
+    def unsubscribe(self, subscription_id: SubscriptionId) -> None:
```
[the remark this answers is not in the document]
SimonHeybrock: Exactly!

tests/dashboard/roi_detector_plot_factory_test.py
```diff
@@ -121,6 +121,24 @@ def get_detector_pipe(
     raise KeyError(f"Key {detector_key} not found in data service")
```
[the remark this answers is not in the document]
SimonHeybrock: Good catch, added in https://github.com/scipp/esslivedata/pull/546/commits/ac9872f90f01608dd77f2d22d5b20daafb4176f7.

tests/dashboard/plot_orchestrator_test.py
SimonHeybrock: Done!

PR #547, opened 2025-11-21
Fix ModelWidget get_defaults to handle default_factory fields

SimonHeybrock:
The get_defaults() function only handled field_info.default, completely ignoring fields with default_factory.

This caused max_roi_count (which uses default_factory=lambda: min(3, _get_default_max_roi_count())) to return an empty dict, leading ParamWidget to default the int field to 0, which violated the ge=1 constraint.


nvaytet (commented):
looks good

PR #552, opened 2025-11-25
Add mantle_strip_view for DREAM instrument

SimonHeybrock:
## Summary
- Add `mantle_strip_view` as an orthogonal view to the existing `mantle_wire_view`
- `mantle_wire_view` sums over the `strip` dimension
- `mantle_strip_view` sums over `wire`, `module`, `segment`, `counter` dimensions
- Produces 1D output with 256 values (one per strip) for detector diagnostics

## Test plan
- Verified module imports successfully
- Tested transform function produces correct output shape and preserves total counts

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -262,7 +262,7 @@ class LinePlotter(Plotter):
         """Create a line plot from a scipp DataArray."""
         # TODO Currently we do not plot histograms or else we get a bar chart that is
         # not looking great if we have many bins.
-        if data.coords.is_edges(data.dim):
+        if data.dim in data.coords and data.coords.is_edges(data.dim):
             da = data.assign_coords({data.dim: sc.midpoints(data.coords[data.dim])})
         else:
             da = data
--- a/tests/dashboard/plots_test.py
+++ b/tests/dashboard/plots_test.py
@@ -156,6 +156,56 @@ class TestImagePlotter:
             render_to_bokeh(hv_element)
 
 
+class TestLinePlotter:
+    @pytest.fixture
+    def line_plotter(self):
+        """Create a LinePlotter instance."""
+        from ess.livedata.dashboard.plot_params import PlotParams1d
+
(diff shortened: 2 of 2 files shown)


src/ess/livedata/config/instruments/dream/factories.py
```diff
@@ -45,6 +45,13 @@ def _get_wire_view(da: sc.DataArray) -> sc.DataArray:
     )
 
 
+def _get_strip_view(da: sc.DataArray) -> sc.DataArray:
+    """Transform function to extract strip view (sum over all but strip)."""
+    return da.fold(dim=da.dim, sizes=_bank_sizes['mantle_detector']).sum(
+        ('wire', 'module', 'segment', 'counter')
```
jl-wynen: Why does this one sum over the other dims but the wire view flattens them?

SimonHeybrock: This view corresponds to the `z` in the mantle. The other view has `(r, phi)` for `r~wire` and `phi` is the flattened other dims. Together these two views give a view of all readout components.

SimonHeybrock: I think this will mainly be for IS/IDS to identify broken components. But obviously one will have to iterate on titles and description. My current plan is to source those from the IS/IDS after we have converged on what they want plotted.

SimonHeybrock: If a counter breaks (I don't know if it can independently) you will see a particular pattern in the wire view, as far as I understand.

The scientists asked for seeing all voxels, but that is not reasonable in streaming. Therefore I pushed for an imho more usable solution, one that shows all independent readout components. Here is a simpler example: On instruments with He3 tubes there is a single wire per tube connected to readout. Therefore, either the entire wire is faulty or not. For diagnostics it should thus be sufficient to show a plot that has one pixel (or value) per wire. Same concept here, except that there is an extra dimension, with 2 sets of orthogonaly arranged components, so we have two plots.

PR #554, opened 2025-11-26
Add ROI support for logical detector views

### LINKED ISSUE #531: ROI for logical views
Several things to note here:

1. It currently does not work, since logical views are based on a simple callable, which may perform sums etc. &mdash; but the ROI mechanism needs to be applied to the `detector_number` `sc.Variable` to obtain indices of pixels/voxels within the ROI.
2. Some scientists have pointed out that they would rather select component ranges (such as "wire 4-8") since this more more precise and reliable than drawing.

We need to think about a slightly different mechanism.

**Important note**
In some cases scientists have asking for this, without clarifying what they really expect to see. For example, selecting wires would not yield a meaningful TOF spectrum if the wire spans a wide range of two-theta, i.e., supporting this does not seem to be useful. Make sure to dig down to uncover the true underlying requirement first!

SimonHeybrock:
## Summary

- Enable ROI support for all logical detector views using `LogicalView` from ess.reduce 25.11.4
- Plotter registry now filters `roi_detector` based on workflow spec's `aux_sources`
- New `register_logical_detector_view_spec()` helper for instrument specs
- Fixes #531 (not addressing component ranges comment, but I think that could be handled using workflow configs instead).

## Why

Previously, `DetectorLogicalView` didn't support ROI because it used a simple projection that lacked index mapping. The frontend incorrectly showed the ROI plotter for any 2D data, even when ROI wasn't actually supported.

Now the backend uses `LogicalView` which provides proper index mapping, and the frontend checks `aux_sources` in the workflow spec to determine ROI availability.

### DIFF
--- a/tests/dashboard/plotting_controller_test.py
+++ b/tests/dashboard/plotting_controller_test.py
@@ -311,3 +311,71 @@ class TestGetAvailablePlottersFromSpec:
         assert len(plotters) > 0
         assert 'lines' in plotters
         assert 'image' in plotters
+
+
+class TestSpecRequirements:
+    """Tests for SpecRequirements validation logic."""
+
+    def test_empty_requirements_always_validates(self) -> None:
+        """Test that empty requires_aux_sources always returns True."""
+        from ess.livedata.dashboard.plotting import SpecRequirements
+
+        spec_req = SpecRequirements(requires_aux_sources=[])
+
+        # Should return True regardless of aux_sources_type
+        assert spec_req.validate_spec(None) is True
+        assert spec_req.validate_spec(object) is True
+
+    def test_returns_false_when_aux_sources_is_none_but_required(self) -> None:
+        """Test that validation fails when aux_sources is None but required."""
+        from ess.livedata.dashboard.plotting import SpecRequirements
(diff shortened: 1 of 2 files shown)


src/ess/livedata/config/instruments/odin/factories.py
```diff
@@ -11,54 +11,31 @@
...
-    #    cannot infer these automatically from the Timepix3 data.
-    # Note: We do this every time on import. Accessing `instrument.nexus_file`
-    # the first time will actually fetch the file using pooch, so it reverts
-    # this change every time.
-    with h5py.File(instrument.nexus_file, 'r+') as f:
-        det = f['entry/instrument/event_mode_detectors/timepix3']
-        trans = det['transformations/translation']
-        trans[...] = 1.0
-        det.attrs['axes'] = ['x_pixel_offset', 'y_pixel_offset']
-        det.attrs['detector_number_indices'] = [0, 1]
```
SimonHeybrock: The file still lacks this but (1) is no longer necessary because the xy-projection (which worked around the ROI limitation for logical views) is no longer needed. (2) is addressed by the change do dim names in `_fold_image` above.

jl-wynen (approved):

PR #571, opened 2025-12-02
Add ratemeter to DetectorView workflow

SimonHeybrock:
Add event count tracking to DetectorView workflow. Each finalize outputs `counts_total` and `counts_in_toa_range` as DataArrays with time coordinates. The time-coord should allow for optionally plotting this as a timeseries, or as an actual rate instead of counts. This would require implementing a rate extractor or support in the plotter, but this is generic work. The alternative would be to publish a rate instead of counts, but I felt that this is less flexible.

### DIFF
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -174,7 +174,7 @@ class Instrument:
         source_names: Sequence[str] | None = None,
         params: type[Any] | None = None,
         aux_sources: type[Any] | None = None,
-        outputs: type[Any] | None = None,
+        outputs: type[Any],
     ) -> SpecHandle:
         """
         Register workflow spec, return handle for later factory attachment.
@@ -208,7 +208,7 @@ class Instrument:
             configuration is handled by the Job layer and is not passed to the workflow
             factory function.
         outputs:
-            Optional Pydantic model class defining workflow outputs with metadata.
+            Pydantic model class defining workflow outputs with metadata.
             Field names should be simplified identifiers (e.g., 'i_of_d_two_theta')
             that match keys returned by workflow.finalize(). Field metadata (title,
             description) provides human-readable information for the UI.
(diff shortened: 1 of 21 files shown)


src/ess/livedata/handlers/detector_view.py
```diff
@@ -110,6 +114,10 @@ def accumulate(
         for roi_state in self._rois.values():
             roi_state.add_data(raw)
 
+        # Ratemeter: accumulate event counts
+        self._counts_total += raw.bins.size().sum().value
```
nvaytet: So is the assumption here that the weights are always 1?
I guess in the nexus or kafka formats, weights are always 1 (because they record time-stamps), so you probably can never have raw data with weights.

tests/handlers/detector_view_test.py
```diff
@@ -1144,3 +1144,148 @@ def test_unchanged_roi_resend_unnecessarily_resets_cumulative(
...
+        view = DetectorView(params=params, detector_view=mock_rolling_view)
+
+        # Accumulate some events (don't finalize yet)
+        view.accumulate(
+            {'detector': sample_detector_events}, start_time=1000, end_time=2000
+        )
+
+        # Clear should reset counts
+        view.clear()
+
```
nvaytet: Could check here that the counts are 0 to verify that the clear actually did what we think, and not for example cut the connection meaning that no new events came in, in which case the assert == TOTAL_SAMPLE_EVENTS below would still pass.

src/ess/livedata/handlers/detector_view.py
```diff
@@ -56,6 +56,10 @@ def __init__(
         self._roi_mapper = get_roi_mapper()
         self._current_start_time: int | None = None
 
+        # Ratemeter: track event counts per finalize period
+        self._counts_total: int = 0
+        self._counts_in_toa_range: int = 0
+
```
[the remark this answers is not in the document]
SimonHeybrock: It is `int`, not a variable, as it is just a number.

src/ess/livedata/handlers/detector_view.py
SimonHeybrock: Correct.

nvaytet (approved):

PR #572, opened 2025-12-02
Integrate monitor ratemeter with monitor-histogram workflow

### LINKED ISSUE #568: Monitor ratemeter in need for update and partially broken
There are several problems:

1. This is still using `TimeseriesAccumulator`, which will gather and restream the full timeseries with every update. This is no longer how we want to operate now (after #533). We only want the latest value (accumulator cleared after every update), i.e., a 0-D result. This is a bit at odds with how `StreamProcessor` currently works. Can we work around this limitation or do we need to change it to be more flexible? Should we use `StreamProcessor` at all or would directly wrapping `GenericTofWorkflow` be a better choice here?
2. The workflow uses `StreamProcessor` with a `GenericTofWorkflow`. This expects `NeXusData` as input. However, since we have defined the workflow as `namespace='monitor_data'` we end up with an (event-data) preprocessor that does not give as an `NeXusData`-compatible input for `ev44` monitors. We either need to use the `data_reduction` namespace or add a provider than does the translation.

SimonHeybrock:
Taking an approach equivalent to #571 instead of a separate workflow. I think this is both more convenient and simpler.

Note that the original motivation for having this as a dedicated workflow was to use a NeXus-based workflow that can compute TOF/wavelength, which users will want to use for defining an interval. As the same can be said for the monitor histogram plot this is however no reason for splitting it &mdash; if we fix one, we fix both.

We now publish two scalar results, total monitor counts and counts in TOA interval. A time coord is included so we can plot as a timeseries. The TOA interval is included which would allow for plotting a normalized rate (but this is not implemented in the frontend plotters right now).

Fixes #568.

image

### DIFF
--- a/docs/user-guide/adding-new-instruments.md
+++ b/docs/user-guide/adding-new-instruments.md
@@ -406,6 +406,36 @@ def setup_factories(instrument: Instrument) -> None:
         )
 ```
 
+## Grid Templates (Optional)
+
+Grid templates are pre-configured plot grid layouts that users can select when creating a new grid in the dashboard. They provide a convenient starting point with commonly used workflow configurations.
+
+### Creating Grid Templates
+
+The easiest way to create a grid template is to design it interactively in the dashboard:
+
+1. **Create and configure a grid in the UI**: Start the dashboard for your instrument, create a new grid, and configure it with the desired layout and workflow subscriptions.
+
+2. **Extract the grid configuration**: The dashboard persists grid configurations to `~/.config/esslivedata/<instrument>/plot_configs.yaml`. Open this file and find the grid you created under the `plot_grids.grids` key.
+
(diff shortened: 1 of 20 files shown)


src/ess/livedata/handlers/monitor_data_handler.py
```diff
@@ -3,41 +3,52 @@
...
-
 
 class MonitorStreamProcessor(Workflow):
-    def __init__(self, edges: sc.Variable) -> None:
+    def __init__(
+        self,
+        edges: sc.Variable,
+        *,
+        toa_range_enabled: bool = False,
+        toa_range: tuple[sc.Variable, sc.Variable] | None = None,
```
nvaytet: Can you make do with just one argument? If `toa_range` is `None`, it means toa_range is not enabled?

Issue #578, opened 2025-12-04
Look into independent creation of ROI image plot and ROI spectrum plot


SimonHeybrock:
Currently these are created in a nested layout, which (a) causes some sizing issues and (b) is inflexible.

PR #579, opened 2025-12-04
Add support for Polygon ROIs

SimonHeybrock:
Support polygon ROIs

Limitations:
- Color assignment when using both boxes and polygons is currently broken (colors in spectrum plot do not correspond). Solving this requires wider code changes, which I wanted to avoid now as the PR is large already. I intend to solve this as part of #578.
- Not adding a `PolyEdit` tool for now. This should be a simple extension in the future.

Example for the DREAM mantle detector:

image

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -224,6 +224,23 @@ All services use `OrchestratingProcessor` for job-based processing. See [src/ess
 - Test files follow pattern `*_test.py`
 - Tests use `pytest` with `--import-mode=importlib`
 
+### Slow Tests
+
+Tests marked with `@pytest.mark.slow` are excluded by default when running `pytest` directly:
+
+```sh
+# Default: runs fast tests only (~25s)
+python -m pytest
+
+# Include slow tests (~85s total)
+python -m pytest -m "not integration"
+
+# Run only slow tests
+python -m pytest -m slow
+```
+
+Slow tests are automatically included when running via `tox` (CI). The slow marker is applied to tests taking >2s, primarily Bifrost and LOKI data reduction workflow tests.
+
 ## Code Style Conventions
 
 - **Formatting**: Enforced by `ruff` (no manual conventions)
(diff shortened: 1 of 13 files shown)


src/ess/livedata/dashboard/roi_publisher.py
```diff
@@ -4,17 +4,25 @@
...
 from ..core.message import Message, MessageSink, StreamId, StreamKind
 
+ROIs = dict[int, RectangleROI] | dict[int, PolygonROI]
+ROIClass = type[RectangleROI] | type[PolygonROI]
+
+# Map ROI class to stream key suffix
+_READBACK_KEYS: dict[ROIClass, str] = {
+    RectangleROI: "roi_rectangle",
+    PolygonROI: "roi_polygon",
+}
```
jl-wynen: These should be defined in terms of `ROIGeometry.readback_key`. In general, I think there should be some association between the ROI models and `ROIGeometry` so we don't rely on ad hoc associations scattered throughout the code base.

src/ess/livedata/handlers/detector_view.py
```diff
@@ -193,19 +209,43 @@ def clear(self) -> None:
...
+        # Determine the index range for this geometry type
+        geometry_type = geometry_key.removeprefix('roi_')  # 'rectangle' or 'polygon'
+        geometry = next(
+            (
+                g
+                for g in self._roi_mapper.geometries
+                if g.geometry_type == geometry_type
+            ),
+            None,
+        )
```
jl-wynen: Move into the roi mapper to simplify this function here.

src/ess/livedata/dashboard/roi_detector_plot_factory.py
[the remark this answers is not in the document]
SimonHeybrock: I agree and dislike it at least as much - but I just had some ideas to improve this somewhat, see #584 (and potentially #583).

PR #582, opened 2025-12-04
Remove release CI action - no more uploads to pypi

MridulS:


SimonHeybrock (approved):

SimonHeybrock (changes requested):
How are we going to build and upload docs?

Issue #588, opened 2025-12-08
[Dream] Overlay list of known peaks


SimonHeybrock:
This is required by diffraction experiments. I think this should/will be addressed via #585.

Note also that via #628 we can now manually draw peak positions. What is missing is some integration with a file-based approach.

Issue #585, opened 2025-12-05
Plot Composition System: Generalize ROI editing and more as composable layers


SimonHeybrock:
# Plot Composition System

## Note

- **This is an alternative suggestion to #584**
- Prototype and implementation plan (copied below) can be found in [plot-composition-prototype](https://github.com/scipp/esslivedata/tree/plot-composition-prototype) branch.

### The ROI Detector Problem

The current `roi_detector` plotter requires special-case handling in `PlottingController`:
- Custom per-detector pipe creation in `setup_data_pipeline`
- Custom plot composition in `create_plot_from_pipeline`
- A dummy factory registration that raises `NotImplementedError`

This exists because ROI editing needs:
- Interactive streams (BoxEdit, PolyDraw) that persist across data updates
- Kafka subscriptions for ROI readback (bidirectional sync)
- Kafka publishing when user edits ROIs

These requirements don't fit the `Plotter` abstraction, which is designed for pure data→visualization transformation.

### The Key Insight

ROI editing is not a different way to plot data. It's **additional visual elements composed with a regular image plot**. The underlying visualization is just `ImagePlotter`—we're adding interactive shapes and Kafka sync alongside it.

### Generalization: Composition

Users have requested other composition features:
- **Peak markers**: Vertical lines at known peak positions on a spectrum
- **Reference curves**: Theoretical values overlaid on measured data
- **Multiple spectra**: Compare two data streams on one axis
- **Masks**: Show masked regions on 2D images

These all follow the same pattern: **multiple visual elements composed on a shared canvas**, each with its own data source and configuration.

## Core Insight: Everything is a Layer

Instead of distinguishing "plotters" from "overlays," we recognize that all visual elements are **layers** that get composed. Layers differ only in:

| Attribute | Examples |
|-----------|----------|
| Element type | image, curve, rectangles, vlines |
| Data source | pipeline, kafka stream, file, user interaction |
| Interaction tool | box_edit, poly_draw, or none |
| Update behavior | reactive (streams) or static |

A spectrum plot with peak markers and ROI editing on a detector image are the same mechanism—just different layer configurations.

### Layer Specification

```python
@dataclass(frozen=True)
class PlotLayer:
    """Specification for one visual layer in a composed plot."""

    name: str                       # Unique identifier within composition
    element: str                    # Element type: "image", "curve", "rectangles", ...
    source: DataSource              # Where data comes from
    params: dict = field(default_factory=dict)  # Element styling/options
[...]

PR #606, opened 2025-12-10
Plot composition

### LINKED ISSUE #607: Axis linking
#574 applied some fixes for making autoscale work in all cases. However, the wrapping in `hv.Layout` now proves problematic when composing multiple `DynamicMaps` in #606, so I have remove that workaround. Instead, we no longer set `shared_axes=False`. This may be unwanted!

We need to think and investigate more. Maybe there is a way that addresses all issues. Or maybe sharing axes is ok, but we just need to be better at naming axes, to avoid unintentional linking &mdash; the `x` and `y` of different detector panels are not really the same, so linking is unwanted, but this could be addressed by giving them a clearer (unique) name.

Relevant pointers can be found in #5252 and https://discourse.holoviz.org/t/axiswise-true-does-not-work/1018?u=linuxiscool.

SimonHeybrock:
This is working towards #585 and #588. We add a mechanism to overlay multiple plots. Previously this was only possible for multiple sources (such as detector banks) from the same workflow. We now have a generalized system on top of this.

In this PR we are not yet adding functionality for adding layers that are static (such as loaded from a file for #588, or from a previous job) or for the ROI shape drawing. This is left for future work.

Note(s):
- Sizing mode should probably be a cell option, not plot option (see hack to extract from first layer). Time will tell if/how we want to address this, but this should work for now.
- `Plotter` still supports creating `hv.Layout` (see special handling here when composing layers). It is unclear if we want to keep this option. Again, time will tell.

Example with 2 "layers", current and cumulative monitors (this particular examples has limited utility in practice, but demonstrates the feature):
image

Also fixes #607 since I eventually found the `linked_axes=False` option.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,7 @@ jobs:
     environment: release
     permissions:
       id-token: write
-    if: github.event_name == 'release' && github.event.action == 'published'
+    if: false # Do not upload to PyPI
     steps:
       - uses: actions/download-artifact@v4
       - uses: pypa/gh-action-pypi-publish@v1.12.4
@@ -50,6 +50,7 @@ jobs:
   docs:
     needs: [upload_pypi]
     uses: ./.github/workflows/docs.yml
+    if: always() # Ensure docs are built even if upload_pypi is skipped
     with:
       publish: ${{ github.event_name == 'release' && github.event.action == 'published' }}
     secrets: inherit
--- a/docs/user-guide/adding-new-instruments.md
+++ b/docs/user-guide/adding-new-instruments.md
@@ -252,24 +252,32 @@ For detectors with complex 3D geometries, use helper functions to register proje
 ```python
 from ess.livedata.handlers.detector_view_specs import register_detector_view_spec
 
-# XY plane projection (e.g., for endcap detectors)
+# Single projection for all detectors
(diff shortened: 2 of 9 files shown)


src/ess/livedata/dashboard/plot_orchestrator.py
```diff
@@ -607,43 +775,83 @@ def parse_raw_cell(self, cell_data: dict[str, Any]) -> PlotCell | None:
...
+            layer = self._parse_raw_layer(layer_data)
+            if layer is not None:
+                layers.append(layer)
+
+        if not layers:
+            return None
+
+        return PlotCell(geometry=geometry, layers=layers)
+
+    def _parse_raw_layer(self, layer_data: dict[str, Any]) -> Layer | None:
```
nvaytet: Do the changes to the config require updates to the `src/ess/livedata/config/instruments/bifrost/grid_templates/*.yaml` files?

Also, are the changes here retro-compatible or do they not work with old grid config files?

SimonHeybrock: Note the legacy-code that should allow reading old config files. But yes, we will want to update our files, latest when we remove legacy handling. Once files are converted they cannot be read with an old dashboard though.

PR #612, opened 2025-12-15
Add announcements pane to dashboard sidebar

SimonHeybrock:
## Summary

- Add a Markdown pane at the top of the sidebar that periodically reloads from a file
- Allows pushing updates (announcements, known issues, tips) to users without restarting the dashboard. This is meant to allow updating this info without redeploying the dashboard. Time will tell if this mechanism is reasonable.
- Pane refreshes every 30 seconds by reading `announcements.md`

image

### DIFF
--- a/src/ess/livedata/dashboard/reduction.py
+++ b/src/ess/livedata/dashboard/reduction.py
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 import argparse
-from pathlib import Path
+import urllib.request
+from urllib.error import URLError
+from urllib.request import urlopen
 
 import holoviews as hv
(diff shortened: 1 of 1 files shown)


src/ess/livedata/dashboard/reduction.py
```diff
@@ -53,6 +56,23 @@ def __init__(
...
+            except FileNotFoundError:
+                return "*No announcements file found.*"
+
+        pane = pn.pane.Markdown(read_announcements(), sizing_mode='stretch_width')
+
+        def refresh():
+            pane.object = read_announcements()
+
+        pn.state.add_periodic_callback(refresh, period=30_000)  # 30 seconds
+        return pane
```
MridulS: Shouldn't this be fetching from a remote url instead of reading from disk?

We shouldn't redeploy the whole application just to update annoucements.

Issue #611, opened 2025-12-12
Plot creation does not work for correlation histogram


SimonHeybrock:
The "new" way of creating plots without existing reference to data in `DataService` is based on selecting a workflow and its outputs. However, correlation-workflows are not uniquely defined by just that &mdash; the aux-source-name(s) is needed as well. The current mechanism simply cannot handle this.

The solution (compatible with the work-in-progress `WorkflowStatusListWidget`) would be to introduce a `WorkflowTemplate` mechanism.

However, after more considerations this solution appears relatively complex. I now realize that the current approach actually has bad UX, since it treats creating a correlation histogram as "running a workflow" (running after primary workflows that generate data, i.e., the correlation-workflow consumes now not raw-data streams but derived-data streams). I don't think users think about it like that. I expect that they rather think of "_plot the detector counts against temperature_". In other words, we should make this a special kind of plot.

Update: `correlation-plotter` branch (and derived) contains a solution for this. There are still some things in `PlottingController` and `PlotOrchestrator` that I am unhappy about (too many changes, too complex), but in principle it is working.

PR #628, opened 2025-12-17
Add static geometry overlays

### LINKED ISSUE #626: Static geometry overlays
There are several cases where we want to overlay things such as `hv.Rectangles` or `hv.VLines` on a plot. Concrete examples are:
- Draw separators for detector elements (#619)
- Draw locations of known peaks

#606 added a way of adding new layers on an existing plot. The idea here is to add layers that do not take *any* input streams, but instead draw a static overlay based on parameters. Concretely my idea is to:

1. Add Pydantic models that can be used to configure `hv.Rectangles`, `hv.VLines`, ...; Model fields should (indirectly) map to constructor inputs and options (lists of rectangle corners, optional colors, ...).
2. Extend the `PlotConfigModal` wizard steps:
   - Step 1: Add a new synthetic (not workflow-based) category (namespace) "Static overlay" -> workflow "Geometric" -> none or dummy output
   - Step 2: Available plotters should list `Rectangles`, `VLines`, ...
   - Step 3: Plot model defined earlier use for regular widget setup.
4. Create the element directly from these models. This will require somehow extending `PlotOrchestrator` to allow setting up layers that have zero workflow subscriptions (there is no stream).

It is really important to think about 4.) and get it right, because it is actually an example of a generalization we need also for #611 where we need to setup a subscription to *multiple* workflows. We should ensure that any change we make here to `PlotOrchestrator` (to support 0 subscriptions) is compatible with that, and ideally already introduces any interface/model/dataclass changes we need there.

SimonHeybrock:
Fixes #626

This PR adds support for static geometry overlays - plot layers that don't subscribe to any workflow data but instead draw shapes directly from user-configured parameters. This enables use cases like:
- Drawing detector element separators
- Marking known peak locations
- Adding reference lines or regions

## Example

image

## Architecture changes

The core architectural change is generalizing `PlotConfig` to support zero, one, or multiple data sources. This prepares the foundation for future multi-source support (correlation histograms, #611):

- Introduced `DataSourceConfig` to encapsulate workflow subscription details
- `PlotConfig.data_sources` is now a list that determines layer type:
  - Empty source_names → static overlay (no subscription)
  - Single data source → standard plot (current behavior)
  - Multiple data sources → future correlation histograms

`PlotOrchestrator` now branches on data source count, creating static plots immediately without workflow subscription.

## New components

- **`static_plots.py`**: Pydantic models and plotters for geometric overlays
  - `RectanglesParams`/`RectanglesPlotter` - draw rectangles from corner coordinates
  - `VLinesParams`/`HLinesParams`/`LinesPlotter` - draw vertical/horizontal lines
  - Full styling support: color, line width, line style (solid/dashed/dotted), opacity

- **`StaticPlotConfigurationAdapter`**: Configuration adapter for static overlays that skips source selection

- **`PlotterCategory` enum**: Distinguishes DATA plotters (require workflow) from STATIC plotters (overlays)

## UI integration

The plot configuration wizard now includes a "Static Overlay" namespace with:
- A name input field for identifying the overlay
- Plotter selection (Rectangles, Vertical Lines, Horizontal Lines)
- Parameter configuration with ColorPicker widget support for color fields

## Input format

Coordinates use a JSON-based format for ease of entry:
- Rectangles: `[x0,y0,x1,y1], [x0,y0,x1,y1], ...`
- Lines: `10, 20, 30, ...`

### DIFF
--- a/src/ess/livedata/config/instruments/bifrost/factories.py
+++ b/src/ess/livedata/config/instruments/bifrost/factories.py
@@ -366,18 +366,18 @@ def _create_base_reduction_workflow():
         )
         per_arc = 3 * 900
         detector_number_step = 100 // pixels_per_tube
-        detector_number_offset = sc.arange(
-            'detector_number_offset', 0, per_arc, step=detector_number_step, unit=None
-        )
+        detector_number = sc.arange(
+            'detector_number', 0, 5 * per_arc, step=detector_number_step, unit=None
+        ).fold(dim='detector_number', sizes={'arc': 5, 'detector_number': -1})
         return SpectrumView(
             data.fold('pixel', sizes={'pixel': pixels_per_tube, 'subpixel': -1})
             .drop_coords(tuple(data.coords))
             .bins.concat('subpixel')
-            .flatten(dims=('tube', 'channel', 'pixel'), to='detector_number_offset')
+            .flatten(dims=('tube', 'channel', 'pixel'), to='detector_number')
             .hist(event_time_offset=edges)
             .assign_coords(
                 event_time_offset=edges.to(unit='ms'),
-                detector_number_offset=detector_number_offset,
+                detector_number=detector_number,
             )
         )
 
--- a/src/ess/livedata/dashboard/autoscaler.py
+++ b/src/ess/livedata/dashboard/autoscaler.py
@@ -14,57 +14,122 @@ class Autoscaler:
     data. However, this leads to losing the current pan/zoom state when new data
     arrives, making it unusable for interactive exploration.
     Instead, we use this class to track the bounds of the data and update the plot with
-    framewise=True only when the bounds *increase*, i.e., new data extends the
-    existing bounds. This way, we keep the current pan/zoom state most of the time while
-    still allowing the plot to grow as new data comes in. This is especially important
-    since there seems to be no way of initializing holoviews.streams.Pipe without
-    initial dummy data (such as `None`), i.e., we need to return an empty plot with no
-    good starting guess of bounds.
+    framewise=True only when the bounds change significantly. This way, we keep the
+    current pan/zoom state most of the time while still allowing the plot to adjust
+    as data changes. This is especially important since there seems to be no way of
+    initializing holoviews.streams.Pipe without initial dummy data (such as `None`),
+    i.e., we need to return an empty plot with no good starting guess of bounds.
     """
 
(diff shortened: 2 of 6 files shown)


nvaytet (commented):
Not much to say, looks good overall :+1:

src/ess/livedata/dashboard/static_plots.py
```diff
@@ -0,0 +1,332 @@
...
+
+    plotter_registry.register_plotter(
+        name='hlines',
+        title='Horizontal Lines',
+        description='Draw static horizontal lines as an overlay. '
+        'Define y-positions for the lines.',
+        data_requirements=static_data_requirements,
+        factory=LinesPlotter.hlines,
+        category=PlotterCategory.STATIC,
+    )
```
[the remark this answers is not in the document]
SimonHeybrock: Yes. When you click gears and change something it is essentially recreating the plot.

src/ess/livedata/dashboard/static_plots.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+def _parse_rectangle_list(v: str) -> list[list[int | float]]:
+    """Parse comma-separated JSON arrays into a list of rectangle coordinates.
+
+    Example: "[0,0,10,10], [20,20,30,30]" -> [[0,0,10,10], [20,20,30,30]]
+    """
+    v = v.strip()
+    if not v or v == '[]':
+        return []
```
SimonHeybrock:
```suggestion
```
as this should be handled equally well below

tests/dashboard/static_plots_test.py
```diff
@@ -0,0 +1,158 @@
...
+        assert coords.parse() == [(0.0, 0.0, 10.0, 10.0), (20.0, 20.0, 30.0, 30.0)]
+
+    def test_empty_string_rejected(self):
+        """Empty string is rejected."""
+        with pytest.raises(ValueError, match="At least one rectangle is required"):
+            RectanglesCoordinates(coordinates="")
+
+    def test_empty_brackets_rejected(self):
+        """Empty brackets is rejected."""
+        with pytest.raises(ValueError, match="At least one rectangle is required"):
```
SimonHeybrock:
```suggestion
        with pytest.raises(ValueError, match="1 validation error for RectanglesCoordinates"):
```

PR #640, opened 2026-01-05
Multi-job subscription abstraction

SimonHeybrock:
This is a preparation for a fix/reimplementation of the correlation-histogram plotting functionality.


SimonHeybrock:
Included in #642 (with further modifications).

PR #644, opened 2026-01-07
Add ROI request plotters as composable layers

### LINKED ISSUE #585: Plot Composition System: Generalize ROI editing and more as composable layers

SimonHeybrock:
## Summary

This PR implements ROI request plotters (BoxEdit, PolyDraw) as composable layers, completing the layer-based composition system for ROI editing.

- Adds `BaseROIRequestPlotter` base class with shared logic for ROI request layers
- Implements `RectanglesROIRequestPlotter` and `PolygonsROIRequestPlotter` as concrete layers
- Removes the old `ROIDetectorPlotFactory` and its special-case handling
- Centralizes `max_roi_count` configuration in `detector_view_specs.py`
- Consolidates ROI plotter registration to prevent `DataRequirements` divergence

Fixes #585 (most of it—the general layer system was implemented in the base branch).

## Known Caveats

These limitations will be addressed in follow-up work:

1. **Three layers complicate setup**: ROI plots now require three separate layers (image, readback, request), which complicates configuration and results in three space-consuming toolbars
2. **Readback no longer updates request**: Editing the request layer does not sync with readback changes
3. **Five layers needed for rectangles + polygons**: If both rectangle and polygon ROIs are needed, this requires five separate layers


src/ess/livedata/dashboard/plotting_controller.py
```diff
@@ -238,36 +206,24 @@ def create_plot_from_pipeline(
...
-                    )
-                )
-                plots.append(detector_with_rois)
-
-            return hv.Layout(plots).opts(shared_axes=False)
-
         plotter = plotter_registry.create_plotter(plot_name, params=params)
 
+        # ROI request plotters need a publisher and return a DynamicMap with
+        # BoxEdit/PolyDraw streams already attached. Don't wrap again.
```
nvaytet: Can you explain a bit more what is meant by "Don't wrap again"?
What is being wrapped below (outside of the if), and what is the wrapper doing?

src/ess/livedata/dashboard/roi_request_plots.py
```diff
@@ -0,0 +1,769 @@
...
+        """
+        if not data or not data.get("xs"):
+            return {}
+
+        xs_list = data.get("xs", [])
+        ys_list = data.get("ys", [])
+
+        rois = {}
+        for i, (xs, ys) in enumerate(zip(xs_list, ys_list, strict=True)):
+            # Skip polygons with fewer than 3 vertices
```
nvaytet: I'm guessing from this that polygons are always closed, so 3 vertices make a triangle which also has an edge between the last and the first vertex. Right? (basically my question is you only need 3 vertices, not 4?).

src/ess/livedata/dashboard/roi_request_plots.py
```diff
@@ -0,0 +1,769 @@
...
+        -------
+        :
+            List of dicts with 'x', 'y' (and optionally 'color') for HoloViews Polygons.
+            Returned in sorted order by ROI index.
+        """
+        polygons = []
+        for idx in sorted(rois.keys()):
+            roi = rois[idx]
+            poly_dict: dict[str, Any] = {
+                'x': [float(v) for v in roi.x],
```
nvaytet: Do you have to convert to lists or can you send numpy arrays to holoviews?

src/ess/livedata/dashboard/roi_request_plots.py
```diff
@@ -0,0 +1,769 @@
...
+            self._x_unit = str(data.coords['x'].unit) if data.coords['x'].unit else None
+        if 'y' in data.coords:
+            self._y_unit = str(data.coords['y'].unit) if data.coords['y'].unit else None
+        initial_rois = self._parse_initial_geometry()
+
+        # Create pipe for programmatic updates
+        pipe = hv.streams.Pipe(data=[])
+        dmap = hv.DynamicMap(self._create_element, streams=[pipe])
+
+        # Create edit stream (store to prevent GC)
```
nvaytet: What exactly gets GCed if this is omitted?

src/ess/livedata/dashboard/roi_request_plots.py
```diff
@@ -0,0 +1,769 @@
...
+        Parameters
+        ----------
+        data:
+            Dictionary from PolyDraw stream with keys 'xs', 'ys'.
+        x_unit:
+            Unit for x coordinates (from the detector data coordinates).
+        y_unit:
+            Unit for y coordinates (from the detector data coordinates).
+        index_offset:
+            Starting index for polygon ROIs (e.g., 4 for indices 4-7).
```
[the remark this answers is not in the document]
SimonHeybrock: It ensures that all ROIs have different colors, i.e., rectangles and polygons do not both start with the same color sequence (blue->red->yellow->...).

src/ess/livedata/dashboard/roi_request_plots.py
SimonHeybrock: Correct, the are always closed.

SimonHeybrock:
> > Five layers needed for rectangles + polygons: If both rectangle and polygon ROIs are needed, this requires five separate layers
> 
> 😞 Can we have a single overlay layer that plots rectangles, polygons and static lines?

No, because that would complicate the mechanism. We will instead likely support "collapsing" the headers, or alternatively support something like layer-groups.

PR #651, opened 2026-01-08
Add title and description metadata for source names

### LINKED ISSUE #604: Titles for source names
For many or most things (workflows, params, outputs, ...) we have a `name` (internal unique identifier), a `title` (to display in the UI), and potentially a `description`. However, for `source_name` we still display just that - but `endcap_forward_detector` may not be ideal for display in UI and plots. We should therefore considering introducing `title` and possibly `description` concepts also for source-names (detectors in particular, but maybe more generally).

SimonHeybrock:
## Summary

Introduces `SourceMetadata` Pydantic model to allow human-readable titles and descriptions for data sources (detectors, monitors, timeseries). This addresses the UX issue where internal identifiers like `endcap_forward_detector` are displayed in the UI instead of user-friendly names.

- Add `SourceMetadata` model with `title` and `description` fields to `Instrument` class
- Wire instrument config through dashboard layer for title lookup
- Update source selector widgets to display titles instead of internal names
- Update plot cell toolbars to use source titles in display and tooltips
- Add DREAM instrument source metadata as example

The title lookup falls back to the source name when no metadata is defined, ensuring backwards compatibility.

Closes #604

## Example

image

## Test plan

- [x] Unit tests added for `SourceMetadata` model and lookup methods
- [x] All existing tests pass (1903 passed)
- [x] Manual verification: workflow source selector shows titles
- [x] Manual verification: plot source selector shows titles
- [x] Manual verification: plot toolbar shows titles
- [x] Manual verification: WorkflowStatusWidget shows titles

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/requirements/basetest.txt
+++ b/requirements/basetest.txt
@@ -62,7 +62,7 @@ essreduce==25.12.1
     #   essspectroscopy
 essreflectometry==25.12.0
     # via -r basetest.in
-esssans==25.11.0
+esssans==25.12.0
     # via -r basetest.in
 essspectroscopy==25.12.0
     # via -r basetest.in
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -41,7 +41,7 @@ class LogicalViewConfig:
     title: str
     description: str
     source_names: list[str]
-    transform: Callable[[sc.DataArray], sc.DataArray]
+    transform: Callable[[sc.DataArray, str], sc.DataArray]
     roi_support: bool = True
     output_ndim: int | None = None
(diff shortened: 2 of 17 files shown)


SimonHeybrock:
> > Currently only one I think: There is a tooltip icon in the toolbar of a plot (the header row).
> 
> I meant I couldn't see `get_source_description` called anywhere, but maybe it is hidden and still used?

Oh, sorry I misunderstood you, I thought you meant the title not being used in tooltips. Indeed, we currently do not display the description field anyway.

nvaytet (approved):

PR #652, opened 2026-01-08
Add start_time and end_time coords for accurate time display in plot titles

### LINKED ISSUE #444: Include start_time and end_time in result messages.
This is included in `JobStatus` messages, but we should add it also as 0-D coords in the result `DataArray`.

### LINKED ISSUE #634: Display current "lag" of plots vs. the data they show.
This could/should probably make use of #444.

SimonHeybrock:
## Summary

This PR fixes #444 and fixes #634 by adding temporal provenance to job outputs and displaying time information in plot titles.

**Key changes:**
- `Job.get()` now adds `start_time` and `end_time` as 0-D coordinates (in ns) to all `DataArray` outputs, enabling the dashboard to show when data was collected and compute lag
- Plot titles now display the time range and lag indicator (e.g., "12:34:56.7 - 12:35:02.3 (Lag: 2.1s)")
- Delta outputs (e.g., "current" counts) set their own time coords representing the period since the last finalize, rather than the entire job duration
- `TemporalBuffer` accumulates `start_time`/`end_time` as 1-D coordinates per time slice, allowing extractors to compute accurate bounds for windowed views

**Note:** This is an initial implementation to evaluate the UX. We expect further changes based on feedback, for example:
- Adjusting what exactly is displayed in the title (format, precision, which time bounds to show)
- Whether the time info belongs in the plot title or should be displayed elsewhere (e.g., a status bar, tooltip, or dedicated widget)
- How to handle edge cases like missing time coordinates or very old/stale data

## Example

Screenshot 2026-01-08 at 15 08 46

## Test plan

- Unit tests added for `Job.get()` time coord injection, extractor time handling, and `TemporalBuffer` accumulation
- Tests verify that workflows can override job-level times for delta outputs
- Tests verify that extractors compute correct time bounds from windowed/aggregated data

### DIFF
--- a/src/ess/livedata/config/__init__.py
+++ b/src/ess/livedata/config/__init__.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 
-from .instrument import Instrument, instrument_registry
+from .instrument import Instrument, SourceMetadata, instrument_registry
 from .models import ConfigKey
 
-__all__ = ['ConfigKey', 'Instrument', 'instrument_registry']
+__all__ = ['ConfigKey', 'Instrument', 'SourceMetadata', 'instrument_registry']
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -7,12 +7,30 @@ from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field
 from typing import Any
 
+import pydantic
 import scipp as sc
 import scippnexus as snx
 
 from ess.livedata.handlers.workflow_factory import SpecHandle, WorkflowFactory
 
 from .workflow_spec import WorkflowSpec
+
+
+class SourceMetadata(pydantic.BaseModel):
+    """Metadata for a data source (detector, monitor, or timeseries).
+
+    Parameters
+    ----------
+    title:
+        Human-readable title for display in the UI.
+    description:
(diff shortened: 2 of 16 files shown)


src/ess/livedata/dashboard/temporal_buffers.py
```diff
@@ -438,6 +488,10 @@ def _metadata_matches(self, data: sc.DataArray) -> bool:
             new = data
 
         # Create template with reference data but incoming metadata
-        template = new.assign(self._reference.data).drop_coords('time')
+        # Drop all accumulated coords for comparison
+        template = new.assign(self._reference.data)
+        for coord in self._ACCUMULATED_COORDS:
+            if coord in template.coords:
+                template = template.drop_coords(coord)
```
nvaytet: Use a single call to `drop_coords` (as above)?

tests/dashboard/extractor_time_coords_test.py
```diff
@@ -0,0 +1,213 @@
...
+        data = _make_buffered_data(num_points=5, time_span_seconds=5.0)
+        extractor = LatestValueExtractor()
+
+        result = extractor.extract(data)
+
+        # Verify coords are from the last slice
+        expected_start = data.coords['start_time'].values[-1]
+        expected_end = data.coords['end_time'].values[-1]
+        assert result.coords['start_time'].value == expected_start
+        assert result.coords['end_time'].value == expected_end
```
nvaytet: Use `sc.identical`?

PR #655, opened 2026-01-09
Support TOF mode for monitors

SimonHeybrock:
## Summary

This implement support for TOF mode (and  prepares for "wavelength" mode) for monitors, similar to the equivalent changes for detector views in #658. This is crucial for users, as time-of-arrival is just an artifical offset since the last pulse and not meaningful . Time-of-arrival mode will likely still be used for commissioning and other sanity checks.

TOF mode is only available for instrument that have a set of known chopper configurations and associated TOF-lookup tables as part of the instrument package.

### Under the hood

Replaces the legacy `MonitorStreamProcessor` with a new Sciline-based workflow using `StreamProcessorWorkflow`, aligning monitor processing with the detector view architecture from #658.

### Key changes

- **Sciline-based workflow**: Monitor processing now uses composable Sciline providers (`histogram_raw_monitor`, `histogram_tof_monitor`, `cumulative_view`, `window_view`, `counts_total`, `counts_in_range`) instead of the bespoke `MonitorStreamProcessor` class
- **Shared infrastructure with detector view**: Uses `NoCopyAccumulator` and `NoCopyWindowAccumulator` from the consolidated `accumulators.py` module
- **TOF coordinate mode support**: New `coordinate_mode` parameter with `'toa'` (time-of-arrival) and `'tof'` (time-of-flight) modes, mirroring the detector workflow capability
- **Standard preprocessors**: Uses `ToNXevent_data` preprocessor for events instead of the custom `CollectTOA` accumulator

### Removed

- `MonitorStreamProcessor` class
- `MonitorHandlerFactory` class  
- `CollectTOA` accumulator
- `monitor_data_handler.py` module

The new workflow produces identical outputs (`cumulative`, `current`, `counts_total`, `counts_in_toa_range`) with the same time coordinate handling.

## Test plan

- Existing monitor workflow tests updated and passing
- Tests for TOF coordinate mode functionality


src/ess/livedata/config/instruments/dream/factories.py
```diff
@@ -76,24 +86,58 @@ def _detector_view_workflow_factory(
...
+        mode = params.coordinate_mode.mode
+        if mode == 'wavelength':
+            raise NotImplementedError(
+                "wavelength mode not yet implemented for monitors"
+            )
+
+        # monitor_bunker is only 6.62 m from the source, which is outside the
+        # DREAM TOF lookup table range (59.85-80.15 m). Only monitor_cave
+        # (Ltotal 72.33 m) is compatible with TOF mode.
+        if mode == 'tof' and source_name == 'monitor_bunker':
```
nvaytet: We could also wait for https://github.com/scipp/essdiffraction/pull/236

SimonHeybrock:
> Generally, I feel that having toa-only monitors and monitors that can compute tof is making things more complicated.
> 
> Can't we just have monitors that can always compute both toa and tof (and then wait for the necessary changes in essreduce and other packages to update the tables)?

You mean, to always compute and publish both?

nvaytet:
The name TOA-only suggests that you have some monitors that are computing/publishing only toa, and you have others that do both toa and tof (maybe that is not correct?).

I was thinking we would have monitors that can do everything (both toa and tof) and whether they compute and publish both is controlled by the user. But we don't need to have 2 different classes.

Alternatively, have monitors that do only toa and others that only do tof.
So you would have `TOAmonitor` and `TOFmonitor`, and the separation is clearer?

But maybe I misunderstood how this works?

SimonHeybrock:
> The name TOA-only suggests that you have some monitors that are computing/publishing only toa, and you have others that do both toa and tof (maybe that is not correct?).
> 
> I was thinking we would have monitors that can do everything (both toa and tof) and whether they compute and publish both is controlled by the user. But we don't need to have 2 different classes.

We need two classes (pydantic models) because the UI is generated from them. We could of course use a single one and than give an error if the user tries to enable TOF mode, but that would be weird.

> Alternatively, have monitors that do only toa and others that only do tof. So you would have `TOAmonitor` and `TOFmonitor`, and the separation is clearer?

What if you don't have chopper cascade of TOF lookup available? I am pretty sure people will need to display the RAW monitor, even if it is just to verify that the TOF lookup is correct, or the phasing is as desired.

src/ess/livedata/config/instruments/bifrost/factories.py
```diff
@@ -83,6 +83,20 @@ def setup_factories(instrument: Instrument) -> None:
         _detector_view_factory.make_workflow
     )
 
+    # Monitor workflow factory (TOA-only)
+    from ess.livedata.handlers.monitor_workflow import create_monitor_workflow
+    from ess.livedata.handlers.monitor_workflow_specs import TOAOnlyMonitorDataParams
+
+    @specs.monitor_handle.attach_factory()
+    def _monitor_workflow_factory(source_name: str, params: TOAOnlyMonitorDataParams):
+        """Factory for Bifrost monitor workflow (TOA-only)."""
```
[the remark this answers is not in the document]
SimonHeybrock: Is it good to assume that it would always be a path, or should the interface be for returning something in-memory?

src/ess/livedata/config/instruments/dream/factories.py
SimonHeybrock: We have many instrument where this needs to be added/enabled. No need to hold this up.

nvaytet (approved):

PR #663, opened 2026-01-19
Display timeseries time axis in local time instead of UTC

### LINKED ISSUE #654: Timeseries time axis should use local time not UTC
See title. We already to this when displaying time info in the plot titles (see PR #652).

SimonHeybrock:
## Summary
- Add shared time utility module (`time_utils.py`) with functions for local timezone handling
- Shift datetime64 coordinates by local timezone offset in `_ensure_datetime_coord()` so Bokeh displays local time on the axis
- Refactor plot title time formatting to use the shared utility

This makes the timeseries time axis display consistent with the plot title time display added in PR #652.

Fixes #654

## Test plan
- Unit tests added for `time_utils.py` functions
- Test added to verify timezone offset is applied in `FullHistoryExtractor`
- All 823 dashboard tests pass

### DIFF
--- a/src/ess/livedata/dashboard/extractors.py
+++ b/src/ess/livedata/dashboard/extractors.py
@@ -9,6 +9,7 @@ from typing import Any
 import scipp as sc
 
 from .plot_params import WindowAggregation
+from .time_utils import get_local_timezone_offset_ns
 
 
 def _extract_time_bounds_as_scalars(data: sc.DataArray) -> dict[str, sc.Variable]:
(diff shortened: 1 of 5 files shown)


MridulS (approved):
Looks okay to me. I'm not necessarily a fan of adding indirection like defining a new function `get_local_timezone_offset_ns` instead of just inlining the offset.

SimonHeybrock:
One could display the offset, but yes it is the server's local timezone, i.e., Lund where the experiment was done. Not the user's browser, as we we care about is experiment-time.

PR #692, opened 2026-02-11
Add v1 of concept inventory, to be curated and extended.

SimonHeybrock:
Mostly generated overview of concepts/topics to cover.

### DIFF
--- a/docs/developer/design/concept-inventory.md
+++ b/docs/developer/design/concept-inventory.md
@@ -140,7 +140,7 @@ Plotters use a two-stage pattern: `Plotter.compute(data)` (stage 1) runs on the 
 
 ### D3. Dirty-Flag and Version-Based Polling
 
-Cross-thread communication uses dirty flags and version counters instead of direct callbacks. The background thread sets dirty flags on plotters when data arrives. The periodic callback checks `is_dirty()` and calls `pipe.send()` only when needed. Status widgets compare `_last_state_version` with the current version from shared services, doing full rebuilds only on change.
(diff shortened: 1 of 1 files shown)


docs/developer/design/concept-inventory.md
```diff
@@ -0,0 +1,287 @@
...
+
+A complete inventory of concepts and behaviors a developer needs to understand to reason about the system. Each entry is a topic that would warrant its own documentation section.
+
+---
+
+## A. System-Wide Concepts
+
+### A1. Service Topology and Data Flow
+
+The system is a pipeline of independent processes communicating via Kafka. Upstream (raw instrument data) flows through four processing services (monitor_data, detector_data, data_reduction, timeseries) which publish results to downstream topics. The dashboard consumes downstream data for visualization and publishes commands/ROI definitions back upstream. Each processing service is a separate OS process with its own OrchestratingProcessor.
```
nvaytet: Easier to review/comment with one sentence per line 🙏

nvaytet (commented):
Generally very useful stuff. You mentioned before that every time you tried to make some documentation describing how livedata works, it became out of date a few days later.
Do we have the same issue here?

SimonHeybrock:
> Generally very useful stuff. You mentioned before that every time you tried to make some documentation describing how livedata works, it became out of date a few days later. Do we have the same issue here?

This was mainly meant as an overview of topics we might want to cover in the "seminars" about ESSlivedata, not to be meant for reading linearly. And (I hope) the difference is that while in the past we made detailed architecture docs, here we focus on concepts which (hopefully) change less frequently, or if they do it has to be very deliberate.

SimonHeybrock:
Seems we never got around to do this.

PR #725, opened 2026-02-25
Add rate normalization option (counts/s) to plot display

### LINKED ISSUE #721: Add rate normalization option (counts per second) to plot display
## Summary

Add a user-facing plot option to display quantities (counts, cumulative counts, detector/monitor images) as **rates** (per second), normalized by the time duration derived from `start_time` and `end_time` coordinates already present on all outputs.

## Motivation

Currently, detector and monitor views display raw counts. For comparing across different accumulation windows or understanding event rates, users need a "counts per second" view. This should be a simple toggle, similar to existing plot options like window mode or log scale.

### Where to normalize: the plotter, not the extractor

Rate normalization is a **display concern** (like log scale), not an extraction concern (like window aggregation). Implementing it in the plotter avoids replacing/rebuilding extractors when the user toggles the option on or off. The plotter already receives data with `start_time`/`end_time` coords and uses them for the plot title.

### Time information is already correct for both output types

`StreamProcessorWorkflow.finalize()` sets per-cycle `start_time`/`end_time` on **window outputs** (e.g., `counts_total`, `current`), then resets its internal time tracking. Cumulative outputs (e.g., `counts_total_cumulative`, `cumulative`) don't receive these cycle-specific coords, so `Job._add_time_coords()` fills them in with the **Job's** full-lifetime `start_time` (first-ever data) and `end_time` (latest data).

This means the normalization `data / (end_time - start_time)` naturally does the right thing:

- **Current window**: current counts / current cycle duration = instantaneous rate
- **Cumulative**: total counts / total elapsed time = average rate

### Conditional availability based on output unit

Rate normalization only makes physical sense for count data. The codebase already has a precedent for unit-based conditioning: `WindowAggregatingExtractor`'s auto-aggregation checks `unit == 'counts'` to decide `nansum` vs `nanmean`.

The same principle applies here. Examples:

| Output | Unit | Rate applicable? |
|--------|------|------------------|
| Detector/monitor views (current, cumulative, counts_*) | `counts` | Yes |
| Bifrost spectrum/ratemeter | `counts` | Yes |
| DREAM `i_of_d`, `i_of_d_two_theta` | `dimensionless` | No (already normalized) |
| LOKI `i_of_q`, `i_of_wavelength` | `dimensionless` | No (already normalized) |

Two levels of conditioning:

- **Runtime guard in the plotter**: The normalization helper checks `da.unit == 'counts'` and is a no-op otherwise. This is the safety net — even if the toggle is on, non-count data isn't touched.
- **UI-level visibility via the output template**: The template is already inspectable via `WorkflowSpec.get_output_template()`. When the dashboard builds plot options, it can check the template unit and only show the rate toggle when `template.unit == 'counts'`. This avoids presenting a meaningless toggle for I(Q) outputs.

No new field metadata is needed — the unit is already on every template. If we later need explicit overrides (e.g., a `counts` output that shouldn't offer rate), we could add `json_schema_extra={'rate_normalizable': False}` to the Field definition.

### Implementation sketch

Add a `normalize_to_rate` boolean to `PlotParamsBase` (available to all plot types).

In `Plotter.__call__()`, before delegating to `self.plot()`:

```python
def _normalize_to_rate(da: sc.DataArray) -> sc.DataArray:
    if da.unit != 'counts':
        return da
    if 'start_time' not in da.coords or 'end_time' not in da.coords:
        return da
    duration_s = (da.coords['end_time'] - da.coords['start_time']).to(
        unit='s', dtype='float64'
    )
    if duration_s.value <= 0:
        return da
    return da / duration_s  # unit: counts -> counts/s
```

### Scope

- Applies to all plotter types (image, line, bars, overlay, slicer) via the base class
- Unit changes automatically (`counts` to `counts/s`) through scipp's unit propagation
- Widget auto-generates from the Pydantic model field on `PlotParamsBase`
- Window mode interaction is correct: in window mode with nansum, summed counts / window duration = average rate

## Blocked by

Blocked by #673 (multi-session support), which reorganizes `Plotter` into a two-stage `compute()`/`present()` pattern. Rate normalization would go into the `compute()` stage. The exact API should be settled before implementing this.

SimonHeybrock:
## Summary

- Adds a "Counts Per Second" toggle that normalizes counts data by `(end_time - start_time)` duration, converting counts to counts/s
- Display-level concern (like log scale) — works for both window outputs (instantaneous rate) and cumulative outputs (average rate)
- Added to the four windowed param classes (`PlotParams1d`, `PlotParams2d`, `PlotParams3d`, `PlotParamsBars`), deliberately excluded from `PlotDisplayParams1d/2d` to avoid conflict with correlation histograms' own per-second normalization

## Example

bar plot and timeseries plot with counts/s

## Motivation

Users want to view detector counts as rates normalized by the measurement time window. Since all plotter outputs already carry `start_time`/`end_time` coordinates, the division `data / duration` is straightforward and handles both instantaneous and average rate semantics naturally.

## Test plan

- [x] Unit tests for `_normalize_to_rate` (counts, non-counts, missing coords, zero/negative duration, 2D, 3D)
- [x] Integration tests through `LinePlotter`, `ImagePlotter`, `SlicerPlotter`
- [x] Param model tests verifying field presence on windowed params and absence on display params
- [x] Manual: enable toggle in dashboard and verify counts/s display for a windowed plot
     - Works with bar-chart
     - Works with timeseries
     - Works on 2d plot
     - Works in combination with window-mode

Closes #721

### DIFF
--- a/src/ess/livedata/core/job_manager.py
+++ b/src/ess/livedata/core/job_manager.py
@@ -367,10 +367,11 @@ class JobManager:
         if job_data.is_empty():
             return None
 
-        # Track primary data updates
-        if job_data.is_active():
-            self._jobs_with_primary_data.add(job.job_id)
         reply = job.add(job_data)
+        # Track primary data updates only after successful add, so that a
+        # failed push does not trigger a finalize attempt on empty accumulators.
+        if not reply.has_error and job_data.is_active():
+            self._jobs_with_primary_data.add(job.job_id)
 
         # Track warnings from job operations, or clear them on success
         if reply.has_error and reply.error_message is not None:
--- a/tests/core/job_manager_test.py
+++ b/tests/core/job_manager_test.py
@@ -1347,6 +1347,108 @@ class TestJobManager:
         assert len(results) == 0
 
 
+class TestPushFailureCascade:
+    """Regression tests: a push failure must not cascade into a finalize error."""
+
+    def test_failed_push_does_not_trigger_finalize(
+        self, fake_job_factory, base_workflow_config
+    ):
+        """When push fails on empty accumulators, compute_results must not finalize."""
+        manager = JobManager(fake_job_factory)
+        job_id = manager.schedule_job("test_source", base_workflow_config)
+
+        processor = fake_job_factory.processors[job_id]
+        processor.fail_finalize_when_empty = True
+        processor.should_fail_accumulate = True
+
+        data = WorkflowData(
+            start_time=100,
+            end_time=200,
+            data={StreamId(name="test_source"): sc.scalar(42.0)},
+        )
+        manager.push_data(data)
+
+        # Job should be in warning state from the push failure
+        status = manager.get_job_status(job_id)
+        assert status.state == JobState.warning
+        assert status.warning_message is not None
+        assert "Accumulate failure" in status.warning_message
+
+        # compute_results() must not attempt to finalize the empty job
+        results = manager.compute_results()
+        assert len(results) == 0
+
+        # Job should still be in warning state, not error
+        status = manager.get_job_status(job_id)
+        assert status.state == JobState.warning
+        assert "Accumulate failure" in status.warning_message
+        assert status.error_message is None
+
+    def test_recovery_after_push_failure(self, fake_job_factory, base_workflow_config):
+        """Job recovers to active state when push succeeds after a failure."""
+        manager = JobManager(fake_job_factory)
+        job_id = manager.schedule_job("test_source", base_workflow_config)
+
+        processor = fake_job_factory.processors[job_id]
+        processor.fail_finalize_when_empty = True
+        processor.should_fail_accumulate = True
+
+        data = WorkflowData(
+            start_time=100,
+            end_time=200,
+            data={StreamId(name="test_source"): sc.scalar(42.0)},
+        )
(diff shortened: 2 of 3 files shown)


src/ess/livedata/dashboard/plots.py
```diff
@@ -180,9 +200,13 @@ def __init__(
         ----------
         layout_params:
             Layout parameters for combining multiple datasets. If None, uses defaults.
+        normalize_to_rate:
+            If True, normalize counts data to rate (counts/s) using
+            start_time/end_time coordinates before plotting.
         **kwargs:
             Additional keyword arguments passed to the autoscaler if created.
         """
+        self._normalize_to_rate = normalize_to_rate
```
nvaytet: Did I understand correctly that this is not a setting users can toggle on and off with a button while the plotter is running?
I think they would find it useful if they could.

SimonHeybrock: It is in the plot options modal (opened with a gear icon), alongside other options such as log-scale. Under the hood it is recreating the plot, but that might not be of concern to user.

So no, it is not a tool-button currently. I see they argument you are making, and a similar one can be made about log-scale (etc). If we want to solve this we should probably address it with a unified solution.

PR #735, opened 2026-02-26
Richer 1D plot rendering options

SimonHeybrock:
## Summary

- Add scatter/points mode and error band display to `LinePlotter`, extending 1D plots beyond just line and histogram
- Compose HoloViews elements directly (e.g., `Curve * ErrorBars`) instead of relying on `to_holoviews()` auto-dispatch, giving separate styling control for the base element vs the error overlay
- Fix black-on-black error bars: previously, data with variances was rendered as a bare `hv.ErrorBars` element with no separate line, making everything the same color

## Example

Cases with disabled errors not shown as they should be obvious:

3 curve modes with 2 error modes each

## Motivation

The `LinePlotter` had three issues: error bars were rendered in black with no line underneath (just a bare `hv.ErrorBars`), there was no scatter/points mode for discrete measurements, and there was no way to suppress error bars when data has variances.

## Changes

**New enums**: `Line1dRenderMode` (line, points, histogram) replaces `Curve1dRenderMode`; new `ErrorDisplay` (bars, band, none) controls error visualization.

**New converters**: `convert_scatter_1d()` → `hv.Scatter`, `convert_spread_1d()` → `hv.Spread` in `scipp_to_holoviews.py`.

**Refactored `LinePlotter`**: `plot()` composes elements via match on mode + error display. Sizing opts are applied to child elements individually before overlay composition (Bokeh requires `responsive` on each element, not just the composite overlay).

## Test plan

- [x] Manual: run dashboard, configure a 1D plot, verify line/points/histogram modes render correctly
- [x] Manual: toggle error display between bars/band/none with data that has variances
- [x] Manual: verify plot sizing is correct (responsive, not collapsed) in all error overlay modes


SimonHeybrock (commented):
LGTM

PR #745, opened 2026-03-02
Fix correlation plot config modal showing stale/wrong values and related issues

### LINKED ISSUE #684: Correlation plot config modal issues
Observing two small issues:
- When reconfiguring an existing correlation plot, the correlation axis selectors in step 2 do not show the current value (showing no selection instead).
- In step 3, the `Histogram Bins` params show the source title as, e.g., `XAxisSource`. However, I had an outdated saved config and it ended up displaying an old source name (before we added titles). Example: Displays `monitor2` instead of `monitor_cave`.

SimonHeybrock:
## Summary

Fixes #684 and several related issues in the correlation histogram configuration modal and plotter.

- **Wrong field labels in step 3**: `ParamWidget` used `snake_to_camel(field_name)` for display names, ignoring Pydantic `title` metadata. Fields like `x_axis_source` showed as "XAxisSource" instead of "X Axis".
- **Stale axis source names in step 3**: Edit mode loaded saved params as-is, preserving old source names (e.g. `monitor2`) instead of showing current titles. Now both edit and new modes inject resolved human-readable titles via `instrument_config.get_source_title()`.
- **Raw source names in step 2 axis selector**: The correlation axis dropdown showed raw Kafka source names. Now resolves to human-readable titles via instrument config.
- **Step 3 not updating after axis change**: Navigating back from step 3 to step 2, changing the correlation axis, and returning to step 3 showed stale values because `axis_sources` was missing from the change-detection check.
- **Correlation histogram plot legend**: `CorrelationHistogramPlotter` was not forwarding the `source_title` callable to its renderer, so plot legends showed raw source names instead of titles (regression from fce25e66).

Also adds a parameterized registry test that verifies the `compute()` signature of all 12 registered plotters, preventing similar signature mismatches.

## Test plan

- [x] Open reconfigure modal on an existing correlation plot — axis selectors (step 2) should show current selection with human-readable titles
- [x] Histogram Bins (step 3) should show "X Axis" / "Y Axis" labels and resolved source titles
- [x] Change the correlation axis in step 2, return to step 3 — should reflect the new axis
- [x] Correlation histogram plot legend should show source titles, not raw names
- [x] Creating a new correlation plot still works correctly

### DIFF
--- a/src/ess/livedata/config/instruments/loki/specs.py
+++ b/src/ess/livedata/config/instruments/loki/specs.py
@@ -74,12 +74,12 @@ class BeamCenterXY(pydantic.BaseModel):
 class LokiAuxSources(AuxSourcesBase):
     """Auxiliary source names for LOKI SANS workflows."""
 
-    incident_monitor: Literal['beam_monitor_1'] = pydantic.Field(
-        default='beam_monitor_1',
+    incident_monitor: Literal['beam_monitor_m1'] = pydantic.Field(
+        default='beam_monitor_m1',
         description='Incident beam monitor for normalization.',
     )
-    transmission_monitor: Literal['beam_monitor_3'] = pydantic.Field(
-        default='beam_monitor_3',
+    transmission_monitor: Literal['beam_monitor_m3'] = pydantic.Field(
+        default='beam_monitor_m3',
         description='Transmission monitor for sample transmission calculation.',
     )
 
@@ -161,11 +161,11 @@ instrument = Instrument(
     name='loki',
     detector_names=detector_names,
     monitors=[
-        'beam_monitor_0',
-        'beam_monitor_1',
-        'beam_monitor_2',
-        'beam_monitor_3',
-        'beam_monitor_4',
+        'beam_monitor_m0',
+        'beam_monitor_m1',
+        'beam_monitor_m2',
+        'beam_monitor_m3',
+        'beam_monitor_m4',
     ],
     source_metadata={
         'loki_detector_0': SourceMetadata(title='Rear'),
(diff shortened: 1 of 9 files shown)


src/ess/livedata/dashboard/widgets/plot_config_modal.py
```diff
@@ -62,29 +62,51 @@
...
+    role_to_field = {X_AXIS: 'x_axis_source', Y_AXIS: 'y_axis_source'}
+    result: dict[str, str] = {}
+    for role, field_name in role_to_field.items():
+        if role in axis_sources and axis_sources[role].source_names:
+            source_name = axis_sources[role].source_names[0]
+            result[field_name] = instrument_config.get_source_title(source_name)
+    return result
+
+
 def _inject_axis_source_names(
```
jl-wynen:
```suggestion
def _inject_axis_source_titles(
```
judging by the change in docstring.

tests/dashboard/widgets/plot_config_modal_test.py
```diff
@@ -96,5 +102,88 @@ class NoBinsParams(pydantic.BaseModel):
...
+        name="test_wf",
+        version=1,
+        title=title,
+        description="",
+        params=None,
+        outputs=outputs,
+    )
+
+
+class TestBuildTimeseriesOptions:
```
jl-wynen: These are tests for implementation details. Do we really need them?

jl-wynen (commented):
Next time, can you split replies to reviews into multiple commits? It's a little difficult to see what change goes with what comment.

tests/dashboard/widgets/plot_config_modal_test.py
jl-wynen: Splitting logic from presentation (widget) seems like a good idea in general.

PR #752, opened 2026-03-04
Remove MultiConsumer and upstream/downstream Kafka split

SimonHeybrock:
## Summary

- Replaces the `kafka_upstream`/`kafka_downstream` config split with a single unified `kafka` config, removing the never-used two-cluster setup and `KAFKA2_*` environment variables
- Removes `MultiConsumer`, which wrapped separate control and data consumers into one interface that polled sequentially anyway
- Removes `make_control_consumer()` — the commands topic is now subscribed to alongside data topics via the adapter's topic list

## Motivation

These three abstractions were intertwined and added complexity without practical benefit. The upstream/downstream Kafka distinction was never used in practice (both always pointed to the same broker), and `MultiConsumer` existed solely to combine the control consumer with data consumers.

## Test plan

- [x] CI passes
- [x] Manually verify services start and connect to Kafka correctly with `--dev`

### DIFF
--- a/src/ess/livedata/config/workflow_spec.py
+++ b/src/ess/livedata/config/workflow_spec.py
@@ -218,6 +218,25 @@ class WorkflowSpec(BaseModel):
             version=self.version,
         )
 
+    def get_output_title(self, output_name: str) -> str:
+        """Get human-readable title for an output field name.
+
+        Parameters
+        ----------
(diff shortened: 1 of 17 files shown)


SimonHeybrock:
@MridulS I think this will require changes to the deployment setup.

PR #755, opened 2026-03-04
Add workflow visualization utility

SimonHeybrock:
## Summary

- Expose `StreamProcessorWorkflow.visualize()` wrapping `StreamProcessor.visualize()`, so callers don't need to reach through private attributes
- Add `esslivedata-workflows` CLI entry point that renders workflow graphs for a given instrument; supports `--list`, `--workflow ID`, `--format`, and `--output-dir`

## Motivation

`ess.reduce.streaming.StreamProcessor.visualize()` produces color-coded dependency graphs showing dynamic keys, accumulators, static nodes, and targets. This was not accessible through our wrapper layers. The CLI makes it easy to generate these graphs for documentation or inspection without writing Python.


SimonHeybrock:
LGTM

PR #766, opened 2026-03-06
Move pixel grouping into shared preprocessor

SimonHeybrock:
## Motivation

When multiple workflows consume the same detector source, each independently runs `group_event_data()` — a costly operation that groups flat events by pixel ID. This duplicates work proportional to the number of downstream workflows per source.

Moving pixel grouping into the shared preprocessor (`GroupByPixel` wrapping `ToNXevent_data`) ensures it runs **once per batch per source**, regardless of how many workflows consume that source.

This can be disabled using the new `--no-group-by-pixel` cmdline arg.

## Approach

`GroupByPixel` is a decorator around `ToNXevent_data`. It calls `group_event_data()` in `get()`, producing events binned by flat 1D `detector_number`. Downstream workflows receive this as `NeXusData[NXdetector, SampleRun]` and run `assemble_detector_data` normally — which calls `group_event_data` again, now made idempotent via a monkey-patch.

The idempotent `group_event_data` detects already-grouped input (by the presence of a `detector_number` coordinate) and skips re-grouping, but still folds to the target multi-dimensional shape from `EmptyDetector`. This means each workflow applies its own calibration (e.g., LOKI's per-run `BeamCenter` offset) correctly.

Applied to both the **detector data** and **data reduction** services.

### Key design decisions

- **Flat 1D output from preprocessor**: `GroupByPixel` always flattens `detector_number` before grouping, producing consistent 1D output regardless of instrument detector shape. Folding to multi-dim shape happens in each workflow's `assemble_detector_data`.
- **Idempotent `group_event_data`** (monkey-patch in `_patch_group_event_data.py`): Temporary measure until `ess.reduce` is updated upstream. Detects already-grouped data and applies only the fold step.
- **No coupling to calibration**: The preprocessor does not apply geometry, masks, or position offsets. Workflows retain full control over calibration via `EmptyDetector`.

## Trade-offs

**Preprocessing runs unconditionally.** The preprocessor runs for every source that has messages, even if no workflow/job has been started yet. Previously this was cheap (`ToNXevent_data` just concatenates arrays), but `GroupByPixel` adds grouping cost. In practice the window where this matters is narrow — preprocessors are created lazily on first message, and jobs are typically started at or shortly after service startup.

**Interaction with parallel job execution (#763).** By moving work from per-job into the preprocessor, we shift O(N_events) work into the sequential preprocessing path. If parallel job execution is added, preprocessing could become the serial bottleneck. A future improvement could parallelize `preprocess_messages` across sources.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,9 @@ COPY . .
 
 RUN pip install --no-cache-dir -e ".[${INSTRUMENT}]"
 
+ENV LIVEDATA_DATA_DIR=/app/data/geometry
+RUN python -m ess.livedata.handlers.download_geometry
+
 USER livedata
 
 ENV LIVEDATA_ENV=docker \
--- a/docs/user-guide/adding-new-instruments.md
+++ b/docs/user-guide/adding-new-instruments.md
@@ -329,11 +329,17 @@ If you configure `detector_number` explicitly via `configure_detector()`, no geo
 
 To provide a geometry file:
 
-1. Create a NeXus geometry file following the naming convention: `geometry-<instrument>-<date>.nxs`
-   - Use `ess-livedata-make-geometry-nexus` to create from a regular NeXus file
-   - The date should be the first date the geometry file is used in production
-2. Add the file's MD5 hash to the `_registry` in [detector_data_handler.py](../../src/ess/livedata/handlers/detector_data_handler.py)
-3. Upload the file to https://public.esss.dk/groups/scipp/beamlime/geometry/
+1. Create a stripped NeXus geometry file from a full NeXus file:
+   ```sh
+   ess-livedata-make-geometry-nexus input.nxs geometry-<instrument>-<YYYY-MM-DD>.nxs
+   ```
+   The date should be the first date the geometry is valid for production use.
+2. Test the output file: load it, verify detector views and workflows produce correct results.
+3. Upload the verified file and get the registry entry:
+   ```sh
+   python -m ess.livedata.handlers.upload_geometry geometry-<instrument>-<YYYY-MM-DD>.nxs
+   ```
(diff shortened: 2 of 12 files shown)


src/ess/livedata/handlers/group_by_pixel.py
```diff
@@ -0,0 +1,51 @@
...
+    ----------
+    inner:
+        The underlying event accumulator.
+    detector_number:
+        Detector pixel numbers used for grouping.
+    """
+
+    def __init__(self, inner: ToNXevent_data, detector_number: sc.Variable) -> None:
+        self._inner = inner
+        # Always use flat 1D detector_number for grouping. The downstream
```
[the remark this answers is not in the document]
SimonHeybrock: I think it is cheap, yes. If we wanted to do it here we'd need to inject the shape info of the given instrument somehow.

src/ess/livedata/services/data_reduction.py
```diff
@@ -13,7 +13,11 @@
 
 
 def make_reduction_service_builder(
-    *, instrument: str, dev: bool = True, log_level: int = logging.INFO
+    *,
+    instrument: str,
+    dev: bool = True,
+    log_level: int = logging.INFO,
+    group_by_pixel: bool = True,
```
nvaytet: A little annoying that we have to have this pretty specific argument here.
Is there an alternative?
What are the cases when we don't want to group by pixel? Can we just always group instead of having to pass the bool arg around?

SimonHeybrock: We could, but I am not sure yet it is beneficial for all instruments. Also, it made benchmarking easier as there was no need to change the code to do A/B comparisons. I don't see the arg as a problem, it joins the group of `--sync-scheduler` and `job-threads`, both of which affect performance in some way and we may need different combinations for different instruments.

PR #785, opened 2026-03-12
Replace AuxSourcesBase with declarative AuxSources

### LINKED ISSUE #694: Aux source display names bypass SourceMetadata titles
## Context

`SourceMetadata` (introduced Jan 2026, `9fb104e6`) added human-readable titles for source names, used in source selectors, plot toolbars, and status widgets via `Instrument.get_source_title()`. However, `AuxSourcesBase` (introduced Oct 2025, `75ccc48c`) predates this and uses a separate mechanism — pydantic Field metadata — that never got connected to the source title system.

## Problem

When the dashboard renders auxiliary source configuration (e.g., LOKI's incident/transmission monitors), users see raw stream names like `monitor1` instead of titles like "Incident Monitor". Two independent gaps cause this:

**Gap A — Widget labels ignore `field_info.title`**: `ParamWidget._create_widget_for_field()` always uses `snake_to_camel(field_name)` for the widget label (e.g., `incident_monitor` → "IncidentMonitor"). It never checks `field_info.title`. This is inconsistent with `ModelWidget`, `PlotConfigModal`, `WorkflowStatusWidget`, and `PlotWidgets`, which all respect `field_info.title`.

**Gap B — Dropdown options show raw stream names**: For `Literal['monitor1']` fields, `ParamWidget` builds options as `{str(val): val}` → `{'monitor1': 'monitor1'}`. There is no mechanism to look up the human-readable title from `SourceMetadata`. The `ConfigurationAdapter.get_source_title()` method exists but is only called by the source selector (MultiChoice for detectors), not by the aux sources widget.

It should be noted that for the *source names* (as opposed to axus source names) the UI does indeed correctly display the titles.

## Affected instruments

All instruments with aux sources are affected:
- **LOKI**: `incident_monitor` (monitor1), `transmission_monitor` (monitor2) — titles defined in SourceMetadata but not displayed
- **DREAM**: `cave_monitor` (monitor_cave)
- **Bifrost**: `detector_rotation`, `sample_rotation`
- **DetectorROIAuxSources**: ROI stream (different pattern — uses `render()` for job-specific names)

## Design tension

The aux source fields serve a dual purpose:
- **Routing**: Field values are stream names used by the backend for message routing (via `AuxSourcesBase.render()`)
- **Display**: Field names and values appear in the configuration UI

`SourceMetadata` was designed for the source selector (detector/monitor MultiChoice) where internal names map to display titles. The aux sources widget is a generic pydantic model renderer (`ParamWidget`) with no access to `Instrument` configuration.

## Possible approaches

These are starting points for discussion, not prescriptions:

- **Minimal fix**: Make `ParamWidget` respect `field_info.title` (fixes Gap A only). Add `title=` to aux source Field definitions. This doesn't fix the dropdown option labels (Gap B).
- **Pass source metadata to ParamWidget**: Thread `Instrument` or a title-lookup callable through `ConfigurationAdapter` → `ConfigurationWidget` → `ParamWidget` so Literal option labels can use `SourceMetadata` titles. This is more invasive but addresses both gaps.
- **Use Field `json_schema_extra` or similar**: Encode display titles in the aux source model itself, removing the need to look up `SourceMetadata` at render time. But this duplicates title information.

SimonHeybrock:
Closes #694

## Summary

- Replace `AuxSourcesBase` (Pydantic model class hierarchy) with a declarative `AuxSources` spec holding a dict of `AuxInput` dataclasses. This separates the specification (available choices, defaults, metadata) from user selections (`WorkflowConfig.aux_source_names: dict[str, str]`, unchanged).
- Close the information flow gap: resolved aux source names are now passed into workflow factory functions via signature introspection, so factories can set `NeXusMonitorName`/`NeXusName` from a single source of truth instead of relying on hardcoded defaults in upstream science workflows.
- Replace the generic `ParamWidget`-based aux source rendering with a purpose-built `AuxSourcesWidget` that uses `AuxInput.title` for widget labels and `SourceMetadata` titles for dropdown option values, fixing both display gaps from the issue.

## Motivation

`AuxSourcesBase` predated `SourceMetadata` and used a parallel naming system (Pydantic Field metadata) that was never connected to the source title system. This caused raw stream names to appear in the UI. More importantly, resolved aux source names were never passed to the workflow factory — LOKI only worked because `LokiWorkflow()` happened to hardcode matching defaults in two unconnected places.

## Design decisions

- `AuxInput.title`/`.description` describe the **workflow role** (what the input is used for), complementing `SourceMetadata` which describes the **physical stream** (the component that generates it). Both are needed and serve different purposes.
- Primary and aux sources remain separate: primary sources determine which Jobs exist (map-over axis), while aux sources are injected into every Job. Unifying them would relocate complexity without removing it.
- Bifrost factories are not changed — their aux sources (rotation angles) are raw f144 scalars injected via `context_keys`, not loaded from NeXus files.

## Upstream

Opened scipp/ess#337 to update stale hardcoded LOKI monitor names (`beam_monitor_1` → `beam_monitor_m1`) in `LokiWorkflow()` defaults.

## Test plan

- [x] Manual check: LOKI/DREAM/Bifrost aux source widgets show human-readable titles
- [x] Manual check: aux source selections propagate to the workflow (monitor geometry)

### DIFF
--- a/src/ess/livedata/config/defaults/raw_data_consumer.yaml
+++ b/src/ess/livedata/config/defaults/raw_data_consumer.yaml
@@ -2,5 +2,6 @@ auto.offset.reset: latest
 enable.auto.commit: true
 fetch.min.bytes: 1
 fetch.max.bytes: 200000000  # might need at least 10x this in RAM for processing
+fetch.wait.max.ms: 200  # reduce broker-side wait to avoid head-of-line blocking on idle partitions
 session.timeout.ms: 30000
 heartbeat.interval.ms: 10000  # long so we do not timeout when processing large data
--- a/src/ess/livedata/config/defaults/reduced_data_consumer.yaml
+++ b/src/ess/livedata/config/defaults/reduced_data_consumer.yaml
@@ -1,6 +1,7 @@
 auto.offset.reset: latest
 enable.auto.commit: true
 fetch.min.bytes: 1
+fetch.wait.max.ms: 200  # reduce broker-side wait to avoid head-of-line blocking on idle partitions
 session.timeout.ms: 6000
 heartbeat.interval.ms: 2000
 max.partition.fetch.bytes: 104857600
--- a/src/ess/livedata/config/instruments/dream/factories.py
+++ b/src/ess/livedata/config/instruments/dream/factories.py
@@ -63,18 +63,23 @@ def setup_factories(instrument: Instrument) -> None:
                 resolution={'arc_length': 80, 'z': 320},
                 pixel_noise=_pixel_noise,
             ),
+            # Backward/upstream detectors: no flip needed, x already points
+            # right when viewed from the sample looking back.
             'endcap_backward_detector': GeometricViewConfig(
                 projection_type='xy_plane',
                 resolution={'y': 240, 'x': 160},
                 pixel_noise=_pixel_noise,
-                flip_x=True,
+                flip_x=False,
             ),
-            'endcap_forward_detector': GeometricViewConfig(
+            'high_resolution_detector': GeometricViewConfig(
                 projection_type='xy_plane',
                 resolution={'y': 160, 'x': 160},
                 pixel_noise=_pixel_noise,
+                flip_x=False,
             ),
-            'high_resolution_detector': GeometricViewConfig(
+            # Forward/downstream detectors: flip_x because x points left
+            # when viewed from the sample along the beam.
+            'endcap_forward_detector': GeometricViewConfig(
                 projection_type='xy_plane',
                 resolution={'y': 160, 'x': 160},
                 pixel_noise=_pixel_noise,
@@ -84,6 +89,7 @@ def setup_factories(instrument: Instrument) -> None:
                 projection_type='xy_plane',
                 resolution={'y': 160, 'x': 160},
                 pixel_noise=_pixel_noise,
+                flip_x=True,
             ),
         },
     )


SimonHeybrock:
## Architectural review: is 29 files a concern?

The aux source concept is inherently cross-cutting — it lives in:

| Layer | Role |
|-------|------|
| **Spec** (instrument specs) | Declares available inputs + defaults |
| **Job/Factory** | Renders selections into stream names for data routing |
| **Dashboard adapter** | Bridges spec ↔ UI |
| **Widget** | Presents dropdowns to the user |

Each layer legitimately needs to know about aux sources. That's **domain-inherent coupling**, not an architectural smell.

The refactoring actually *improved* the architecture:

- **Before**: Pydantic model *types* (not instances) flowed through the entire stack. Every layer called Pydantic-specific APIs (`model_dump()`, `model_fields`, `Field(...)`). That's tight coupling to a serialization framework as a cross-layer contract.
- **After**: Plain dicts and a simple `AuxSources` container flow through. The layers are coupled to a thin domain abstraction, not a framework.
- **Boilerplate dropped ~50-70%** in instrument specs — no more class-per-instrument with `Literal` field gymnastics.
- **Dependency flow is clean and unidirectional**: `WorkflowSpec` → `JobFactory` → `ConfigurationAdapter` → `Widget`. No cycles.

The factory signature introspection pattern (opt-in `aux_source_names` parameter) is also well-designed — it avoids forcing a uniform interface on workflows that handle aux sources differently (LOKI/DREAM configure NeXus name keys, Bifrost routes via context_keys).

29 files is a lot, but roughly half are tests (which *should* change when the contract changes). The production changes span 4 well-defined layers that all legitimately interact with aux sources. The old design made this worse by coupling every layer to Pydantic internals. The new design is simpler, more declarative, and the coupling that remains is to a thin domain concept — which is appropriate.

PR #819, opened 2026-03-19
Skip redundant deepcopy in NoCopyWindowAccumulator used by detector view workflows

SimonHeybrock:
## Summary

This change optimizes a component used in all "Detector View" workflows. Even for data volumes that are small enough to handle within an update cycle this change is useful: It will improve latency, reducing lag experienced by users.

`NoCopyWindowAccumulator` deepcopied the input histogram on every push cycle. Since `on_finalize` clears the value, every push was effectively a "first push" triggering `EternalAccumulator`'s `deepcopy(value)`. This copy is redundant because the paired Cumulative accumulator (`NoCopyAccumulator`) deepcopies on its first push, isolating its buffer — no consumer ever mutates the shared input.

Overrides `_do_push` to store the value by reference on first push, eliminating ~25–29% of the per-cycle histogram pipeline cost for non-trivial view sizes:

| View | Bins | Buffer size | deepcopy cost | % of pipeline total |
|------|------|------------|--------------|-------------------|
| 100×100 | 1000 | 38 MB | 7.8 ms | 27% |
| 200×200 | 2000 | 305 MB | 38.7 ms | 25% |
| 500×500 | 2000 | 1.9 GB | 249.8 ms | 29% |
| 500×500 | 5000 | 4.7 GB | 558.9 ms | 28% |

Safety relies on `+=` only mutating the left operand, never the right — so evaluation order and push count don't matter. A docstring warning documents the constraint that at most one `NoCopyWindowAccumulator` may subscribe to a given shared input.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -57,6 +57,26 @@ Note that `dynamic=True` only gates Bokeh model creation. Python-side periodic c
 (e.g., `SessionUpdater` custom handlers) still run for all registered widgets regardless
 of which tab is visible. Use an `is_visible` predicate to skip refresh work for hidden tabs.
 
+## Colors and styling
+
+All colors must come from `dashboard/widgets/styles.py`. Do not hard-code hex color values
+or rgba strings in widget files. The shared module provides:
+
+- `StatusColors` — semantic status indicators (ERROR, SUCCESS, WARNING, etc.)
+- `HoverColors` — translucent hover backgrounds derived from StatusColors
+- `Colors` — neutral palette (BORDER, BG_LIGHT, TEXT, TEXT_MUTED, etc.)
+- `ErrorBox` / `WarningBox` — alert box color sets (BG, BORDER, TEXT)
+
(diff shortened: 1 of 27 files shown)


src/ess/livedata/handlers/accumulators.py
```diff
@@ -102,8 +102,26 @@ class NoCopyWindowAccumulator(NoCopyAccumulator):
...
+    safe because the paired Cumulative accumulator (NoCopyAccumulator) deepcopies on
+    its first push, isolating its buffer. Since ``+=`` only mutates the left operand
+    (``self._value``), never the right (the input), no consumer ever mutates the shared
+    input histogram — regardless of push count or accumulator evaluation order.
+
+    .. warning::
+
+       At most one ``NoCopyWindowAccumulator`` may subscribe to a given shared input.
+       If two instances received the same object, the first one's ``+=`` on a subsequent
+       push would mutate the reference held by the second, silently corrupting its data.
```
jl-wynen: Is there a way to enforce this?

PR #821, opened 2026-03-19
Add adaptive message batching to reduce overhead under load

SimonHeybrock:
## Summary

Under sustained load, fixed per-batch overhead (workflow graph evaluation, serialization, thread dispatch) dominates cycle time. `AdaptiveMessageBatcher` wraps `SimpleMessageBatcher` and widens the batch window when consecutive batches signal overload, then cautiously de-escalates when headroom is available. This is lossless: the dashboard updates less frequently but no data is dropped.

This might make #739 redundant, or at least less urgent.

## Design

Uses a **Multiplicative Increase, Additive Decrease (MIAD)** strategy inspired by TCP congestion control:

- **Escalation** (+2 half-steps = x2 window): Triggered after 2 consecutive batches where processing time exceeds the batch window. Fast response to overload.
- **De-escalation** (-1 half-step = x1/√2 window): Triggered after 3 consecutive batches with >25% headroom. The asymmetric step sizes mean two de-escalation steps undo one escalation, providing natural damping.
- **Idle de-escalation**: Wall-clock fallback after 3 idle windows with no data.
- **Batch window grid**: `base × √2^n` (1.0s → 1.41s → 2.0s → 2.83s → 4.0s → ...), capped at `base × 2^max_level`. Fixed grid avoids floating-point drift.

Key properties:
- Idle poll cycles between batches do not reset the consecutive counters, which is essential for de-escalation under continuous light load (e.g., cosmic background after shutter close).
- Empty time-gap batches (`message_count=0`) are treated as no-ops and don't interfere with escalation or de-escalation.
- `batch_interval_s` flows through `ServiceStatus` → x5f2 → `BackendStatusWidget`.

## Scenario tests

Comprehensive simulation-based tests in `adaptive_batching_scenarios_test.py` drive a `MessageBatcher` through realistic load patterns and assert on observable properties (escalation time, backlog bounds, oscillation count, de-escalation). All acceptance thresholds are collected in a central `LIMITS` table for tuning.

**These tests are very long, therefore I created plots for each of them, shown in a comment below. I recommend reviewing those instead of the testcases.**

Scenarios covered: step-function overload at four severity levels, no-escalation under light load (20-85% utilization), GC jitter resilience, boundary oscillation, creeping overload, de-escalation under continuous load (light, moderate, multi-level, partial), shutter open/close with cosmic background, backlog draining, and the 70-100% utilization dead zone.

## Test plan

- [x] Verify adaptive batching behavior on a live dashboard under load

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -57,6 +57,26 @@ Note that `dynamic=True` only gates Bokeh model creation. Python-side periodic c
 (e.g., `SessionUpdater` custom handlers) still run for all registered widgets regardless
 of which tab is visible. Use an `is_visible` predicate to skip refresh work for hidden tabs.
 
+## Colors and styling
+
+All colors must come from `dashboard/widgets/styles.py`. Do not hard-code hex color values
+or rgba strings in widget files. The shared module provides:
+
+- `StatusColors` — semantic status indicators (ERROR, SUCCESS, WARNING, etc.)
+- `HoverColors` — translucent hover backgrounds derived from StatusColors
+- `Colors` — neutral palette (BORDER, BG_LIGHT, TEXT, TEXT_MUTED, etc.)
+- `ErrorBox` / `WarningBox` — alert box color sets (BG, BORDER, TEXT)
+
+`ButtonStyles` in `buttons.py` re-exports commonly used color+hover pairs
+(e.g., `DANGER_RED`/`DANGER_HOVER`, `PRIMARY_BLUE`/`PRIMARY_HOVER`).
+
+Widget-specific decorative colors (e.g., output chip colors, grid preview cell colors)
+that are not shared across widgets may stay local.
+
+Panel does not support CSS custom properties (`var()`) in `styles=` dicts or inline
+HTML `style=` attributes — only in `stylesheets=` parameters. This is why we use
+Python constants rather than CSS variables for centralized color management.
+
 ## Avoiding flicker
 
 Make sure all widget-updates that touch more than a single widget (or a single widget multiple times) use `pn.io.hold()`.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,9 @@
 name: livedata-kafka
+networks:
+  default:
+    ipam:
+      config:
+        - subnet: "10.200.0.0/24"
 services:
   kafka:
     image: confluentinc/cp-kafka:latest
--- a/src/ess/livedata/config/defaults/raw_data_consumer.yaml
+++ b/src/ess/livedata/config/defaults/raw_data_consumer.yaml
@@ -2,5 +2,6 @@ auto.offset.reset: latest
 enable.auto.commit: true
 fetch.min.bytes: 1
 fetch.max.bytes: 200000000  # might need at least 10x this in RAM for processing
+fetch.wait.max.ms: 200  # reduce broker-side wait to avoid head-of-line blocking on idle partitions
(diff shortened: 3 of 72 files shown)


SimonHeybrock:
## Scenario timeline visualizations

Script to generate timeline plots for all adaptive batching scenario tests:
https://gist.github.com/SimonHeybrock/53d183c337a74282493d59cfefbaebd1

Run with `python plot_adaptive_batching_scenarios.py` from the repo root (after `pip install -e ".[test]"`). Generates 31 PNG plots + an `index.html` for browsing.

Each plot shows 4 panels:
- **Level** — batch level over time with assertion bounds
- **Processing time** — dots colored by utilization class (green < 75%, yellow 75-100% dead zone, red > 100% overloaded) vs batch window
- **Utilization** — ratio with headroom/dead-zone/overloaded bands
- **Backlog** — accumulation with limit annotations

SimonHeybrock:
## Scenario timeline plots

Generated from the scenario test suite. Each plot shows 4 panels: batch level, processing time vs window, utilization ratio, and backlog.

### Step-function escalation

step_escalation_time
moderate_overload_step
overhead_dominated_step
severity_severe
severity_extreme

### Non-default base batch length

non_default_base

### No escalation when not needed

light_load_20
light_load_60
light_load_80
light_load_85
gc_jitter

### Steady overload

steady_moderate_overload
boundary_jitter

### Creeping overload

creeping_overload
mild_creeping_overload

### De-escalation

deescalation_to_idle
deescalation_to_light
deescalation_moderate
multi_level_deescalation
partial_deescalation

### Realistic shutter scenarios

shutter_open_close
repeated_shutter_cycles
severe_to_cosmic

### Processing-time awareness

fast_escalation_clear_overload

### Dead zone (known limitation)

dead_zone_stuck

jl-wynen:
> - Number of neutron events in the current stream

This doesn't get better with increasing the time window for batches as both our equations show

> - Random stuff that might make processing slower for a short time, such as GC.
> - Network and Kafka, e.g., if broker was busy and we then suddenly get a larger batch of messages

Do these last longer than 1s? They should be short fluctuations that the adaptive batcher should ignore.

So we only have
> - Number of running workflows

So wouldn't it be enough to use something simple like `batch_time = A * n_workflows` with a factor `A` that we determine with benchmarks?

On a separate note, I think you are seeing the limitations of a monolithic application. It becomes increasingly difficult to scale as you add features. Did you reconsider micro services?

SimonHeybrock:
> > * Number of neutron events in the current stream
> 
> This doesn't get better with increasing the time window for batches as both our equations show

I did not say that. I said it influences the load. Decreasing the repeat rate of the constant cost will reduce the overall performance, making processing more events possible.

> > * Random stuff that might make processing slower for a short time, such as GC.
> > * Network and Kafka, e.g., if broker was busy and we then suddenly get a larger batch of messages
> 
> Do these last longer than 1s? They should be short fluctuations that the adaptive batcher should ignore.

It does not matter if it is longer than 1s. If it takes 100ms it eats into the 1 second budget we have to process a batch.

> So we only have
> 
> > * Number of running workflows
> 
> So wouldn't it be enough to use something simple like `batch_time = A * n_workflows` with a factor `A` that we determine with benchmarks?

No, because we (1) do not know `n_workflows` ahead of time, (2) it still depends on the events rate.

> On a separate note, I think you are seeing the limitations of a monolithic application. It becomes increasingly difficult to scale as you add features. Did you reconsider micro services?

Umm, I don't understand. Let us have a call.

PR #822, opened 2026-03-20
Fix DREAM detector view flip_x to show detectors as seen from sample

SimonHeybrock:
## Summary

- The `flip_x` values for DREAM detector views were inverted: backward-facing detectors (`endcap_backward`, `high_resolution`) had `flip_x=True` while forward-facing detectors (`endcap_forward`, `sans`) had `flip_x=False`
- In the ESS coordinate system, x points left when looking along the beam from the sample, so downstream/forward detectors need flipping while upstream/backward detectors do not
- Reordered and commented the config entries to make the convention clear and prevent future confusion

image

## Test plan

- [x] Visual verification on DREAM dashboard that detector images match expected orientation

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -57,6 +57,26 @@ Note that `dynamic=True` only gates Bokeh model creation. Python-side periodic c
 (e.g., `SessionUpdater` custom handlers) still run for all registered widgets regardless
 of which tab is visible. Use an `is_visible` predicate to skip refresh work for hidden tabs.
 
+## Colors and styling
+
+All colors must come from `dashboard/widgets/styles.py`. Do not hard-code hex color values
+or rgba strings in widget files. The shared module provides:
+
+- `StatusColors` — semantic status indicators (ERROR, SUCCESS, WARNING, etc.)
+- `HoverColors` — translucent hover backgrounds derived from StatusColors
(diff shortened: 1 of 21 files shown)


jl-wynen:
If you do not rotate the coord system, then the endcaps have the same x values, not different signs. You are flipping the x axis in the backwards plot. That doesn't seem natural to me. You can flip the display to get the view from the sample, but the coord should not change. Otherwise, relating the coord value to the physical setup is more complicated.

SimonHeybrock:
> I don't care which one is flipped. What I care about is that all plots make sense with as little mental gymnastics as possible. If we look from the sample and in the forward plot, the x axis point to the right, I would expect it to point to the left in the backwards plots because that is what I would see when I turn around 180deg. Or, if you want x to always point right from the point of view, then it should always do that, including in the mantle plot.

Sure that would make sense, but it is unrelated to this PR. This is a discussion to be had with the scientists, so we either (A) plot axes with decreasing coordinates in some cases, or (B) rename dimensions/axes to UP and RIGHT.

THIS pull request makes the plots consistent. Before this PR all but the mantle detector were flipped w.r.t. how they actually appear as seen from the sample. This changes makes this consistent. The problem you worry about - `x` in the plots not corresponding to the`x` of the instrument coord system existing *before* this PR, here we just change which panels are wrong in this sense.

Issue #828, opened 2026-03-23
Automatic accumulator reset on position context change


SimonHeybrock:
## Context

PR #827 adds infrastructure for monitor and detector view workflows to receive dynamic position data (f144 streams) via aux_sources and context_keys. However, when a component moves, the accumulated histogram becomes physically meaningless — it sums data from different spatial configurations.

Currently, users must reset manually via the UI. This is acceptable because position changes mid-measurement are rare, but an automatic mechanism would be better.

## Problem

When a context key (e.g., monitor position) changes value during a measurement, accumulated data in view workflows should be invalidated. Reduction workflows should NOT be reset, as they accumulate across positions into a shared output space (e.g., Q-space).

A naive equality check (`old_value != new_value`) does not work because position sensors have readout noise — consecutive readings are never exactly equal even when the component hasn't moved.

## Design options

Each has trade-offs; none is obviously the only correct choice:

**A. Preprocessor-level filtering**: The `ToNXlog` preprocessor (or a wrapper) detects significant changes and emits a reset sentinel. The change detection logic lives close to the raw data and can apply instrument-specific thresholds or dead-band filtering. Downstream components see only "meaningful" changes.

**B. Callable predicate on StreamProcessorWorkflow**: Accept a `should_reset(prev_context, curr_context) -> bool` callback. The instrument-specific factory provides the function with appropriate thresholds. The workflow calls it before `set_context()` and clears accumulators if it returns True. Simple and explicit, but couples reset logic to an opaque callback.

**C. Accumulator coord-based detection**: The workflow sets a coordinate (e.g., position value) on the intermediate data being accumulated. The accumulator detects a coord mismatch between its stored state and the new data, and clears itself. This is the most self-contained approach — the accumulator "knows" when its data is incompatible — but requires threading position metadata through the sciline pipeline to the accumulator level.

## Considerations

- Readout noise: position sensors report slightly different values even when stationary. Any solution needs a threshold or dead-band mechanism.
- Per-key granularity: a detector view might have both ROI context keys (should NOT trigger reset) and position context keys (should trigger reset).
- Instrument-specific semantics: what constitutes a "real" position change depends on the instrument (threshold, discrete positions, commanded vs actual setpoint).
- Timing: if a position change arrives mid-window, should the partial window be discarded or finalized first?

## Related

- PR #827: infrastructure for f144 aux sources in monitor/detector view workflows
- Issue #780: LOKI beam_monitor_m4 NXlog transform crash

[...]

PR #846, opened 2026-04-07
LOKI: dynamic rear-detector geometry from f144 carriage stream

### LINKED ISSUE #845: Dynamic geometry for detector view workflows (LOKI rear-detector carriage)
## Context

PR #827 added infrastructure for detector view workflows to receive f144 log streams via `aux_sources`. However, the existing detector view path (`DetectorViewFactory` + `GeometricViewConfig` + `NeXusDetectorSource`) is built around **static** NeXus geometry: pixel positions are read once at startup and `project_xy` derives the projected bin coordinates from them. There is no seam to inject a dynamic position value into the projection, so even though #827 makes the f144 message reach the detector handler, it currently has nowhere to go.

## Concrete case: LOKI rear-detector XY view

- Today's active geometry file `geometry-loki-2026-03-11.nxs` contains `detector_carriage/value = 0.0 mm` on top of a `detector_carriage_zero` translation of ~0.983 m in z. The XY projection is therefore frozen at the carriage's parked/zero position regardless of where the carriage actually is in production.
- The `detector_carriage` f144 stream (\"Sample-Rear Distance\") is registered in `loki/specs.py` and confirmed to be flowing in production, but no consumer exists.
- A new dated geometry snapshot has to be added to the LUT in `detector_data_handler.py` every time the carriage moves — this is the workaround we are implicitly relying on today.

## Proposal

Make the detector view workflow accept `context_keys` analogous to what #827 did for `create_monitor_workflow`, so that a position value arriving via `aux_sources` can be threaded into the projection as dynamic geometry:

1. Teach `GeometricViewConfig` / `DetectorViewFactory` to optionally take a context key for the relevant transformation (e.g. the carriage z), and apply it on top of the static NeXus pixel positions when computing projected coordinates.
2. Wire this through `register_detector_view_spec` so per-view aux source subclasses (e.g. a `LokiRearDetectorAuxSources(DetectorROIAuxSources)`) can declare both ROI and position streams.
3. LOKI side: subclass `DetectorROIAuxSources` for the rear bank and pass it into the rear-bank `register_detector_view_spec` call in `loki/factories.py`.

## Accumulator reset (option C from #828, in scope here)

The dynamic-geometry change is not safe without a reset mechanism: once the projection consumes a new carriage value, incoming data carries new bin coordinates and an existing accumulator holding the old histogram will either hit a coord-mismatch error in scipp or silently misalign. Manual UI reset is not a viable interim — it requires the user to notice the move, press reset between the move and the next message, and assumes no in-flight partial window is already poisoned.

Option **C** from #828 (coord-based detection in the accumulator) falls out of the same change essentially for free: we are already threading the carriage value through to recompute the projection, so attaching it as a coord on the intermediate data is one extra step, and the accumulator's \"reset on coord mismatch\" check becomes the natural enforcement of its own invariant. No thresholds, no dead-band tuning — the comparison is on a value we set as a coord, not on a raw sensor reading.

This issue should land C as part of the work, scoped to the detector-view path. It is intended as the **reference implementation** for other workflows that may later need the same treatment; #828 stays open and will be updated to point here once this is working.

## Out of scope

- Options A and B from #828 (preprocessor-level filtering, callable predicate). Tracked in #828 as escape hatches for cases where the context value cannot naturally manifest as a coord.
- Other instruments' dynamic geometry needs.

## Related

- #827 — aux source infrastructure for monitor and detector view workflows
- #828 — automatic accumulator reset on position context change (broader design discussion; this issue implements option C for the detector-view path)

SimonHeybrock:
## Summary

Wires LOKI's `detector_carriage` f144 position stream into the rear-bank detector-view workflow so the published image tracks the carriage position as it moves along the beam. Implements the plan in `docs/developer/plans/dynamic-detector-geometry-loki-carriage.md`.

The seam is generic: any instrument can opt in by passing `dynamic_transforms={source: TransformValueStream(...)}` to `DetectorViewFactory` and using `DetectorROIAuxSources(dynamic_transforms=...)` in the workflow spec.

Fixes #845.

## How it works

- `add_dynamic_transform` inserts two providers into the workflow for sources that opt in: one that extracts the latest sample from the NXlog `set_context` stream into a `TransformValue`, and one that replaces essreduce's `get_transformation_chain` with a variant that deepcopies the chain and injects the live value. If the f144 stream has not yet produced a sample, the provider raises: the reference geometry file's baked-in value is intentionally never used, because it may be stale or invalid and a wrong result is worse than no result. Workflows without dynamic geometry keep the unchanged upstream provider.
- `DetectorROIAuxSources` now advertises the configured f144 streams alongside the ROI streams. These streams are physical properties of the instrument (not job-specific), so `render()` routes them un-prefixed and only to the jobs whose `source_name` actually consumes them.
- LOKI rear bank (`loki_detector_0`) maps the `detector_carriage` aux stream onto the NeXus entry `/entry/instrument/detector_carriage/value`. The mapping lives in a single `LOKI_DYNAMIC_TRANSFORMS` dict shared between the spec (for routing) and the factory (for graph wiring).
- Dashboard `--dev` slider via `configs/log_producer_loki.json` drives the stream interactively.

## Known limitation (deferred to #828)

A carriage move mid-run raises on the next `accumulate` (different bin edges / different `z`) and wedges the workflow until a manual UI reset or next `run_start`. This is expected and documented. The natural follow-up is auto-reset keyed on a change in the scalar `z` coord now that the signal is explicit.

## Test plan

- [x] Unit tests for chain deepcopy/injection, NXlog → `TransformValue` extraction, missing-entry error, and scalar-coord stamping (including a negative case for non-scalar entries)
- [x] End-to-end run with real LOKI geometry file: two-batch accumulate at fixed position, scalar `z` present on every output, carriage move correctly wedges the accumulator
- [x] Picks up f144 values from before workflow start (green workflow status from the get go)
- [x] Starts in warning (yellow) state when to prior f144 value, recovers to green after first f144.
- [x] f144 change after prior values turns workflow yellow (see #828). Manual reset after a move recovers the workflow.


src/ess/livedata/handlers/detector_view/factory.py
```diff
@@ -73,9 +76,23 @@ def __init__(
         *,
         data_source: DetectorDataSource,
         view_config: ViewConfig | dict[str, ViewConfig],
+        dynamic_transforms: dict[str, TransformValueStream] | None = None,
     ) -> None:
+        """
+        Parameters
+        ----------
+        dynamic_transforms:
```
jl-wynen: Should be in the class docstring.

src/ess/livedata/handlers/detector_view/projectors.py
```diff
@@ -149,7 +164,10 @@ def project_events(self, events: sc.DataArray) -> sc.DataArray:
         )
 
         # Bin by screen coordinates (preserving events)
-        return flat_events.bin(self._edges)
+        binned = flat_events.bin(self._edges)
+        for name, value in self._scalar_coords.items():
+            binned.coords[name] = value
+        return binned
```
jl-wynen:
```suggestion
        return flat_events.bin(self._edges).assign_coords(self._scalar_coords)
```

src/ess/livedata/handlers/detector_view/workflow.py
```diff
@@ -56,10 +59,66 @@
...
+    TransformName,
+    TransformValue,
+    TransformValueLog,
     UsePixelWeighting,
 )
 
 
+def get_transformation_chain_with_value(
+    detector: NeXusComponent[snx.NXdetector, SampleRun],
+    transform_value: TransformValue,
```
jl-wynen, line 71: Do you foresee cases where more than one entry needs to be replaced? E.g., where something gets translated and rotated?

SimonHeybrock: Yes I do. My thought was that we'd just to turn this into something like `Transform1Value` and `Transform2Value` then. I did not introduce this here since it is in principle a minor change, once this here is proven to work in practice as planned.

jl-wynen (approved):

PR #850, opened 2026-04-08
Fix Kafka broker timestamps treated as ns instead of ms

### LINKED ISSUE #848: Kafka broker timestamps treated as ns instead of ms in 5 adapter sites
## Problem

`confluent_kafka.Message.timestamp()` returns `(type, milliseconds_since_epoch)`, but five adapters in `src/ess/livedata/kafka/message_adapter.py` wrap the ms value with `Timestamp.from_ns(...)`, producing a silent 1,000,000× unit error:

- `message_adapter.py:149` — `KafkaToEv44Adapter` fallback (when `ev44.reference_time` is empty)
- `message_adapter.py:262` — `RunControlAdapter` envelope timestamp (note: the payload `start_time`/`stop_time` inside `RunStart`/`RunStop` already use `from_ms` correctly; only the outer `Message.timestamp` is wrong)
- `message_adapter.py:322` — `KafkaToMonitorEventsAdapter` fallback (when `reference_time` is empty)
- `message_adapter.py:402` — `CommandsAdapter`
- `message_adapter.py:413` — `ResponsesAdapter`

Pre-#829 these were bare ints and the unit mismatch was silent. Post-#829 they become `Timestamp` values carrying the wrong magnitude, which may distort batching/ordering/lag calculations anywhere downstream that relies on `Message.timestamp`.

Most other adapters correctly use payload-level timestamps (ev44 `reference_time`, da00 `timestamp_ns`, f144 `timestamp_unix_ns`, ad00 `timestamp_ns`) and are not affected. The sixth instance of `message.timestamp()` usage (`X5f2ToStatusAdapter`) was fixed in #847.

## Fix

Replace `Timestamp.from_ns(message.timestamp()[1])` with `Timestamp.from_ms(message.timestamp()[1])` at the five sites above.

Worth auditing downstream consumers of `Message.timestamp` at the same time to confirm nothing was quietly compensating for the old units.

## How this should have been caught

Related gap tracked separately: no `KafkaSink`/adapter tests exercise the `confluent_kafka.Message` path end-to-end, so these unit bugs slipped through. See the companion test-coverage issue.

SimonHeybrock:
## Summary

`confluent_kafka.Message.timestamp()` returns milliseconds since epoch, but five adapters in `message_adapter.py` wrapped the value with `Timestamp.from_ns`, producing a silent 1e6× unit error. Switches all five sites to `Timestamp.from_ms`.

## Impact

Only two of the five sites actually had observable effect:

- **`KafkaToEv44Adapter` and `KafkaToMonitorEventsAdapter` fallback paths** (when `ev44.reference_time` is empty) — the envelope timestamp flows into downstream batching/ordering, so the bug bit here. These fallbacks exist mainly for replaying serialized test data, but are the critical case to get right.
- **`RunControlAdapter`, `CommandsAdapter`, `ResponsesAdapter`** — latent. Tracing the consumers shows `Message.timestamp` is never read for these streams:
  - `RunStart`/`RunStop` consumers in `JobManager` only read the payload `start_time`/`stop_time` (already correctly converted from ms).
  - `ConfigProcessor` reads only `message.value` (`RawConfigItem`) and does not propagate the timestamp into the responses it emits.
  - The dashboard `Orchestrator` forwards only `stream_id` + `value` for responses.

  Fixing these anyway to remove the landmine for any future consumer.

## Tests

The existing fallback tests for ev44/monitor adapters already exercised the relevant code path, but baked in the buggy `from_ns` expectation. Updated them to assert `from_ms`, which is what exercises the fix.

Closes #848

### DIFF
--- a/src/ess/livedata/kafka/message_adapter.py
+++ b/src/ess/livedata/kafka/message_adapter.py
@@ -146,7 +146,7 @@ class KafkaToEv44Adapter(KafkaAdapter[eventdata_ev44.EventData]):
         if ev44.reference_time.size > 0:
             timestamp = Timestamp.from_ns(ev44.reference_time[-1])
         else:
-            timestamp = Timestamp.from_ns(message.timestamp()[1])
+            timestamp = Timestamp.from_ms(message.timestamp()[1])
         return Message(timestamp=timestamp, stream=stream, value=ev44)
 
 
@@ -259,7 +259,7 @@ class RunControlAdapter(MessageAdapter[KafkaMessage, Message[RunStart | RunStop]
     def adapt(self, message: KafkaMessage) -> Message[RunStart | RunStop]:
         buf = message.value()
         schema = streaming_data_types.utils.get_schema(buf)
-        timestamp = Timestamp.from_ns(message.timestamp()[1])
+        timestamp = Timestamp.from_ms(message.timestamp()[1])
         if schema == 'pl72':
             info = run_start_pl72.deserialise_pl72(buf)
             stop_time = (
@@ -319,7 +319,7 @@ class KafkaToMonitorEventsAdapter(KafkaAdapter[MonitorEvents]):
         if reference_time.size > 0:
             timestamp = Timestamp.from_ns(reference_time[-1])
         else:
-            timestamp = Timestamp.from_ns(message.timestamp()[1])
+            timestamp = Timestamp.from_ms(message.timestamp()[1])
         return Message(
             timestamp=timestamp,
             stream=stream,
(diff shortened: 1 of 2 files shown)


jl-wynen:
Are there any tests that would catch mismatches with kafka like this? Or does it only show up during production?

SimonHeybrock:
> Are there any tests that would catch mismatches with kafka like this? Or does it only show up during production?

It never showed up, see PR description: We generally do not use the Kafka message timestamps as our message payloads carry their own timestamps (from the ESS timing system). It was only used as a fallback in some cases, but we never hit that in practice. `TestKafkaToEv44Adapter.test_no_reference_time_uses_message_timestamp` and `TestKafkaToMonitorEventsAdapter.test_no_reference_time_uses_message_timestamp` cover the fallback case.

jl-wynen (approved):

PR #851, opened 2026-04-08
Refactor KafkaSink to a symmetric serializer architecture

### LINKED ISSUE #849: No test coverage for KafkaSink / x5f2 serialization path
## Problem

`grep -r KafkaSink tests/` returns nothing. There is no test anywhere that exercises `KafkaSink.publish_messages`, which is where `service_status_to_x5f2` / `job_status_to_x5f2` are actually invoked in production. Similarly, no test drives a real `confluent_kafka.Message` through the adapter layer.

`FakeMessageSink` skips serialization entirely, so every `LivedataApp`/`OrchestratingProcessor`-based integration test silently avoids the Kafka encoding path. This is how both issues fixed in #847 slipped through CI:

1. `ServiceStatusPayload.started_at: int` vs `ServiceStatus.started_at: Timestamp` mismatch — killed every backend service on its first heartbeat on `main` after #829.
2. `X5f2ToStatusAdapter` using positional `Timestamp(...)` — killed dashboard status consumption.

Both would have been caught by a test that ran `OrchestratingProcessor.process()` through a sink that performs x5f2 encoding, or an adapter test that fed a real Kafka message object into `X5f2ToStatusAdapter.adapt`.

## Suggested approaches

Two options, not mutually exclusive:

1. **Serializing test sink.** Add a variant of `FakeMessageSink` that mirrors `KafkaSink.publish_messages`' serialization dispatch (`service_status_to_x5f2` / `job_status_to_x5f2` / `self._serializer(msg)`), but stores the resulting bytes instead of producing to Kafka. Existing `LivedataApp` tests can opt in and would catch this entire class of type drift for free.
2. **Direct `KafkaSink` tests.** A small test file using a fake `confluent_kafka.Producer` (just capturing `produce()` calls) to round-trip `ServiceStatus`/`JobStatus` through the real `KafkaSink`. Cheap and focused.

Option 1 gives broader coverage for free; option 2 is a targeted safety net. I'd suggest starting with option 1.

## Related

- #847 (fix for both the backend heartbeat and dashboard adapter crashes)
- #848 (ms-as-ns unit bug in five other adapter sites — same coverage gap)

SimonHeybrock:
## Summary

Closes #849.

The sink side of the Kafka boundary now mirrors the source-side adapter architecture: a `MessageSerializer` protocol, one concrete class per encoding concern (data/status/commands/responses), and routing primitives for composition. `KafkaSink` itself no longer knows about `StreamKind`, `STATUS_STREAM_ID`, instrument names, hostname/pid, or topic resolution — it delegates encoding entirely to its injected serializer and handles only producer I/O, error logging, and metrics.

This closes the coverage gap #849 describes, which let the type-drift regressions in #847 (and the unit confusion in #848) ship. Each new serializer class has a round-trip unit test pairing it with its source-side `MessageAdapter`, explicitly asserting that decoded `Timestamp` fields are of the correct type. `KafkaSink` itself gets producer-interaction tests via a hand-rolled fake producer and a new `producer_factory` injection seam.

## Why this shape

The suggested fix in #849 was "add tests" (option 1: serializing test sink; option 2: direct `KafkaSink` tests with a fake producer). Both would work, but they'd leave the underlying asymmetry in place: the source side has per-schema `MessageAdapter` classes + routers, while the sink side inlines everything in one big `if/elif`. Making the sink symmetric to the source yields the same test coverage "for free" (each serializer is trivially unit-testable in isolation), removes the only reason a test fake would need to duplicate production dispatch logic, and makes encode/decode pairs obviously reviewable as round-trip pairs.

## Test plan

- [x] Round-trip assertions explicitly check `isinstance(decoded.started_at, Timestamp)` — these would have caught #847


src/ess/livedata/kafka/sink.py
```diff
@@ -1,99 +1,81 @@
...
-        )
-    except (ValueError, TypeError) as e:
-        raise SerializationError(f"Failed to serialize message: {e}") from None
-    return f144
+    Mirrors the source-side :class:`~ess.livedata.kafka.message_adapter.MessageAdapter`
+    protocol. Implementations are responsible for choosing the destination topic and
+    producing the key/value bytes; :class:`KafkaSink` only handles producer I/O.
+    """
+
+    def serialize(self, message: Message[T]) -> SerializedMessage: ...
```
YooSunYoung, line 49: We could keep the serializer as Callable like before, sth like...
 
```python
class MessageSerializer(Protocol, Generic[T]):
    def __call__(self, message: Message[T]) -> SerializedMessage: ...
```

but... maybe class with the specific method makes it easier to read.

So ... :+1:

src/ess/livedata/kafka/sink.py
```diff
@@ -1,99 +1,81 @@
...
+    """
+
     def __init__(
         self,
         *,
-        instrument: str,
         kafka_config: dict[str, Any],
-        serializer: Serializer[T] = serialize_dataarray_to_da00,
+        serializer: MessageSerializer[T],
+        producer_factory: Callable[[dict[str, Any]], kafka.Producer] = kafka.Producer,
```
YooSunYoung, line 66: [Not super important question]

Will it be easier if we just pass the producer instead of config and factory?

What is the benefit of using factory in this context...?

YooSunYoung (approved):
The serialization pipeline looks more encapsulated and easier to follow 👍 
I don't have much to add.

src/ess/livedata/kafka/sink.py
SimonHeybrock: We want to create it only in `__enter__` and manage the context. Same as before this change, except that the factory was just hard-coded to `kafka.Producer` (see `__enter__` further down).

PR #852, opened 2026-04-08
Migrate to essreduce 26.4.0: replace TOF mode with wavelength mode

SimonHeybrock:
## Motivation

essreduce 26.4.0 replaced `ess.reduce.time_of_flight` with `ess.reduce.unwrap`. The lookup table now converts time-of-arrival → **wavelength** directly (previously → time-of-flight), making the old `'tof'` coordinate mode a misnomer. This PR completes the migration and renames it to `'wavelength'` throughout.

essspectroscopy 26.3.0 and essreflectometry 26.4.0 are also required to stay compatible with the new essreduce.

## Changes

**API / coordinate mode**
- `CoordinateMode` is now `Literal['toa', 'wavelength']` — `'tof'` is removed with no migration shim (units changed from ns → Å, so no meaningful value to preserve)
- `GenericTofWorkflow` → `GenericUnwrapWorkflow`; `TofDetector`/`TofMonitor` → `WavelengthDetector`/`WavelengthMonitor`; `TofLookupTableFilename` / `TimeOfFlightLookupTableFilename` → `LookupTableFilename`
- New required input `LookupTableRelativeErrorThreshold` set to `{source_name: float('inf')}` (no masking), consistent with `LokiWorkflow`/`DreamWorkflow` defaults

**Pydantic model cleanup**
- `tof_edges` / `tof_range` fields removed from `DetectorViewParams` and `MonitorDataParams`
- `TOFEdges` / `TOFRange` removed from `parameter_models.py`

**Instrument factories** (DREAM, LOKI, Bifrost) updated to use the new names.

**Nightly CI**: disabled git-based nightly dependencies for now — essreduce moved to the `scipp/ess` monorepo and the old `scipp/essreduce` repo is stale. Needs a separate fix once the monorepo migration strategy is clear.

## Test plan

- [x] Manual smoke test with LOKI detector view in wavelength mode


src/ess/livedata/handlers/detector_view/factory.py
```diff
@@ -142,9 +142,10 @@ def make_workflow(
             coordinate_mode=mode,
         )
 
-        # Set lookup table filename for TOF/wavelength modes
-        if mode in ('tof', 'wavelength'):
-            workflow[TofLookupTableFilename] = tof_lookup_table_filename
+        # Set lookup table filename and error threshold for wavelength mode
+        if mode == 'wavelength':
+            workflow[LookupTableFilename] = tof_lookup_table_filename
+            workflow[LookupTableRelativeErrorThreshold] = {source_name: float('inf')}
```
SimonHeybrock, line 148: What is a good default here?

src/ess/livedata/config/instruments/bifrost/factories.py
```diff
@@ -150,9 +151,17 @@ def _init_q_cut_workflow() -> sciline.Pipeline:
         fname = simulated_elastic_incoherent_with_phonon()
         with snx.File(fname) as f:
             detector_names = list(f['entry/instrument'][snx.NXdetector])
+            monitor_names = list(f['entry/instrument'][snx.NXmonitor])
         workflow = BifrostQCutWorkflow(detector_names)
         workflow[Filename[SampleRun]] = fname
-        workflow[TimeOfFlightLookupTableFilename] = tof_lookup_table_simulation()
+        workflow[LookupTableFilename] = lookup_table_simulation()
+        # The spectroscopy TofWorkflow looks up the detector threshold under the
```
nvaytet: I am not sure `TofWorkflow` exists in the spectroscopy package?

src/ess/livedata/config/instruments/dream/factories.py
```diff
@@ -100,7 +100,7 @@ def _detector_view_workflow_factory(
     ) -> StreamProcessorWorkflow:
         """Factory for Sciline-based detector view workflow."""
         tof_lookup_table_filename = None
```
nvaytet: The lookup table no longer gives tof. This should probably renamed to something like `lookup_table_filename` or `wavelength_lookup_table_filename`

src/ess/livedata/handlers/detector_view/factory.py
```diff
@@ -142,9 +142,10 @@ def make_workflow(
             coordinate_mode=mode,
         )
 
-        # Set lookup table filename for TOF/wavelength modes
-        if mode in ('tof', 'wavelength'):
-            workflow[TofLookupTableFilename] = tof_lookup_table_filename
+        # Set lookup table filename and error threshold for wavelength mode
+        if mode == 'wavelength':
+            workflow[LookupTableFilename] = tof_lookup_table_filename
```
nvaytet:
```suggestion
            workflow[LookupTableFilename] = lookup_table_filename
```

src/ess/livedata/handlers/detector_view/factory.py
nvaytet: I think I would need a bit more context. Does this apply to all instruments? I forget.
I would have said if we are using the workflows from the various instrument packages, I think most/all of them define defaults...
But I guess the `source_name` here is not defined in the instrument packages, it's specific to livedata right?

Then `inf` for now should be fine. To be honest, if instruments are correctly set-up, they should not need to mask. Having large uncertainties is a sign that the chopper cascade is badly configured.

tests/handlers/monitor_workflow_test.py
```diff
@@ -727,23 +732,25 @@ def lookup_table_filename(self):
...
         This is a regression test for the issue where histogram data with
-        frame_time coordinate was not properly handled by the TOF conversion.
-        The essreduce _time_of_flight_data_histogram function expects one of:
+        frame_time coordinate was not properly handled by the TOA-to-wavelength
+        conversion. The essreduce unwrap function expects one of:
         'time_of_flight', 'tof', or 'frame_time' as the coordinate name.
         """
         # Use monitor_cave because its Ltotal (72.33 m) is within the DREAM
         # lookup table range (59.85-80.15 m). monitor_bunker has Ltotal of
         # only 6.62 m which is outside the lookup table range.
```
nvaytet: Unrelated to these changes, but note that the instrument luts (including Dream) now extend all the way to 5m: https://scipp.github.io/essdiffraction/user-guide/dream/dream-make-wavelength-lookup-table.html

src/ess/livedata/handlers/detector_view/factory.py
SimonHeybrock: Yes, all detector views across all instruments, and similar for monitors. `source_name` typically maps to NeXus group names, so roughly what we also have in instrument packages.

tests/handlers/monitor_workflow_test.py
SimonHeybrock: 👍

SimonHeybrock:
@nvaytet I don't know if you had more comments but our deployment was broken so we merged this early.

nvaytet:
It looks fine to me 👍

PR #854, opened 2026-04-09
Notify dashboard of job state on backend shutdown

### LINKED ISSUE #823: Backend shutdown or crash leaves jobs in "active" state
When a backend worker shuts down (or crashes), the dashboard keeps displaying any running jobs as active/green. The user can then just click "stop" and start a new job, but it is confusing and bad UX. To solve this, I think we could simply make sure that a backend shutdown (gracefully or not?) sends a final "stopped" state for each of its jobs.

Open questions:
- Can our STOPPED job-heartbeat message include a reason? Might be useful to distinguish from user-initiated shutdown.
- Should the frontend display a notification popup? Per-job, or just a single one about the backend worker going down (potentially including a list of jobs)?

SimonHeybrock:
## Summary

Fixes #823 — when a backend worker shuts down, the dashboard keeps displaying jobs as active/green. The worker's `finalize()` now sends per-job `JobStatus(stopped)` messages alongside the final `ServiceStatus`, so the dashboard can deactivate affected workflows immediately.

**Backend**: Consolidated the previous three-method shutdown sequence (`shutdown()` / `report_stopped()` / `report_error()`) into a single idempotent `finalize()`. It reports all jobs as stopped on the wire without mutating internal job state — just `dataclasses.replace(status, state=JobState.stopped)` at message construction time.

**Dashboard**: `Orchestrator.forward()` reacts to incoming `JobState.stopped` statuses by calling `JobOrchestrator.check_stopped()`, which deactivates the workflow once all its jobs have reported stopped. A `StoppedReason` enum distinguishes user-initiated stops from backend shutdowns, shown as "Backend shut down" in the workflow status widget. Also removes `ServiceState.stopping` (the old intermediate shutdown state) throughout.

## Test plan

- [x] Integration tests for `finalize()`: sends correct job/service statuses, handles errors, idempotent
- [x] Unit tests for `check_stopped()` / `_all_jobs_stopped()` reconciliation
- [x] Manual: start dummy services, kill backend with SIGTERM, dashboard shows STOPPED with "Backend shut down"


src/ess/livedata/dashboard/job_orchestrator.py
```diff
@@ -151,6 +161,7 @@ def __init__(
 
         # Workflow state tracking
         self._workflows: dict[WorkflowId, WorkflowState] = {}
+        self._job_states: dict[JobId, JobState] = {}
```
MridulS, line 164: There is no clean up of `_job_states`? Wouldn't a restart of the workflow always keep adding new JobIDs?

SimonHeybrock: Good catch!

MridulS (approved):

PR #865, opened 2026-04-13
Hide window and rate params from timeseries plotter config

SimonHeybrock:
## Summary

The timeseries plotter uses `FullHistoryExtractor`, which collapses per-frame `start_time`/`end_time` coords into a single pair spanning the entire buffer. Rate normalization would therefore divide every point by the total buffer duration rather than each point's own frame duration, giving wrong results. Hide the "Counts Per Second" toggle from the timeseries config to prevent this.

Windowing is also not applicable — the timeseries plotter always shows the full time history via `FullHistoryExtractor`.

The plotter registry derives the config UI model from the factory's type hint, so switching to `PlotDisplayParams1d` (which lacks `window` and `rate`) removes both fields from the config modal. Existing saved configs with old `window`/`rate` keys load cleanly — Pydantic silently discards unknown fields.

## Test plan

- [x] Open the plot config modal and add a Timeseries plot — confirm the modal has no windowing or rate normalization options
- [x] Check that Lines / Bars plotters still show the windowing and rate options as before
- [x] Load an existing dashboard config that has a timeseries plot — verify it loads without errors and displays correctly

Closes #844

### DIFF
--- a/src/ess/livedata/dashboard/correlation_plotter.py
+++ b/src/ess/livedata/dashboard/correlation_plotter.py
@@ -279,7 +279,7 @@ class CorrelationHistogram1dPlotter(CorrelationHistogramPlotter):
                 bins=params.bins.x_bins,
             )
         ]
-        renderer = LinePlotter.from_params(params)
+        renderer = LinePlotter.from_display_params(params)
         super().__init__(
             axes=axes, normalize=params.normalization.per_second, renderer=renderer
         )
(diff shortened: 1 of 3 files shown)


jokasimr (commented):
I'm afraid I don't understand the context here.

Could you give a bit more details about how timeseries are plotted currently, what's `window` and `rate`?

SimonHeybrock:
> I'm afraid I don't understand the context here.
> 
> Could you give a bit more details about how timeseries are plotted currently, what's `window` and `rate`?

`window` and `rate` are options for plotting data, say a monitor histogram. `window` is the integration time (like "last 30 seconds"), `rate` whether to normalize, i.e., plot as `counts/second` or just `counts`. Neither of these options make sense when plotting, e.g., `sample_temperature(time)`, so this PR removes them from the params, which implies they are removed from the widgets.

jokasimr (approved):

PR #866, opened 2026-04-13
Update LOKI geometry and monitor config from latest CODA file

### LINKED ISSUE #781: [LOKI] Geometry file contains duplicate (outdated) monitor groups
Latest file from CODA (`coda_loki_999999_00021677.hdf`) contains monitor groups with the old naming (`beam_monitor_<i>`) in addition to the new naming (`beam_monitor_m<i>`). Wait until this is fixed upstream, then make a new geometry file. Our config already points at the new group names so the extra groups should not actually break anything, unless there are further issues in the files.

### LINKED ISSUE #806: [LOKI] Revert cbm monitor indexing back to 1-based once producers ship the rename
When LOKI detector producers change their CBM source naming from 0-based (\`cbm0..4\`) to 1-based (\`cbm1..5\`), revert commit c5683398.

## Background

PR #749 switched to 1-based indexing based on a promised producer change. That change had not happened as of the time this issue was opened (#804), so commit c5683398 added a LOKI-specific \`cbm_start=0\` override to keep monitors functional during the commissioning period. All other instruments continue to use 1-based indexing.

Closes #804 (once the revert is applied).

SimonHeybrock:
## Summary

- Revert LOKI cbm monitor indexing from 0-based back to 1-based — producers have shipped the rename (cbm1..5), confirmed in `coda_loki_999999_00026352.hdf`
- Add new geometry file `geometry-loki-2026-04-13.nxs` with consistent monitor naming (no duplicates)
- Fix detector_carriage label and description to reflect it is a translation from instrument origin, not a sample-read distance

Closes #781
Closes #806

### DIFF
--- a/src/ess/livedata/config/instruments/loki/specs.py
+++ b/src/ess/livedata/config/instruments/loki/specs.py
@@ -228,8 +228,8 @@ instrument = Instrument(
             description='Downstream, movable (on detector carriage)',
         ),
         'detector_carriage': SourceMetadata(
-            title='Sample-Rear Distance',
-            description='Rear detector carriage position along the beam axis (z).',
+            title='Rear Detector Carriage',
+            description='Rear detector carriage position (5.098 m offset from origin).',
         ),
     },
 )
--- a/src/ess/livedata/config/instruments/loki/specs.py
+++ b/src/ess/livedata/config/instruments/loki/specs.py
@@ -229,7 +229,8 @@ instrument = Instrument(
         ),
         'detector_carriage': SourceMetadata(
             title='Rear Detector Carriage',
-            description='Rear detector carriage position w.r.t. its zero at z=5.098 m after the sample.',
+            description='Rear detector carriage position w.r.t. its zero at z=5.098 m '
+            'after the sample.',
         ),
     },
 )
--- a/src/ess/livedata/config/instruments/loki/specs.py
+++ b/src/ess/livedata/config/instruments/loki/specs.py
@@ -229,7 +229,7 @@ instrument = Instrument(
         ),
         'detector_carriage': SourceMetadata(
             title='Rear Detector Carriage',
-            description='Rear detector carriage position (5.098 m offset from origin).',
+            description='Rear detector carriage position w.r.t. its zero at z=5.098 m after the sample.',
         ),
     },
 )
--- a/tests/config/monitor_source_names_test.py
+++ b/tests/config/monitor_source_names_test.py
@@ -17,14 +17,10 @@ from ess.livedata.config.instruments import available_instruments
 
 @pytest.mark.parametrize(
     'instrument',
-    [i for i in available_instruments() if i != 'loki'],
+    available_instruments(),
 )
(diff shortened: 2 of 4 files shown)


nvaytet:
Not sure I follow. In `coda_loki_999999_00026352.hdf` I am still seeing monitors 0-4, not 1-5?
Screenshot_20260414_101615

Maybe I'm just not looking at the right thing?

SimonHeybrock:
> Not sure I follow. In `coda_loki_999999_00026352.hdf` I am still seeing monitors 0-4, not 1-5?

Those are the *names of the monitors (and monitor groups in the NeXus file)*. 1-5 is the `source_name` in Kafka, where the indices start at 1, i.e., `cbm1`. This is now the same for all instruments.

src/ess/livedata/config/instruments/loki/specs.py
```diff
@@ -228,8 +228,8 @@ class SansWorkflowParams(pydantic.BaseModel):
             description='Downstream, movable (on detector carriage)',
         ),
         'detector_carriage': SourceMetadata(
-            title='Sample-Rear Distance',
-            description='Rear detector carriage position along the beam axis (z).',
+            title='Rear Detector Carriage',
+            description='Rear detector carriage position (5.098 m offset from origin).',
```
nvaytet:
```suggestion
            description='Rear detector carriage position (+5.098 m offset from origin).',
```

nvaytet: Also, did I understand correctly that the 5.098 m is an offset to which another moving offset is then added to get the actual position of the monitor? Or is the 5.098 the actual position of the monitor that can change?
Basically, it's a little unclear what the number represents.

tests/config/monitor_source_names_test.py
```diff
@@ -38,23 +34,19 @@ def test_production_monitors_do_not_use_cbm0(instrument: str) -> None:
...
         for key, value in stream_mapping.monitors.items()
         if key.source_name.startswith('cbm')
     }
     expected_mapping = {
-        'cbm0': 'beam_monitor_m0',
-        'cbm1': 'beam_monitor_m1',
-        'cbm2': 'beam_monitor_m2',
-        'cbm3': 'beam_monitor_m3',
-        'cbm4': 'beam_monitor_m4',
+        'cbm1': 'beam_monitor_m0',
```
nvaytet, line 45: 👍

src/ess/livedata/config/instruments/loki/specs.py
SimonHeybrock:
```suggestion
            description='Rear detector carriage position w.r.t. its zero at z=5.098 m after the sample.',
```
clearer?

SimonHeybrock: > Or is the 5.098 the actual position of the monitor that can change? Basically, it's a little unclear what the number represents.

It is just the motor reading of the detector carriage. As far as I understand both the rear detector and the last monitor are mounted on it, each with some offset.

nvaytet (approved):

PR #867, opened 2026-04-13
Add counts-per-pixel workflow for pixellated beam monitors

SimonHeybrock:
## Summary

- LOKI beam_monitor_m3 is a pixellated monitor (5 pixels, IDs 4-8). This adds a lightweight counts-per-pixel workflow that runs alongside the existing monitor histogram workflow.
- The adapter emits `DetectorEvents` for pixellated monitors (preserving pixel_id) while keeping `StreamKind.MONITOR_EVENTS`, so both workflows share the same `ToNXevent_data` preprocessor.
- `Instrument.create_pixellated_monitor` marks a monitor as pixellated; `load_factories` will auto-load pixel IDs from NeXus when available (LOKI's geometry file lacks them, so they're hard-coded).
- Re-uses the standard detector-view workflow, creating a logical view.

## Test plan

- [x] Existing monitor workflow displayed and works unchanged
- [x] New monitor workflow for pixel counts displayed and working for LOKI


src/ess/livedata/handlers/monitor_workflow.py
jl-wynen, line 1: Shouldn't this be in ESSsans? Since this is a diagnostics tool, it needs to work on files, not just live data.

src/ess/livedata/handlers/monitor_workflow_specs.py
jl-wynen, line 1: Don't we already have a workflow that shows counts per _detector_ pixel? I am surprised at how much new code there is in this PR. I would have expected that you can just mark this monitor as a detector and run the same workflow but only request `RawData` and then histogram that.

SimonHeybrock: In fact I tried this initially, but it actually turned out more complicated. I think it is partially (or mainly) because `GenericNeXusWorkflow` was not made for such a case. Can have another look if you like?

jl-wynen: At least out of curiosity: What aspect of the workflow doesn't work here?

src/ess/livedata/handlers/monitor_workflow.py
SimonHeybrock: I don't think think this would add anything to ESSsans. You can basically just `scipp.group` the monitor. Only reason for not using `scipp.group` here is performance - we do not need the bin structure, we only care about the counts.

src/ess/livedata/handlers/monitor_workflow_specs.py
SimonHeybrock: - Hard coded to only work for NXdetector.
- Tries to read `detector_number` and maybe other bits that are not in NXmonitor.
- The detector-view workflow integrates a bunch of other things that are not needed, such as ROI computation.

That being said, I think it makes sense to look into it again. Maybe it is easier to fix than I thought initially.

SimonHeybrock: Ok, it turned out this actually works. I believe I failed originally because I had at first tried to make a unified workflow that did both the pixel-view and the existing monitor histogram. Once I figured out that that was a bad idea I had already discarded the approach to use the detector-view and forgot about it.

jl-wynen (approved):

PR #868, opened 2026-04-14
Fix monitor histogram output losing user-configured edge unit

SimonHeybrock:
## Summary

- c319ffc4 fixed wavelength-mode `UnitError` by dynamically inferring the event coordinate unit, but forgot to convert the histogram coordinate back to the user's requested unit — output was always in the event unit (e.g., nanoseconds)
- The detector view already handled this correctly; apply the same pattern to the monitor workflow


SimonHeybrock:
LGTM

PR #872, opened 2026-04-14
Make motion topic subscription selective for detector and monitor services

SimonHeybrock:
## Summary

- Add `has_detector_motion` and `has_monitor_motion` flags to `Instrument`
- `detector_data` and `monitor_data` services only subscribe to f144 motion topics when the corresponding flag is set, avoiding unnecessary Kafka subscriptions for instruments without moving detectors or monitors
- `data_reduction` and `timeseries` services are unchanged (unconditional subscription)
- Both flags set for LOKI (detector carriage + future monitor motion)

## Test plan

- [ ] Verify LOKI detector and monitor services still receive motion data correctly
- [ ] Verify other instruments (e.g., dummy) do not subscribe to motion topics


SimonHeybrock:
Replaced by better approach in #874.

PR #879, opened 2026-04-17
Rate-aware message batcher with per-stream pulse-slot completion

### LINKED ISSUE #875: Rate-aware message batcher to fix aliasing with same-frequency producers
# Rate-Aware Message Batcher

## Problem

The current `SimpleMessageBatcher` uses fixed time windows (`[T, T+1s)`) and closes a batch when a message with a timestamp beyond the window arrives. When a producer's rate is close to the batch frequency (e.g., 1 Hz producer with 1s batches), message timestamps land right on window boundaries. Network/timing jitter determines which side they fall on, producing an erratic pattern of 0, 1, or 2 messages per batch instead of a consistent 1.

The same aliasing occurs at other rates (e.g., 14 Hz), though the relative effect is smaller. No fixed window placement can avoid this: shifting phase just moves the problem to a different boundary. Incommensurate batch lengths (e.g., sqrt(2) seconds) avoid the worst jitter but produce a repeating cycle of uneven batch sizes (e.g., 1, 2, 1, 2, ...).

Multiple streams with different rates are subscribed simultaneously, so per-stream phase tracking with a single unified batch boundary is not viable.

## Example

Image

## Proposed Solution

A new `RateAwareMessageBatcher` that:

- Learns the per-stream message rate from observed data.
- Holds a batch open until each stream has delivered its expected message count.
- Uses a tight wall-clock timeout so a single late/lost message cannot block processing indefinitely.
- Quantizes batch lengths to multiples of 1/14 s (the ESS pulse period) so that expected message counts are always clean integers for any pulse-aligned producer.

### Pulse-Quantized Batch Lengths

ESS operates at 14 Hz. All detector event streams are inherently pulse-aligned. Batch lengths that are exact multiples of 1/14 s guarantee that the expected message count per stream is always an integer (for any producer whose rate is a multiple or divisor of 14 Hz).

The adaptive escalation grid becomes:

| Level | Pulses | Length (s) |
|-------|--------|------------|
| 0     | 14     | 1.000      |
| 1     | 20     | 1.429      |
| 2     | 28     | 2.000      |
| 3     | 40     | 2.857      |
| 4     | 56     | 4.000      |

Each step is `round(14 * sqrt(2)^n)` pulses, preserving the ~sqrt(2) escalation ratio while snapping to the pulse grid. The exact values can be tuned; the key constraint is that each level is an integer number of pulses.

### Per-Stream Rate Estimation

For each `StreamId`, the batcher tracks an exponential moving average (EMA) of messages received per batch period:

```
expected[stream] = EMA of (messages_received / batch_periods_elapsed)
```

On startup (before the rate is known), the batcher falls back to pure timestamp-based batching (current behavior). Once enough samples are collected (e.g., 3 batches), rate-aware logic kicks in.

When the adaptive batcher changes the batch length, expected counts are rescaled proportionally. The EMA should be slow enough to be stable but fast enough to track instrument reconfigurations (e.g., alpha = 0.1-0.2, so ~5-10 batches to converge).

### Batch Completion Logic

A batch is considered complete when **all known streams** (not counting livedata streams, i.e., only raw streams) have delivered `>= round(expected[stream])` messages for the current batch period.

Streams are identified by `StreamId` (which carries `StreamKind` and `name`). Command/control streams are excluded (already filtered in `OrchestratingProcessor.process()`).

When a stream delivers more messages than expected, the surplus is buffered for the next batch (same as current behavior for future messages).

### Wall-Clock Timeout

If the batch has not completed after a configurable timeout (e.g., `batch_length * 1.5`), the batch closes with whatever messages have arrived. This handles:

- Lost messages (network issues, producer restarts).
- Streams that temporarily stop producing.
- Startup transients before rates are learned.

The timeout must be tight enough that a single lost message does not cause a noticeable processing delay. A good starting point: `batch_length + 2/14 s` (i.e., two extra pulse periods of grace). This is generous enough to absorb jitter but short enough to keep latency bounded.

### Stream Lifecycle

Streams can appear and disappear (e.g., instrument reconfiguration). The batcher should:

- **New stream**: Start tracking immediately, fall back to timestamp-based batching for that stream until rate is estimated.
- **Disappeared stream**: If a stream has not produced any messages for N consecutive batches (e.g., N=5), stop expecting messages from it. This prevents a decommissioned stream from triggering timeouts forever.

### Interaction with Adaptive Escalation

`RateAwareMessageBatcher` replaces `SimpleMessageBatcher` as the inner batcher. `AdaptiveMessageBatcher` is reused unchanged — it wraps `RateAwareMessageBatcher` exactly the way it wraps `SimpleMessageBatcher` today:

```
AdaptiveMessageBatcher          (escalation logic, unchanged)
  └── RateAwareMessageBatcher   (replaces SimpleMessageBatcher)
```

The layering works because the two concerns are orthogonal:

- **`AdaptiveMessageBatcher`**: observes `report_batch()` processing times, decides when to change the batch length, calls `inner.set_batch_length()`. Never touches message routing or batch completion.
- **`RateAwareMessageBatcher`**: decides when a batch is *complete* (per-stream expected counts + wall-clock timeout). Receives batch length changes via `set_batch_length()`.

The only coupling: `set_batch_length()` in `RateAwareMessageBatcher` must snap the requested length to the pulse grid and rescale per-stream expected counts. `AdaptiveMessageBatcher` does not need to know about rates or pulses.

The pulse-grid quantization means the adaptive layer's sqrt(2) steps get rounded, so the actual escalation grid becomes the pulse-quantized table above (14, 20, 28, 40, 56 pulses). This is handled entirely inside `RateAwareMessageBatcher.set_batch_length()` — the adaptive layer requests an arbitrary length, the inner layer snaps it.

## Interface

`RateAwareMessageBatcher` implements the existing `MessageBatcher` ABC and exposes `set_batch_length()` (same as `SimpleMessageBatcher`). No changes to `AdaptiveMessageBatcher` or `OrchestratingProcessor` are needed beyond constructing the new inner batcher.

```python
class RateAwareMessageBatcher(MessageBatcher):
    def __init__(
        self,
        batch_length_s: float = 1.0,  # quantized to nearest 1/14
        pulse_rate_hz: float = 14.0,
        timeout_grace_pulses: int = 2,
        rate_ema_alpha: float = 0.15,
        clock: Callable[[], float] = time.monotonic,
    ) -> None: ...

    def set_batch_length(self, batch_length_s: float) -> None:
        """Snap to pulse grid and rescale expected counts."""

    def batch(self, messages: list[Message[Any]]) -> MessageBatch | None: ...
```

`AdaptiveMessageBatcher` is unchanged; only the constructor wiring changes:

```python
# Before
self._inner = SimpleMessageBatcher(batch_length_s=base_batch_length_s)

# After
self._inner = RateAwareMessageBatcher(batch_length_s=base_batch_length_s)
```

## Message Flow (Steady State)

```
poll 1: receive 12 msgs from detector_stream (expect 14)
         -> return None (incomplete)
poll 2: receive 2 msgs from detector_stream, 1 msg from monitor_stream (expect 1)
         -> all streams met expected count -> return batch
```

```
poll 1: receive 14 msgs from detector_stream (expect 14)
         receive 0 msgs from monitor_stream (expect 1)
         -> return None (incomplete, waiting for monitor)
poll 2: receive 1 msg from monitor_stream
         -> return batch
```

```
poll 1: receive 14 msgs from detector_stream (expect 14)
         -> return None (waiting for monitor)
poll 2: no messages, wall-clock timeout fires
         -> return batch with 14 detector msgs, 0 monitor msgs
```

## Open Questions

- Should the pulse rate (14 Hz) be a global constant or configurable per instrument? Currently all ESS instruments share 14 Hz, but this is a facility property, not a physics one.
- How should streams with very low rates (e.g., slow log values updating once per minute) interact with the rate tracking? Probably: streams with expected count < 0.5 per batch should be excluded from the completion gate and treated as opportunistic (include if available, don't wait for them).
- Should the batcher expose per-stream rate estimates in its status for diagnostics?

SimonHeybrock:
Closes #875.

Implements `RateAwareMessageBatcher` as an alternative inner batcher to `SimpleMessageBatcher` under `AdaptiveMessageBatcher`, addressing the aliasing issue where producers at frequencies close to the batch rate yield erratic 0/1/2 messages per batch.

## Example

Example from production, which actually triggered this:

### Before
Jumping monitor rates at Bifrost, leading to multiple lines in timeseries plot:
Screenshot 2026-04-17 at 12 35 11

### After
Screenshot 2026-04-21 at 09 48 57

## Approach

Completion is decided per stream by a **pulse-slot gate** on timestamps: a message whose timestamp falls in the last expected pulse slot closes that stream's contribution. This is robust to missed/split pulses and out-of-order arrivals, and avoids count-vs-rate drift. Per-stream period is inferred from inter-arrival diffs via a median-seeded integer-multiple fit, snapped to integer Hz (the format ESS sources publish).

Selectable via `--batcher {simple,rate-aware}` on services; default remains `simple`. `AdaptiveMessageBatcher`, `OrchestratingProcessor`, and the `MessageBatcher` ABC are unchanged.

## Stream classes handled

- **Gridded gated streams** (detector/monitor events): slot-gate closure, overflow re-routing.
- **Sub-Hz gated streams** (rate too low for a grid): ride along in the currently active batch without gating.
- **Non-gated streams** (logs, etc.): appended to the active batch, never block closure.
- **Broken streams** (disjoint epoch or origin implausibly far from batch start): rejected from the grid, still delivered non-gating.

## Robustness

- **Logical-clock timeout** (`timeout_factor` × batch length, default 1.2×): a later message advances the high-water mark past the threshold and closes an incomplete batch. No wall-clock dependency.
- **HWM clamping** against DoS: a single far-future timestamp is clamped to `start + 3 × batch_length` with a monotonicity floor.
- **Phase offset**: grid origin picks a real pulse, not the window start.
- **Gap recovery**: long silences jump the active window forward instead of emitting placeholder batches.
- **Stream lifecycle**: absent counter evicts streams after 5 absent batches; late rejoin rebuilds the grid.
- **Batch-length change** via `set_batch_length`: slot count recomputes; origin preserved when still plausible.
- Invalid-timestamp messages are dropped with a warning instead of silently corrupting the grid.

## Docs

A single-page visual overview is added under Developer Design → Rate-Aware Message Batcher, covering the 4 core mechanics (period estimator, grid, slot gate, HWM) and 15 scenarios.

## Test plan

Unit tests cover: slot-gate closure, missing/split pulses, overflow re-routing, multi-stream gating, phase offset, gap recovery, logical-clock timeout, sub-Hz and non-gated ride-along, broken-stream rejection, HWM clamp, stream eviction/rejoin, batch-length changes, jitter tolerance, multi-rate robustness, and adaptive integration.

- [x] Tested in long run of Bifrost `monitor_data` service with `--batcher rate-aware`.


src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+# Must be >= ``timeout_factor`` (default 1.2) for the timeout path to
+# ever fire -- and comfortably above that for sub-Hz-only streams whose
+# sparse arrivals rely on multi-batch HWM jumps to trigger cascading
+# timeout closes of empty batches between pulses.  Three batches allows
+# one pulse's worth of HWM advance to cover the preceding empty batch,
+# matching the natural cadence of 0.5 Hz-and-below gated streams.
+_MAX_HWM_PAST_WINDOW_BATCHES = 3
+
+
+@dataclass
```
jl-wynen, line 90:
```suggestion
@dataclass(slots=True)
```
And the same for all dataclasses here. This might not matter here, but it slightly speeds up attribute access.

src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+class _GatedStream:
+    """Per-stream state for a gated stream.
+
+    Owns both persistent state (rate estimator, pulse grid, absence
+    counter) and transient per-batch state (bucketed messages and highest
+    slot seen).  The grid's Optional nature is sealed inside this class:
+    callers observe arrivals, route messages, and ask whether the gate is
+    satisfied -- they never branch on ``None``.  Transient state is reset
```
jl-wynen, line 207: So should `grid` be protected? (`_grid`) I am anyway wondering which parts of these dataclasses are a public interface and which an implementation detail.

src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+        batch_length_s = batch_length.to_seconds()
+        if int_rate * batch_length_s < 1.0:
+            self.grid = None
+            return
+        period_ns = round(1e9 / int_rate)
+        slots_per_batch = round(int_rate * batch_length_s)
+        origin = self._choose_origin(batch_start, batch_length)
+        if origin is None:
+            self.grid = None
+            return
```
jl-wynen, line 305:
```suggestion
        origin = self._choose_origin(batch_start, batch_length)
        if origin is None:
            self.grid = None
            return
        slots_per_batch = round(int_rate * batch_length_s)
```

src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+        candidate = self._pick_origin(batch_start)
+        if candidate is None or _origin_too_far(candidate, batch_start, batch_length):
+            return None
+        return candidate
+
+    def _pick_origin(self, batch_start: Timestamp) -> int | None:
+        """Pick an origin timestamp that lies on the pulse grid."""
+        if self.messages:
+            for m in self.messages:
+                if not m.timestamp < batch_start:
```
jl-wynen, line 334:
```suggestion
                if m.timestamp >= batch_start:
```

src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+        return set(self._streams)
+
+    def grid_of(self, stream_id: StreamId) -> PulseGrid | None:
+        """Return the pulse grid for a tracked stream, or ``None``."""
+        stream = self._streams.get(stream_id)
+        return stream.grid if stream is not None else None
+
+    def is_gating(self, stream_id: StreamId) -> bool:
+        """True if the stream has a converged grid and gates batch closure."""
+        return self.grid_of(stream_id) is not None
```
jl-wynen, line 412:
```suggestion
    def is_gating(self, stream_id: StreamId) -> bool:
        """True if the stream has a converged grid and gates batch closure."""
        stream = self._streams.get(stream_id)
        return stream.is_gating if stream is not None else False
```
Primarily because of the comment above about not checking whether `grid` is None.

src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+        """
+        self._pending_batch_length = Duration.from_seconds(batch_length_s)
+
+    def batch(self, messages: list[Message[Any]]) -> MessageBatch | None:
+        if messages:
+            latest = max(m.timestamp for m in messages)
+            if self._high_water_mark is None:
+                self._high_water_mark = latest
+            elif self._high_water_mark < latest:
+                self._high_water_mark = self._clamped_hwm(latest)
```
jl-wynen, line 440:
```suggestion
            self._high_water_mark = self._clamped_hwm(latest)
```

src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+        gap advance) may briefly leave HWM past ``start + cap``, which
+        the next update would otherwise regress.
+        """
+        if self._active_window is None or self._high_water_mark is None:
+            return latest
+        cap = Duration.from_ns(
+            _MAX_HWM_PAST_WINDOW_BATCHES * self._batch_length.to_ns()
+        )
+        max_allowed = self._active_window.start + cap
+        return min(latest, max(max_allowed, self._high_water_mark))
```
jl-wynen, line 474: I'm, confused about the `max` here. Doesn't this make `max_allowed` a lower bound of HWM? I thought it would be an upper bound.

src/ess/livedata/core/rate_aware_batcher.py
```diff
@@ -0,0 +1,640 @@
...
+        """True if ``msg`` belongs in a future window within the hold-back cap."""
+        if not (msg.timestamp > window.end):
+            return False
+        cap = _MAX_HWM_PAST_WINDOW_BATCHES * self._batch_length
+        return msg.timestamp - window.end <= cap
+
+    def _is_batch_complete(self, window: _ActiveWindow) -> bool:
+        if self._high_water_mark is not None:
+            threshold = window.start + Duration.from_seconds(self.timeout_s)
+            if not self._high_water_mark < threshold:
```
jl-wynen, line 532:
```suggestion
            if self._high_water_mark >= threshold:
```

jl-wynen (commented):
I can't claim to understand all of this. But the code looks fine overall based on what we discussed. I just have some minor comments.

src/ess/livedata/core/rate_aware_batcher.py
SimonHeybrock: `_GatedStream` itself is module-private. Within the module, the batcher treats `_GatedStream` as an intimate collaborator - an extra underscore on grid doesn't change what's public, since the class isn't. The `is_gating` delegation (addressed in the other comment) is now the actual seal: external callers can't observe the Optional-ness anymore.

jl-wynen (approved):

PR #883, opened 2026-04-20
Remove backwards-compatibility shims from dashboard APIs

SimonHeybrock:
Two cleanup passes over the dashboard layer, removing code paths that existed solely to support legacy formats and optional dependencies during migration.

**Drop legacy cell config and flat-dict subscriber output**

`DataSubscriber._assemble()` now always returns `dict[role, dict[ResultKey, data]]`. The single-role flat-dict shortcut (which existed for backward compatibility with standard plotters) is gone; callers extract the `primary` role explicitly. `PlotOrchestrator._parse_raw_layer` and `_parse_raw_cell` now require the current YAML format (`layers` + `data_sources` dict keyed by role); the legacy `config`, `data_sources` list, and top-level `workflow_id` branches are removed. Grid templates updated accordingly.

**Tighten dashboard APIs: remove test-only optional dependencies**

`Orchestrator`, `SessionUpdater`, and `DataService` previously accepted optional constructor arguments so tests could inject `None` and skip real dependencies. These are now required (non-optional), the `None`-guard branches are removed, and `nullcontext` fallback in `Orchestrator.ingest()` is gone. Tests updated to always supply real objects.

### Test plan

- [x] Open dashboard with existing plot configs works
- [x] Load Bifrost plot grid templates works

### DIFF
--- a/src/ess/livedata/config/instruments/bifrost/grid_templates/bragg_peak_monitor.yaml
+++ b/src/ess/livedata/config/instruments/bifrost/grid_templates/bragg_peak_monitor.yaml
@@ -9,11 +9,13 @@ cells:
     col: 0
     row_span: 2
     col_span: 2
-  config:
-    workflow_id: bifrost/monitor_data/monitor_histogram/1
-    output_name: counts_in_toa_range
-    source_names:
-    - bragg_peak_monitor
+  layers:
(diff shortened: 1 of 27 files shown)


SimonHeybrock:
LGTM

PR #885, opened 2026-04-20
Convert workflow/plot config modal to vertical tabs layout

SimonHeybrock:
## Summary

- Convert the workflow/plot configuration modal from stacked `pn.Card`s to a vertical tabs layout, with the modal title/description sticky above the tabbed body and the navigation row sticky at the bottom.
- Route error focus to the offending tab: source-selection errors activate the General tab; parameter errors activate the owning parameter tab.
- Extract shared modal sizing to `ModalSizing` in `styles.py` and HTML-escape user-visible title/description.

## Why

Stacked cards forced the user to scroll through every parameter group to reach a specific section, and validation errors deep in a card were easy to miss. Vertical tabs keep every group one click away, keep the header/footer visible, and let us focus the failing tab on validation.

### Before

screenshot-2026-04-20T13:30:30

### After

screenshot-2026-04-20T13:29:08
screenshot-2026-04-20T13:28:58
screenshot-2026-04-20T13:28:08
screenshot-2026-04-20T13:27:41

## Notes

- `ConfigurationPanel.panel` now returns the combined title + body. `split_for_sticky_header()` exposes the two slots for the modal's sticky-header layout.
- `ModelWidget` no longer owns the outer container; it exposes `param_group_tabs` so the configuration widget can compose them with source selection into a single `pn.Tabs`. Failing field names are cached from the last `validate_parameters()` to avoid a second validation pass.
- When there are no sources and no params, the body falls back to a `NoParamsWidget` directly (no empty tabs).

## Test plan

- [x] Open a workflow config modal with multiple parameter groups — tabs on the left, title/footer stay pinned while scrolling.
- [x] Clear the source selection and click the action button — General tab activates, error visible.
- [x] Enter an out-of-range value in a parameter tab, switch to General, click the action button — the offending parameter tab activates.
- [x] Open a workflow with no params  — modal shows only the general tab.
- [x] Plot-config wizard still renders correctly across all steps (back/next navigation, sticky header, scrollable body).

### DIFF
--- a/scripts/generate_widget_screenshots.py
+++ b/scripts/generate_widget_screenshots.py
@@ -92,7 +92,7 @@ def create_widget_app(
     # Create adapter (dummy start callback)
     adapter = WorkflowConfigurationAdapter(
         spec=spec,
-        persistent_config=None,
+        config_state=None,
         start_callback=lambda *args, **kwargs: None,
     )
 
--- a/src/ess/livedata/dashboard/widgets/configuration_widget.py
+++ b/src/ess/livedata/dashboard/widgets/configuration_widget.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 from __future__ import annotations
 
+import html
 from collections.abc import Callable
 from typing import Any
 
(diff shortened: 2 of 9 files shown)


SimonHeybrock:
LGTM

PR #888, opened 2026-04-22
Fix Workflows tab populating one at a time on first open

SimonHeybrock:
## Summary

- `pn.io.hold()` only works in session-callback contexts. The session factory (`create_layout`) runs outside that context, so rendering the dynamic tab content produced one PATCH-DOC per workflow widget, causing them to appear one at a time.
- `WorkflowStatusListWidget._create_panel()` now builds all `WorkflowStatusWidget` objects up front (pure Python, no document context needed) but stores their panels in `_pending_panels` rather than inserting them into the column immediately.
- `register_periodic_refresh()` schedules `_populate_pending()` via `pn.state.onload()`, which fires when the browser first connects. At that point `pn.state.curdoc` is set and `pn.io.hold()` is effective, so all workflow panels are extended into the column in a single batch.
- `WorkflowStatusWidget.__init__()` now initialises `_last_state_version` to the actual current version right after `_build_widget()`, eliminating an unnecessary full rebuild on the first periodic tick that caused a one-time flicker.

## Test plan

- [x] Open dashboard, verify Workflows tab populates all workflows at once (no sequential appearance)
- [x] Verify no flicker ~1 s after initial render
- [x] Stage/commit/stop a workflow and verify status updates still work correctly
- [x] Expand all / collapse all buttons work immediately on load

### DIFF
--- a/src/ess/livedata/dashboard/widgets/workflow_status_widget.py
+++ b/src/ess/livedata/dashboard/widgets/workflow_status_widget.py
@@ -216,11 +216,12 @@ class WorkflowStatusWidget:
         self._header: pn.Row | None = None
         self._body: pn.Column | None = None
 
-        # Track the orchestrator's workflow state version for change detection.
-        # refresh() compares this to detect structural changes (staging, commit, stop).
-        self._last_state_version: int | None = None
-
         self._build_widget()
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
LGTM

PR #897, opened 2026-04-27
Add static-flatten plotter for N-D data

SimonHeybrock:
## Summary

New \"static-flatten\" plotter: reduces N-D data (3–6 dims) to a 2D image by partitioning the input dims into an X-axis group and a Y-axis group at config time. Groups with two or more dims are flattened together; the partition is stored as a set of dim positions (not names), so saved configs survive per-instrument dim renames.

Key design points:
- Custom hover decomposes each flattened axis back into per-dim labels with coord values and units (integer position fallback when no coord is present), so the plot remains explorable despite the joined axes.
- `params_factory` hook on `PlotterEntry` (moved off `PlotterSpec`) lets the plotter narrow the `axis_x_dims` MultiSelect to the actual workflow output dim names.
- MultiSelect enforces partition validity on submit rather than disabling options dynamically, avoiding tricky interaction states.
- `DataRequirements.max_dims` is now optional (default `None` is uncapped); the flatten plotter uses `min_dims=3, max_dims=6` (2D is degenerate — no flattening would occur; 6 is scipp's dim-count ceiling).
- The `<spectral_coord>` placeholder dim in the SpectrumView output template was renamed from `spectrum` to make clear it is a runtime placeholder, not a spectrum-number axis.

*Not included*:
- Decided against doing anything particular to the flat axis - just showing array indices, no reference to any of the coords.

## Examples

screenshot-2026-04-28T12:51:18
screenshot-2026-04-28T12:52:07

## Test plan

- [x] Created a flatten plot on a 3D ; verified the X-axis dims MultiSelect shows real dim names and the resulting image axes are correct..
- [x] Hover over flattened axes; verify outer/inner labels with coord values, units, and integer-index fallback when no coord is present.
- [x] Submit an invalid partition in the MultiSelect (e.g., all dims in X); verify validation error is shown and the plot is not updated.
- [x] Open the plot config widget for a SpectrumView output; verify the placeholder dim shows as `<spectral_coord>` rather than `spectrum`.

### DIFF
--- a/src/ess/livedata/dashboard/flatten_plotter.py
+++ b/src/ess/livedata/dashboard/flatten_plotter.py
@@ -1,11 +1,11 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
-"""Plotter wrapping ImagePlotter with static (config-time) flattening of 3D data.
+"""Plotter wrapping ImagePlotter with static (config-time) flattening of N-D data.
 
-The user picks which input dim stays as one image axis; the other two are
-flattened together into the second image axis. A custom hover decomposes
-the flat-axis cell index back into outer/inner labels (with coord values
-when available) so the plot stays explorable despite the synthetic axis.
+Each image axis is built from one or more input dims, flattened together in a
+fixed order. A custom hover decomposes each axis cursor index back into
+per-dim labels (with coord values when available) so the plot stays explorable
+despite the synthetic axes.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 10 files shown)


src/ess/livedata/dashboard/flatten_plotter.py
```diff
@@ -0,0 +1,359 @@
...
+    ``axis_x_dims`` holds *positions* (0-based input-dim indices) — saving
+    integers rather than dim names keeps configs invariant under
+    per-instrument dim renames. The Y group is always the complement.
+    Ordering within each group is natural input order, with the
+    ``transpose_*_flatten`` flags reversing it when the natural order is
+    wrong; arbitrary K ≥ 3 permutations are not exposed at the params layer
+    (use the plotter's direct constructor if needed).
+    """
+
+    axis_x_dims: set[int] = pydantic.Field(
```
nvaytet, line 39: I'm wondering if instead of having `Dims on X axis` and `Transpose`, we should have `Dims on X axis` and `Dims on Y axis`?
But maybe you decided against that because there would need to be some logic about not selecting the same dim on both y and x axes?

src/ess/livedata/dashboard/flatten_plotter.py
```diff
@@ -0,0 +1,359 @@
...
+        return None
+    if data.coords.is_edges(dim):
+        # float64 to dodge scipp/3765 mixed-precision issues with midpoints
+        return sc.midpoints(coord.to(dtype='float64', copy=False), dim=dim)
+    return coord
+
+
+def _joined_dim_name(names: tuple[str, ...], existing: tuple[str, ...]) -> str:
+    """Join dim names with '·', suffixing '_' until unique against ``existing``."""
+    name = '·'.join(names)
```
nvaytet: Can we use something else than `.`? I don't find it super readable/obvious on the screenshot. I think it's easy to miss the dot and then you just think the dim is called 'arc detector number'.
Maybe `/` is better? Or something else that symbolizes better the fact that it's multiple dims flattened into one?

src/ess/livedata/dashboard/flatten_plotter.py
```diff
@@ -0,0 +1,359 @@
...
+        args={
+            'names': list(names),
+            'sizes': list(sizes),
+            'strides': strides,
+            'values_by_dim': values_by_dim,
+        },
+        code="""
+        const k = names.indexOf(format);
+        if (k < 0) return '';
+        const idx = Math.round(value);
```
nvaytet, line 199: Does this work with mesh spacing which isn't constant?

Or maybe non-regular images are not handled in the plotting anyway?

src/ess/livedata/dashboard/flatten_plotter.py
SimonHeybrock: Correct, it would require special validation logic and UX would be bad. If you look at the screenshot you can see that the current widget nicely visually partitions the dims (selected -> X, non-selected -> Y).

src/ess/livedata/dashboard/flatten_plotter.py
SimonHeybrock: Avoid `/` since I feared it might be confused with a division. Do you have other ideas?

src/ess/livedata/dashboard/flatten_plotter.py
SimonHeybrock: @nvaytet Yes — for **flattened** axes (multiple dims combined) the image always uses integer indices, so `Math.round` is correct there. For **single-dim** axes with physical coords (e.g., wavelength) it was broken: Bokeh gives `$x` in physical coord space, so `Math.round(1.5 Å)` gives 2 rather than the array index near 1.5. Fixed in the latest push: single-dim axes now use a binary search over the coord values array to find the nearest grid point, while multi-dim flattened axes keep the stride-math path.

src/ess/livedata/dashboard/flatten_plotter.py
SimonHeybrock: @nvaytet Concrete suggestion: `×` (U+00D7, multiplication sign). It conveys that the flat axis is a Cartesian product of the input dims and is visually distinct enough from a period. Alternative: `⊗` (tensor product) — more technical but unambiguous. Happy to go with whichever you prefer, or a different one if you have something in mind.

SimonHeybrock: Switched to tuple notation: `(a,b,c)`. More legible and unambiguous.

nvaytet (approved):

Issue #894, opened 2026-04-27
Support wavelength mode universally via chopperless TOF lookup table


SimonHeybrock:
## Background

Monitor workflows currently expose either `MonitorDataParams` (TOA + wavelength) or `TOAOnlyMonitorDataParams` (TOA only). The TOA-only variant is used for instruments that don't have a configured TOF lookup table (NMX, ODIN, and now ESTIA via the generic `cbm` placeholder).

`ess.reduce.unwrap.lut.LookupTableWorkflow` accepts `DiskChoppers = {}`, producing a chopperless TOA → wavelength mapping from source position, flight-path range, and pulse period alone. Validated chopperless tables already exist upstream:

- LOKI — `loki-wavelength-lookup-table-no-choppers.h5` (esssans)
- TBL — `TBL-wavelength-lookup-table-no-choppers-5m-35m.h5` (essimaging)
- ESTIA — notebook to build one (essreflectometry)

## Proposal

Support wavelength mode for any instrument by falling back to a generic chopperless TOF lookup table when no instrument-specific table is configured. This would:

- Remove `TOAOnlyMonitorDataParams` and `TOAOnlyCoordinateModeSettings`. All instruments use `MonitorDataParams`.
- Collapse the duplicated factory boilerplate in NMX/ODIN/ESTIA `factories.py` to `specs.monitor_handle.attach_factory()(create_monitor_workflow_factory)`, since the type-hint mismatch goes away.
- Make wavelength mode available for any new instrument added via the generic `cbm` placeholder.

## Open questions

- Where does the fallback table live? Built on the fly from defaults (source at origin, flight-path range, ESS pulse period), shipped as a data artifact, or generated lazily and cached?
- Should the UI flag wavelength values as "approximate, no chopper info" when the fallback is used, to avoid misleading users on instruments that *do* have choppers but no configured table?
- What flight-path range covers all ESS instruments without producing an absurdly large table?

## References

- Generic `cbm` monitor placeholder: #TBD (PR adding this introduces the TOA-only restriction for ESTIA — would be lifted by this change)
- `LookupTableWorkflow`: `ess.reduce.unwrap.lut`

nvaytet:
The issue with this is that the wavelength in a chopper-less lut would be nonsense (apart maybe for the shortest instruments, and even then...).

There would be so much overlapping regions and mixing of wavelengths that I am not sure outputting nonsense data is better than raising an error?

[...]

PR #898, opened 2026-04-28
Add one-shot job mechanism for zero-physical-stream sources

SimonHeybrock:
## Summary
- Generic mechanism for jobs whose primary source resolves to zero physical streams: finalize inline during `schedule_job`, no scheduled/active state, result emitted through the config-handler return path on the same loop tick as the ack — no dependency on data traffic for activation or emission.
- Detection in `JobFactory` via `resolve_stream_names`; one-shot jobs bypass the normal lifecycle.
- Action return type changes from `CommandAcknowledgement | None` to `list[Message]` so a single command can carry an ack plus any synchronously-produced result messages; `ConfigProcessor` extends from this list and `OrchestratingProcessor` publishes via its existing per-loop sink call.
- Preparatory work for #894 (dynamic TOF lookup-table workflow); chopperless instruments are the first concrete user.

## Test plan
- [x] Unit tests for one-shot detection in `JobFactory` (with/without stream mapping, logical→physical expansion still resolves non-empty).
- [x] Unit tests for synchronous activation in `JobManager.schedule_job` — no entry in `_scheduled_jobs`/`_active_jobs`, `drain_pending_results` semantics.
- [x] Unit tests for the new `list[Message]` action signatures in `JobManagerAdapter` and aggregation in `ConfigProcessor`.
- [x] Unit tests for `JobResult.to_message()`.
- [x] End-to-end test in `tests/services/one_shot_job_test.py`: config message → single `step()` → ack + result both reach the sink, no data ever fed in.

### DIFF
--- a/src/ess/livedata/core/job.py
+++ b/src/ess/livedata/core/job.py
@@ -12,6 +12,7 @@ import scipp as sc
 from ess.livedata.handlers.workflow_factory import Workflow
 
 from ..config.workflow_spec import JobId, ResultKey, WorkflowId
+from .message import STATUS_STREAM_ID, Message, StreamId, StreamKind
 from .timestamp import Timestamp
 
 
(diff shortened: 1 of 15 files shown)


SimonHeybrock:
Closing since we most likely will go with a different approach that does not need this (synthetic chopper-cascade trigger, see #902).

PR #896, opened 2026-04-27
Add ESTIA detector rotation f144 stream

SimonHeybrock:
## Summary

Expose the multiblade detector bank rotation (`ESTIA-DtRot:MC-RotZ01:Mtr.RBV` on topic `estia_motion`) as a timeseries so users can plot it in the dashboard. The timeseries workflow spec is auto-registered from `f144_attribute_registry` by `Instrument.__post_init__`.

Scope is intentionally narrow: log stream + timeseries plotting only. Sample-detector distance is invariant under this rotation, so the value is not (yet) needed for wavelength reconstruction or motion-aware geometry.

## Notes

PV channel and topic taken from the f144 writer attributes on the NXpositioner NXlog `/entry/instrument/detector_arm/detector_rotation/value` in `coda_estia_999999_00027641.hdf`. The same positioner also publishes `.VAL` (setpoint) and `.DMOV` (done-moving) on the same topic; not exposed here.

A separate file-writer issue was flagged upstream: the chain entry at `multiblade_detector/transformations/detector_rotation` is a duplicate NXlog whose `@source` lacks the `.RBV` suffix and would never be populated. Doesn't affect this change (we ingest from Kafka directly), but it would block a LOKI-style motion-aware geometry wiring later.

PR #900, opened 2026-04-28
Estia live reduction workflows

### LINKED ISSUE #899: Data reduction workflow for ESTIA
We should add a couple of data reduction workflows for ESTIA:

* Intensity over `Q`, created by histogramming `ReducibleData[SampleRun]` over Q.
* Intensity over `wavelength` and `theta`, created by histogramming `ReducibleData[SampleRun]` over wavelength and theta (angle of reflection).

For testing [the mcstas data filled nexus files](https://scipp.github.io/ess/reflectometry/generated/modules/ess.estia.data.estia_mcstas_nexus_sample_example.html) might be useful.

jokasimr:
Fixes #899

### DIFF
--- a/src/ess/livedata/config/instruments/estia/factories.py
+++ b/src/ess/livedata/config/instruments/estia/factories.py
@@ -68,12 +68,9 @@ def setup_factories(instrument: Instrument) -> None:
         wf[QBins] = params.q_edges.get_edges()
         wf[ThetaBins[SampleRun]] = params.theta_edges.get_edges()
 
-        wf[YIndexLimits] = (sc.scalar(0), sc.scalar(63))
-        wf[ZIndexLimits] = (sc.scalar(0), sc.scalar(1535))
-        wf[BeamDivergenceLimits] = (
-            sc.scalar(-0.75, unit='deg'),
-            sc.scalar(0.75, unit='deg'),
-        )
+        wf[YIndexLimits] = params.y_index_limits.get_limits()
+        wf[ZIndexLimits] = params.z_index_limits.get_limits()
+        wf[BeamDivergenceLimits] = params.beam_divergence_limits.get_limits()
 
         IntensityThetaWavelength = NewType('IntensityThetaWavelength', sc.DataArray)
         IntensityWavelength = NewType('IntensityWavelength', sc.DataArray)
--- a/src/ess/livedata/config/instruments/estia/factories.py
+++ b/src/ess/livedata/config/instruments/estia/factories.py
@@ -14,10 +14,65 @@ def setup_factories(instrument: Instrument) -> None:
 
     The multiblade detector view (with its spectrum output) is wired via
     ``add_logical_view`` in ``specs.py``. The generic ``cbm`` monitor workflow
-    factory is attached here.
+    factory and live reflectometry diagnostics are attached here.
     """
+    from typing import NewType
+
+    import sciline.typing
+    import scipp as sc
+    from ess.estia import EstiaWorkflow, data as estia_data
+    from ess.reduce.nexus.types import NeXusData
+    from ess.reflectometry.types import (
+        BeamDivergenceLimits,
+        CorrectionsToApply,
+        DetectorRotation,
+        Filename,
+        LookupTableFilename,
+        LookupTableRelativeErrorThreshold,
+        NeXusDetectorName,
+        ProtonCurrent,
+        QBins,
+        ReducibleData,
+        SampleRun,
+        SampleRotation,
+        ThetaBins,
+        WavelengthBins,
+        YIndexLimits,
+        ZIndexLimits,
+    )
+    from scippnexus import NXdetector
+
+    from ess.livedata.handlers.detector_data_handler import get_nexus_geometry_filename
     from ess.livedata.handlers.monitor_workflow import create_monitor_workflow
     from ess.livedata.handlers.monitor_workflow_specs import TOAOnlyMonitorDataParams
+    from ess.livedata.handlers.stream_processor_workflow import StreamProcessorWorkflow
+
+    IntensityWavelength = NewType('IntensityWavelength', sc.DataArray)
+    IntensityQ = NewType('IntensityQ', sc.DataArray)
+    IntensityThetaWavelength = NewType('IntensityThetaWavelength', sc.DataArray)
+
+    def _intensity_over_wavelength(
+        events: ReducibleData[SampleRun],
+        wavelength_bins: WavelengthBins,
+    ) -> IntensityWavelength:
+        return IntensityWavelength(events.hist(wavelength=wavelength_bins, dim=events.dims))
+
+    def _intensity_over_q(
+        events: ReducibleData[SampleRun],
+        q_bins: QBins,
+    ) -> IntensityQ:
+        return IntensityQ(events.hist(Q=q_bins, dim=events.dims))
+
+    def _intensity_over_theta_wavelength(
+        events: ReducibleData[SampleRun],
+        wavelength_bins: WavelengthBins,
+        theta_bins: ThetaBins[SampleRun],
+    ) -> IntensityThetaWavelength:
+        return IntensityThetaWavelength(
+            events.hist(theta=theta_bins, wavelength=wavelength_bins, dim=events.dims).transpose(
+                ['theta', 'wavelength']
+            )
+        )
 
     @specs.monitor_handle.attach_factory()
     def _monitor_workflow_factory(source_name: str, params: TOAOnlyMonitorDataParams):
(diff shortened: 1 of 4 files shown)


src/ess/livedata/config/instruments/estia/specs.py
```diff
@@ -18,35 +21,111 @@
...
-# (done-moving) on the same topic; not exposed here.
-f144_log_streams = {
-    'detector_rotation': {
-        'source': 'ESTIA-DtRot:MC-RotZ01:Mtr.RBV',
-        'topic': 'estia_motion',
-        'units': 'deg',
-    },
-}
+
+class ThetaEdges(parameter_models.EdgesModel):
```
SimonHeybrock: Could put this into `parameter_models.py` as it is useful in general.

src/ess/livedata/config/instruments/estia/specs.py
```diff
@@ -18,35 +21,111 @@
...
+            unit=parameter_models.WavelengthUnit.ANGSTROM,
+        ),
+    )
+    q_edges: parameter_models.QEdges = pydantic.Field(
+        title='Q bins',
+        description='Define the bin edges for intensity histograms in Q.',
+        default=parameter_models.QEdges(
+            start=0.01,
+            stop=0.5,
+            num_bins=401,
```
SimonHeybrock:
```suggestion
            num_bins=400,
```
note this model takes `num_bins`, not `num_edges` like a raw `scipp.linspace`.

src/ess/livedata/config/instruments/estia/specs.py
```diff
@@ -18,35 +21,111 @@
...
+        },
+    )
+
+
+class EstiaLiveDiagnosticsOutputs(WorkflowOutputsBase):
+    """Outputs for the ESTIA live diagnostics workflow."""
+
+    i_of_wavelength: sc.DataArray = pydantic.Field(
+        default_factory=_make_wavelength_template,
+        title='I(wavelength)',
```
SimonHeybrock: Fine, but can consider using unicode lambda as `title` is used for user display. Same below.

src/ess/livedata/config/instruments/estia/specs.py
```diff
@@ -78,3 +157,16 @@ def _estia_spectrum_transform(histogram: sc.DataArray) -> sc.DataArray:
         extra_description='Summed across strips, yielding per-blade, per-wire spectra.',
     ),
 )
+
+live_diagnostics_handle = instrument.register_spec(
+    name='live_diagnostics',
+    version=1,
+    title='Live diagnostics',
```
SimonHeybrock: What is actually diagnostics about this? Is it not an actual reduction workflow (minus some normalization)?

src/ess/livedata/config/instruments/estia/factories.py
```diff
@@ -27,3 +81,39 @@ def _monitor_workflow_factory(source_name: str, params: TOAOnlyMonitorDataParams
...
+        wf[LookupTableRelativeErrorThreshold] = {source_name: float('inf')}
+        wf[WavelengthBins] = params.wavelength_edges.get_edges()
+        wf[QBins] = params.q_edges.get_edges()
+        wf[ThetaBins[SampleRun]] = params.theta_edges.get_edges()
+        wf[YIndexLimits] = (sc.scalar(0), sc.scalar(63))
+        wf[ZIndexLimits] = (sc.scalar(0), sc.scalar(1535))
+        wf[BeamDivergenceLimits] = (
+            sc.scalar(-0.75, unit='deg'),
+            sc.scalar(0.75, unit='deg'),
+        )
```
SimonHeybrock: Can you clarify what these are? Why is it not user inputs via the params model?

src/ess/livedata/config/instruments/estia/factories.py
```diff
@@ -14,10 +14,64 @@ def setup_factories(instrument: Instrument) -> None:
...
+    def _intensity_over_theta_wavelength(
+        events: ReducibleData[SampleRun],
+        wavelength_bins: WavelengthBins,
+        theta_bins: ThetaBins[SampleRun],
+    ) -> IntensityThetaWavelength:
+        return IntensityThetaWavelength(
+            events.hist(
+                theta=theta_bins, wavelength=wavelength_bins, dim=events.dims
+            ).transpose(['theta', 'wavelength'])
+        )
```
SimonHeybrock: Assuming `ReducibleData` is event mode(?), I don't think this will work:
- Calling `hist` *after* the accumulator (configured below to be `ReducibleData[SampleRun]`) will mean the cost to process a batch of messages (usually 1 second of event data) will grow linearly. Depending on the event rate this will go wrong sooner or later.
- If `ReducibleData` stores event data then the memory usage will also grow linearly.

Conclusion:
- Accumulators should almost always be histogrammed data.
- In this case, can you define three accumulators for the 3 output histograms?

SimonHeybrock (commented):
Good start. Minor comments below, and one main one about the accumulation:

src/ess/livedata/config/instruments/estia/specs.py
jokasimr: The purpose is to see if everything looks as expected, if the sample is aligned, if something is wrong, to figure out how the experiment is going etc. I'd call that diagnostics, but I'm fine with another name too if you have one in mind.

src/ess/livedata/config/instruments/estia/factories.py
```diff
@@ -88,23 +60,59 @@ def _live_diagnostics_workflow_factory(
...
+            graph: CoordTransformGraph[SampleRun],
+        ):
+            return (
+                intensity.assign_coords(
+                    wavelength=sc.midpoints(intensity.coords['wavelength']),
+                    theta=sc.midpoints(intensity.coords['theta']),
+                )
+                .transform_coords('Q', graph=graph)
+                .hist(Q=q_bins)
+            )
```
SimonHeybrock, line 121: This worries me a bit, as probably many papers have been written about how to do this correctly. Starting with NaN or out-of-range values in the 2D to various normalization issues. Is this good enough for what is needed here?

jokasimr: Can you clarify more precisely what issues you are hinting at?
I can't tell what aspect you are concerned about.

SimonHeybrock: I am concerned about how users might interpret this quantity, given that it is completely unclear (I think) of what is means. See for example https://docs.mantidproject.org/nightly/algorithms/Q1D-v2.html#normalizing-the-input-workspace (which is conceptually a similar operation) and the things handled there. So my worry is whether this histogram (as computed here) is a meaningful quantity.

src/ess/livedata/config/instruments/estia/specs.py
SimonHeybrock: Arguably such diagnostics is the purpose of every single workflow we have in ESSlivedata.

Please find something more meaningful. The "Live" bit is also redundant.

jokasimr:
Looks fine!

Screenshot from 2026-04-30 15-38-33

src/ess/livedata/config/instruments/estia/factories.py
```diff
@@ -27,3 +56,83 @@ def _monitor_workflow_factory(source_name: str, params: TOAOnlyMonitorDataParams
...
             coordinate_mode='toa',
         )
+
+    @specs.reflectometry_reduction_handle.attach_factory()
+    def _reflectometry_reduction_workflow_factory(
+        source_name: str,
+        params: specs.EstiaReflectometryReductionParams,
+    ) -> StreamProcessorWorkflow:
+        wf = EstiaWorkflow()
+        wf[Filename[SampleRun]] = estia_data.estia_mcstas_nexus_sample_example(
```
jokasimr, line 68: Eventually we should use another file here.

This nexus file is one I took from Coda not so long ago, and I added a `sample_rotation` log to it, which is currently missing in the real Coda files, but we need it for reduction.

src/ess/livedata/config/instruments/estia/factories.py
jokasimr: We talked about this in person and concluded that it is a meaningful quantity, but like always you have to be aware of what it represents.
It's very likely we have to iterate a bit on this to figure out exactly what they want to see here.

To make sure that is the case, we should probably expand the "documentation string for this quantity a bit.

src/ess/livedata/config/instruments/estia/factories.py
jokasimr: Fixed!

src/ess/livedata/config/instruments/estia/factories.py
SimonHeybrock: We should not use a wrong rotation log here, especially since it is hidden deep down and not obvious. Can you add the sample angle as an aux stream? You already posted a table of logs found in the CODA file: https://github.com/scipp/esslivedata/pull/896#issuecomment-4326712626.

src/ess/livedata/config/instruments/estia/specs.py
```diff
@@ -18,35 +21,161 @@
...
+    )
+    i_of_q: sc.DataArray = pydantic.Field(
+        default_factory=_make_q_template,
+        title='I(Q)',
+        description='Detector intensity histogrammed by momentum transfer Q.',
+    )
+    i_of_theta_wavelength: sc.DataArray = pydantic.Field(
+        default_factory=_make_theta_wavelength_template,
+        title='I(θ, λ)',
+        description='Detector intensity histogrammed by θ and wavelength λ.',
```
SimonHeybrock: You explained to me that this is not normalized. Is the `I(...)` (and "intensity" wording) the correct notation then, or is it misleading?

src/ess/livedata/config/instruments/estia/factories.py
jokasimr: But the sample rotation is not in that list of logs :/ 
And I don't know where it will eventually be added.

How about I add a TODO instead, and/or create an issue for it?

src/ess/livedata/config/instruments/estia/specs.py
jokasimr: It is partly normalized, it is normalized by footprint correction and proton charge (but the proton charge is fake atm.)

I don't know why `I` could only ever represent normalized quantities, that is not a notation I've seen anywhere else.
If anything I'd expect `I` to mean a non-normalized quantity, and `S` to mean a generic (fully normalized) scattering cross-section, but I guess people use different notations.

Eventually we want to:

1. also correct `I` by the monitor, it will then only be missing the normalization by the reference
2. add a "fully normalized" quantity called `R` (should represent the reflectivity).

MridulS (approved):
merging this in for now, for the estia test on Monday.

PR #901, opened 2026-04-29
Drop `WorkflowId.namespace`, introduce `WorkflowGroup`

SimonHeybrock:
## Summary

Reduces `WorkflowId` identity from `(instrument, namespace, name, version)` to `(instrument, name, version)` and replaces the `namespace` field with a display-oriented `WorkflowGroup`.

## Why

`WorkflowSpec.namespace` was carrying three unrelated responsibilities: workflow identity, the routing key `JobFactory` used to accept/reject jobs, and the dashboard's display category. Splitting them:

- Routing now uses an explicit `service` tag tracked by `WorkflowFactory`, consulted via `get_service(workflow_id)`. The service's own name is the comparison key, no longer read from `WorkflowId`.
- Display grouping becomes a typed `WorkflowGroup` with canonical `REDUCTION` / `MONITORS` / `DETECTORS` / `TIMESERIES` constants and a `Literal`-locked `name` to keep ad-hoc categories from sneaking in.
- Identity drops the field entirely; `WorkflowId` strings are now 3-part (`instrument/name/version`).

With the awkward semantics gone, the upcoming move into `ess.livedata.schemas` becomes mechanical.

## Note

- **No wire-format compatibility.** `JobStatusPayload` workflow-id strings, `ResultKey` JSON, and persisted configs all pick up the 3-part form.

### DIFF
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -259,6 +259,7 @@ class Instrument:
         source_names: Sequence[str],
         transform: Callable[[sc.DataArray, str], sc.DataArray] | None = None,
         namespace: str = 'detector_data',
+        service: str | None = None,
         roi_support: bool = True,
         output_ndim: int | None = None,
         reduction_dim: str | list[str] | None = None,
@@ -322,6 +323,7 @@ class Instrument:
         params = make_detector_view_params(spectrum_view=spectrum_view)
         handle = self.register_spec(
             namespace=namespace,
+            service=service,
             name=name,
             version=1,
             title=title,
@@ -351,6 +353,7 @@ class Instrument:
         self,
         *,
         namespace: str = 'data_reduction',
+        service: str | None = None,
         name: str,
         version: int,
         title: str,
@@ -372,6 +375,10 @@ class Instrument:
         ----------
         namespace:
             Namespace for the workflow (default: 'data_reduction').
+        service:
+            Name of the backend service responsible for running this workflow.
+            Defaults to ``namespace`` (service and namespace match today; the
+            coupling is removed in a later stage).
         name:
             Name to register the workflow under.
         version:
@@ -415,7 +422,7 @@ class Instrument:
             outputs=outputs,
             reset_on_run_transition=reset_on_run_transition,
         )
-        return self.workflow_factory.register_spec(spec)
+        return self.workflow_factory.register_spec(spec, service=service)
 
     def load_factories(self) -> None:
         """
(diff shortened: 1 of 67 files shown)


src/ess/livedata/config/instrument.py
```diff
@@ -289,9 +289,10 @@ def add_logical_view(
             If reduction_dim is specified, the transform should NOT include
             summing - that is handled separately to enable proper ROI index mapping.
             If None, identity (no reshaping).
-        namespace:
-            Service namespace this view belongs to. Determines which service
-            runs the workflow (e.g. ``'detector_data'`` or ``'monitor_data'``).
+        group:
+            Display-oriented :class:`WorkflowGroup` this view belongs to.
+            The group's ``name`` (e.g. ``'detector_data'`` or ``'monitor_data'``)
+            also determines which service runs the workflow today.
```
jl-wynen: Missing `service`.

src/ess/livedata/config/instrument.py
```diff
@@ -370,8 +373,13 @@ def register_spec(
...
         ----------
-        namespace:
-            Namespace for the workflow (default: 'data_reduction').
+        group:
+            Display-oriented :class:`WorkflowGroup` for the workflow
+            (default: ``REDUCTION``).
+        service:
+            Name of the backend service responsible for running this workflow.
+            Defaults to ``group.name`` (service and group identifier match
+            today; the coupling is removed in a later stage).
```
jl-wynen: > (service and group identifier match today; the coupling is removed in a later stage)

This doesn't sound like an explanation for a user but a commit/PR comment.

Is it correct that currently `service == group.name` everywhere?

jl-wynen (commented):
I think it makes sense to split display grouping and job assignment.
Can you remind me what a 'service' is exactly?

Is it safe to drop the 'namespace' / 'service' from `WorkflowId`? Or can there be clashes now?

SimonHeybrock:
> I think it makes sense to split display grouping and job assignment. Can you remind me what a 'service' is exactly?

Backend is split into shards ("services"), handling different subsets of workflows.

> Is it safe to drop the 'namespace' / 'service' from `WorkflowId`? Or can there be clashes now?

Did check for clashing during this work and found none. Option for clashes existed even before, now we simply have a larger set of workflows that need unique names.

jl-wynen (approved):
I think it makes sense to split display grouping and job assignment.
Can you remind me what a 'service' is exactly?

Is it safe to drop the 'namespace' / 'service' from `WorkflowId`? Or can there be clashes now?

PR #902, opened 2026-04-29
Add chopperless wavelength lookup-table workflow

SimonHeybrock:
## Summary

MVP for the dynamic TOF lookup-table workflow (issue #894) — chopperless instruments only. Added in `timeseries service as it is conceptually similar. Framework pieces (synthetic primary stream, ChopperSynthesizer wrapper, context-replay-on-job-start, sciline-driven LUT computation, da00 publish with provenance coords) are exercised against a real upstream pipeline. Chopper-equipped support (NeXus geometry loading, plateau detection on phase NXlogs, RawChoppers→DiskChoppers translation) is deliberately out of scope for this PR.

## Examples

settings-dialog
wavelength-table-plotted-in-dashboard

## What's in

- New `wavelength_lut` service: subscribes to commands/run-control, runs `ess.reduce.unwrap.lut.LookupTableWorkflow()` with `DiskChoppers={}`, publishes a single `wavelength_lut` da00 message per fire.
- `ChopperSynthesizer`: `MessageSource` decorator; v0 emits one synthetic `chopper_cascade` LogData(value=1) tick on the first `get_messages` call. Conceptually the same shape as an f144 boolean "at-setpoint" signal — `ToNXlog` accumulates it as context, `MessagePreprocessor.get_context` replays the cached value when the operator schedules a job.
- `Pulse(frequency: float = 14.0, stride: int = 1)` replaces a period-in-ms input — no more typing `71.4286` for 14 Hz.
- `outer_source_wrapper` hook on `DataServiceBuilder` so the synthesizer can sit between the Kafka adapter and the orchestrator.
- Workflow registered for both `dummy` (used in CI tests) and `loki` (chopperless against the upstream `loki_tof_lookup_table_no_choppers` reference).

## Out of scope (deferred to follow-up)

- Plateau detection on phase NXlogs and per-chopper `phase_setpoint` synthesis
- `NXdisk_chopper` copy in `make_geometry_nexus.py` and per-instrument geometry artifact regen
- `RawChoppers + cached aux setpoints → DiskChoppers` translation provider
- Real `SourcePosition` from NeXus (currently a placeholder vector — only used inside the upstream simulator's per-chopper loop, which is empty for chopperless)
- f144 routing extension for the chopper topic and per-instrument chopper-PV stream declarations


src/ess/livedata/config/instruments/loki/factories.py
```diff
@@ -126,6 +126,17 @@ def _detector_view_workflow_factory(
...
+        create_chopperless_wavelength_lut_workflow,
+    )
+    from ess.livedata.handlers.wavelength_lut_workflow_specs import (
+        WavelengthLutParams,
+    )
+
+    @specs.wavelength_lut_handle.attach_factory()
+    def _wavelength_lut_workflow_factory(params: WavelengthLutParams):
+        """Factory for LOKI's chopperless wavelength lookup-table workflow."""
+        return create_chopperless_wavelength_lut_workflow(params=params)
```
nvaytet, line 139: Here it is hard-coded to be chopperless. Is the plan to then update this once we have support for reading chopper info and build a table on the fly?

src/ess/livedata/config/instruments/loki/factories.py
```diff
@@ -126,6 +126,17 @@ def _detector_view_workflow_factory(
...
 
     @specs.monitor_handle.attach_factory()
     def _monitor_workflow_factory(source_name: str, params: MonitorDataParams):
```
nvaytet, line 142: Why was it only added to dummy and Loki instruments?
Is it just for testing, and we will roll out to other instruments later?

src/ess/livedata/handlers/timeseries_handler.py
```diff
@@ -12,13 +12,25 @@
...
 logger = structlog.get_logger(__name__)
 
+#: Attributes used by ``ToNXlog`` for synthetic LOG streams that are not real
+#: upstream f144s and therefore not in :attr:`Instrument.f144_attribute_registry`
+#: (registering them there would also register unwanted timeseries workflows).
+#: The ``chopper_cascade`` tick from :class:`ChopperSynthesizer` is the only
+#: such stream today; its value is a boolean "at setpoint" indicator and
+#: carries no unit (distinct from ``'dimensionless'``, which a number with
+#: cancelled units would have).
+_SYNTHETIC_LOG_ATTRS: dict[str, dict[str, Any]] = {
```
nvaytet, line 30: So the idea would be that this single-tick stream would be later replaced by listening to the chopper info streams?

src/ess/livedata/handlers/wavelength_lut_workflow.py
```diff
@@ -0,0 +1,122 @@
...
+    wf = LookupTableWorkflow()
+    wf[PulsePeriod] = params.pulse.get_period()
+    wf[PulseStride] = int(params.pulse.stride)
+    wf[DistanceResolution] = params.distance_resolution.get()
+    wf[TimeResolution] = params.time_resolution.get()
+    wf[LtotalRange] = (
+        params.distance_range.get_start(),
+        params.distance_range.get_stop(),
+    )
+    wf[NumberOfSimulatedNeutrons] = int(params.simulation.num_simulated_neutrons)
```
nvaytet, line 98: Did I understand correctly that if you want to play with these params, you have to stop the job/workflow, start a new job and change the number of neutrons?
I don't know how often users would play with the params, probably not very often at all, and just use the defaults?

src/ess/livedata/handlers/wavelength_lut_workflow_specs.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def get(self) -> sc.Variable:
+        return sc.scalar(self.value, unit=self.unit.value)
+
+
+class LtotalRange(RangeModel):
+    """Range of total flight paths covered by the lookup table."""
+
+    start: float = pydantic.Field(default=5.0, description="Shortest L_total.")
+    stop: float = pydantic.Field(default=30.0, description="Longest L_total.")
```
nvaytet, line 72: For loki, we go up to 35m in `esssans`. Maybe have the default cover the same range?

This is a general parameter for all instruments, where one single default is not really super adapted...
Can we have a default per instrument?

src/ess/livedata/handlers/wavelength_lut_workflow_specs.py
```diff
@@ -0,0 +1,178 @@
...
+        version=1,
+        title='Wavelength lookup table',
+        description=(
+            'Compute a wavelength lookup table from the current chopper-cascade '
+            'configuration. Refires when chopper setpoints change.'
+        ),
+        source_names=[CHOPPER_CASCADE_SOURCE],
+        params=params,
+        outputs=WavelengthLutOutputs,
+        reset_on_run_transition=False,
```
nvaytet, line 177: Can you briefly explain what `reset_on_run_transition` does?

src/ess/livedata/config/instruments/loki/factories.py
SimonHeybrock: Yes.

src/ess/livedata/config/instruments/loki/factories.py
SimonHeybrock: Just for testing, yes.

src/ess/livedata/handlers/timeseries_handler.py
SimonHeybrock: Yes, basically I want to synthesize an "everything is in phase" signal (or maybe "there is a new setpoint and we wnet out of phase for the previous one", so it gets computed as soon as possible).

src/ess/livedata/handlers/wavelength_lut_workflow.py
SimonHeybrock: Yes, parameter changes require restarts. This means files have to be read again, but hopefully just reading the choppers is quick.

> I don't know how often users would play with the params, probably not very often at all, and just use the defaults?

Yes, there are defaults. The user can thus just click "Start" without changing them, if they don't want to.

src/ess/livedata/handlers/wavelength_lut_workflow_specs.py
SimonHeybrock: We would need to create instrument-specific params, yes. But actually me idea was to either (1) hard-code it per instrument or (2) read min/max distance from the NeXus geometry file (the one which will also be used to read chopper info).

src/ess/livedata/handlers/wavelength_lut_workflow_specs.py
SimonHeybrock: It clears the accumulators in the workflow when there is a run-start or run-stop message from NICOS. This workflow currently has no accumulators so this is not actually relevant I think.

src/ess/livedata/handlers/wavelength_lut_workflow_specs.py
nvaytet: Hard coding per instrument sounds fine to me.

nvaytet (approved):

PR #904, opened 2026-04-29
Fix plot config dialog ignoring plotter change in edit mode

SimonHeybrock:
When reconfiguring an existing plot via the gear menu, navigating back to step 2 to select a different plotter and then proceeding to step 3 had no effect — the original plotter was always used.

**Root cause:** `Wizard.reset_to_step(n)` initialised `_step_results = []`. When the user navigated back from step `n` to step `n-1` and committed, the result was appended at list index 0. On re-entering step `n`, `_update_content` looked for `_step_results[n-1]`, found nothing, and passed `input_data=None` to `on_enter`. `SpecBasedConfigurationStep.on_enter` then fell back to `initial_config`, silently discarding the user's new selection.

**Fix:** `reset_to_step(n)` now pre-pads `_step_results` with `[None] * n`, so committed results land at the correct index regardless of where the wizard started.

Regression test added to `wizard_test.py`.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/wizard.py
+++ b/src/ess/livedata/dashboard/widgets/wizard.py
@@ -239,7 +239,9 @@ class Wizard:
 
         self._current_step_index = step_index
         self._finished = False
-        self._step_results = []
+        # Pad with None so result indices align with step indices when the user
+        # navigates back and commits a step that was skipped on entry.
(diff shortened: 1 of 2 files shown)


nvaytet:
The joys of developing/maintaining GUIs...

nvaytet (approved):

PR #909, opened 2026-04-30
fix: meaning of `start` and `stop` in bin-edge widget with log scale

### LINKED ISSUE #908: Logscale should use `np.geomspace` instead of `np.logspace`
See current implementation here:
https://github.com/scipp/esslivedata/blob/ccb1ab0d9a52802b274f4464fc6cac89f322c2aa/src/ess/livedata/parameter_models.py#L236-L241

When the scale is set to `scale.LOG` the `start` and `stop` entries in the selector represent `log10(start)` and `log10(stop)` of the range.
To make the `start` and `stop` represent the start and stop of the desired range we should use `np.geomspace`.

jokasimr:
Fixes #908 

Bug in the edge-widget in the workflow configuration page:

When the scale was set to scale.LOG the `start` and `stop` entries in the selector represented `log10(...)` of the start and stop of the desired range.

With this fix the `start` and `stop` entries in the widget represent the start and stop value of the range, which is the behavior users would expect.

### DIFF
--- a/src/ess/livedata/parameter_models.py
+++ b/src/ess/livedata/parameter_models.py
@@ -235,7 +235,7 @@ class EnergyEdges(EdgesModel):
 
 def make_edges(*, model: EdgesModel, dim: str, unit: str) -> sc.Variable:
     """Convert the edges to a scipp variable."""
-    op = {Scale.LINEAR: sc.linspace, Scale.LOG: sc.logspace}[model.scale]
+    op = {Scale.LINEAR: sc.linspace, Scale.LOG: sc.geomspace}[model.scale]
     return op(
         dim=dim, start=model.start, stop=model.stop, num=model.num_bins + 1, unit=unit
     )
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
~Can you include some before/after screenshots in the PR description?~

Got confused due to the lack of description, and because my brain read your issue title and said "Logscale" -> it is about plotting. But this is not about logscale, right? It is about how define define/configure edges to use in a workflow.

Can you please clarify issue and PR titles and add a description in the PR itself? We are using those to create user-facing docs about changes.

jokasimr:
> Got confused due to the lack of description

There is a description in the issue?


> Can you please clarify issue and PR titles and add a description in the PR itself? We are using those to create user-facing docs about changes.

I've added the description to the PR and clarified the title.

SimonHeybrock (approved):

PR #915, opened 2026-05-05
Add WorkflowId migration script for dashboard configs

SimonHeybrock:
PR #901 dropped the namespace segment from WorkflowId. Existing dashboard config YAMLs still hold 4-part IDs; this script rewrites them to the new 3-part form, with .bak backups and atomic writes. Idempotent.


SimonHeybrock:
@MridulS Should run this script on all "dashboard" deployments after shutting down pre 26.5 and before starting 26.5, if possible.

MridulS:
Why are we running this script again? Is this in case someone has created a dashboard and we want to update it? Aren't the standard configs already in the repo?

SimonHeybrock:
> Why are we running this script again? Is this in case someone has created a dashboard and we want to update it? Aren't the standard configs already in the repo?

It is to keep the current configs that where not added as templates yet (templates were migrated in the PR that made the breaking change). I don't think anything major was lost, but I thought it would be nice to do this.

MridulS (approved):

PR #916, opened 2026-05-05
Geometry script: handle NXdisk_chopper and trim NXlogs

SimonHeybrock:
## Summary

- Add an `NXdisk_chopper` branch to `write_minimal_geometry` so the geometry artifact carries chopper static fields (slits, radius, slit_edges, delay, etc.) and NXlog placeholders for streamed values (rotation_speed, phase). Required for the upcoming dynamic wavelength lookup-table workflow which feeds chopper geometry into `DiskChopper.from_nexus`.
- Trim every NXlog group to length-0 datasets that preserve dtype, units, and attributes — the artifact must never carry historical samples regardless of what the source contains. Production NeXus files with populated motor readbacks or chopper logs were previously bloating the output via `_resolve_depends_on_chains` and the `NXtransformations` branch.
- Make `NXtransformations` and `NXdisk_chopper` branches idempotent. ESTIA exposed a latent bug where a recursive NXlog copy pulls in a nested `NXtransformations` before the visitor's dedicated branch reaches it; reuse the existing destination group when present and only copy missing children.

Verified end-to-end on real LOKI / ESTIA / ODIN files: 4–10 NXdisk_chopper groups preserved per file, output 1.4–1.7 MB regardless of input size (largest input 2.4 GB → 1.5 MB).

## Test plan

- [x] Unit tests cover NXdisk_chopper static fields, NXlog trimming inside choppers, NXlog inside NXtransformations, motor logs reached via depends_on, and the nested NXtransformations-in-NXlog case.
- [x] Manual: ran the script against `coda_loki_999999_00026352.hdf` (4 choppers), `coda_estia_999999_00027641.hdf` (1), `coda_estia_999999_00000814.hdf` (1, regression case for the idempotence fix), `odin_chopper_test_424250_00000034.hdf` (10), `coda_odin_999999_00000800.hdf` (10).

### DIFF
--- a/src/ess/livedata/scripts/make_geometry_nexus.py
+++ b/src/ess/livedata/scripts/make_geometry_nexus.py
@@ -45,6 +45,47 @@ def _copy_monitor_fields(src_group: h5py.Group, dst_group: h5py.Group) -> None:
     src_group.copy('depends_on', dst_group)
 
 
+def _copy_nxlog_placeholder(src: h5py.Group, dst: h5py.Group) -> None:
+    """Copy an NXlog group structure with sample arrays truncated to length 0.
+
+    Production NeXus files may contain logs with thousands of samples; the
+    geometry artifact wants only the schema (dtype, units, attributes) so
+    downstream consumers see the field shape without historical data.
+    Scalar children (e.g. ``average_value``) are copied verbatim.
+    """
+    _copy_attributes(src, dst)
+    for child_name, child in src.items():
+        if isinstance(child, h5py.Dataset) and child.ndim >= 1:
+            ds = dst.create_dataset(
+                child_name,
+                shape=(0, *child.shape[1:]),
+                dtype=child.dtype,
+            )
+            _copy_attributes(child, ds)
+        else:
+            src.copy(child_name, dst)
+
+
+def _nx_class(obj: h5py.Group | h5py.Dataset) -> str:
+    nx_class = obj.attrs.get('NX_class', b'')
+    if isinstance(nx_class, bytes):
+        nx_class = nx_class.decode()
+    return nx_class
+
+
+def _copy_child(src: h5py.Group, key: str, dst: h5py.Group) -> None:
+    """Copy ``src[key]`` into ``dst``, trimming NXlog children to length-0.
+
+    No NXlog is ever copied verbatim: the geometry artifact carries only
+    schema (dtype, units, attributes), never historical samples.
+    """
+    child = src[key]
+    if isinstance(child, h5py.Group) and _nx_class(child) == 'NXlog':
+        _copy_nxlog_placeholder(child, dst.create_group(key))
+    else:
+        src.copy(key, dst)
+
+
 def _read_depends_on(value: bytes | str) -> str | None:
     if isinstance(value, bytes):
         value = value.decode()
@@ -89,8 +130,10 @@ def _resolve_depends_on_chains(fin: h5py.File, fout: h5py.File) -> None:
 
     After copying geometry components, ``depends_on`` chains may reference
     nodes that were not copied — for example an NXlog group inside an
-    NXpositioner that acts as a transformation node.  This function follows
-    all chains and copies missing nodes (groups and datasets) as-is.
+    NXpositioner that acts as a transformation node. NXlog groups are
+    copied as length-0 placeholders so historical motor samples in the
+    source do not bloat the geometry artifact; everything else is copied
+    as-is.
     """
     resolved: set[str] = set()
     while True:
@@ -104,7 +147,8 @@ def _resolve_depends_on_chains(fin: h5py.File, fout: h5py.File) -> None:
                 continue
             _ensure_parent_groups(fin, fout, path)
             parent_path = path.rsplit('/', 1)[0] if '/' in path else ''
-            fin.copy(path, fout[parent_path or '/'])
+            leaf = path.rsplit('/', 1)[-1]
+            _copy_child(fin[parent_path or '/'], leaf, fout[parent_path or '/'])
 
 
 def write_minimal_geometry(
@@ -153,9 +197,24 @@ def write_minimal_geometry(
                         ensure_parent_groups(name)
                         dst_group: h5py.Group = fout.create_group(name)
                         _copy_attributes(obj, dst_group)
-                        # Copy all contents of the transformations group
                         for key in obj:
-                            obj.copy(key, dst_group)
+                            _copy_child(obj, key, dst_group)
+                    elif nx_class == 'NXdisk_chopper':
+                        ensure_parent_groups(name)
+                        dst_group: h5py.Group = fout.create_group(name)
+                        _copy_attributes(obj, dst_group)
+                        # Static fields (slits, radius, delay, …) copy
+                        # verbatim; NXlog children (rotation_speed, phase)
+                        # become length-0 placeholders via ``_copy_child``.
+                        # Nested NXtransformations is skipped — the visitor
+                        # reaches its dedicated branch on the descent.
+                        for key, child in obj.items():
+                            if (
+                                isinstance(child, h5py.Group)
+                                and _nx_class(child) == 'NXtransformations'
+                            ):
+                                continue
+                            _copy_child(obj, key, dst_group)
 
         # Copy root attributes
         _copy_attributes(fin, fout)
--- a/src/ess/livedata/scripts/make_geometry_nexus.py
+++ b/src/ess/livedata/scripts/make_geometry_nexus.py
@@ -195,14 +195,25 @@ def write_minimal_geometry(
                         obj.copy('depends_on', dst_group)
                     elif nx_class == 'NXtransformations':
                         ensure_parent_groups(name)
-                        dst_group: h5py.Group = fout.create_group(name)
-                        _copy_attributes(obj, dst_group)
+                        # The group may already exist because a parent NXlog
+                        # placeholder copied a nested NXtransformations
+                        # recursively. Treat the branch as idempotent: copy
+                        # attributes and missing children only.
+                        if name in fout:
+                            dst_group: h5py.Group = fout[name]
+                        else:
+                            dst_group: h5py.Group = fout.create_group(name)
+                            _copy_attributes(obj, dst_group)
                         for key in obj:
-                            _copy_child(obj, key, dst_group)
+                            if key not in dst_group:
+                                _copy_child(obj, key, dst_group)
                     elif nx_class == 'NXdisk_chopper':
                         ensure_parent_groups(name)
-                        dst_group: h5py.Group = fout.create_group(name)
-                        _copy_attributes(obj, dst_group)
+                        if name in fout:
+                            dst_group: h5py.Group = fout[name]
+                        else:
+                            dst_group: h5py.Group = fout.create_group(name)
+                            _copy_attributes(obj, dst_group)
                         # Static fields (slits, radius, delay, …) copy
                         # verbatim; NXlog children (rotation_speed, phase)
                         # become length-0 placeholders via ``_copy_child``.
@@ -214,7 +225,8 @@ def write_minimal_geometry(
                                 and _nx_class(child) == 'NXtransformations'
                             ):
                                 continue
-                            _copy_child(obj, key, dst_group)
+                            if key not in dst_group:
+                                _copy_child(obj, key, dst_group)
 
         # Copy root attributes
         _copy_attributes(fin, fout)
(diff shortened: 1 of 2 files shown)


src/ess/livedata/scripts/make_geometry_nexus.py
```diff
@@ -45,6 +45,47 @@ def _copy_monitor_fields(src_group: h5py.Group, dst_group: h5py.Group) -> None:
...
+    geometry artifact wants only the schema (dtype, units, attributes) so
+    downstream consumers see the field shape without historical data.
+    Scalar children (e.g. ``average_value``) are copied verbatim.
+    """
+    _copy_attributes(src, dst)
+    for child_name, child in src.items():
+        if isinstance(child, h5py.Dataset) and child.ndim >= 1:
+            ds = dst.create_dataset(
+                child_name,
+                shape=(0, *child.shape[1:]),
```
nvaytet, line 69: Are we sure that the first index in the shape is always the one we want to trim?

src/ess/livedata/scripts/make_geometry_nexus.py
```diff
@@ -151,11 +195,38 @@ def visit_and_copy(name: str, obj: h5py.Group | h5py.Dataset) -> None:
...
-                            obj.copy(key, dst_group)
+                            if key not in dst_group:
+                                _copy_child(obj, key, dst_group)
+                    elif nx_class == 'NXdisk_chopper':
+                        ensure_parent_groups(name)
+                        if name in fout:
+                            dst_group: h5py.Group = fout[name]
+                        else:
+                            dst_group: h5py.Group = fout.create_group(name)
+                            _copy_attributes(obj, dst_group)
```
nvaytet: This code is repeated above (and maybe other places?
Is it worth making a helper function? Would it be used in a new `if` branch in the future for other fields?

src/ess/livedata/scripts/make_geometry_nexus.py
```diff
@@ -151,11 +195,38 @@ def visit_and_copy(name: str, obj: h5py.Group | h5py.Dataset) -> None:
...
+                        # Static fields (slits, radius, delay, …) copy
+                        # verbatim; NXlog children (rotation_speed, phase)
+                        # become length-0 placeholders via ``_copy_child``.
+                        # Nested NXtransformations is skipped — the visitor
+                        # reaches its dedicated branch on the descent.
+                        for key, child in obj.items():
+                            if (
+                                isinstance(child, h5py.Group)
+                                and _nx_class(child) == 'NXtransformations'
```
nvaytet: Why is the NXtransformation the only special case? (maybe it's fine, I'd just like to understand)

tests/make_geometry_nexus_test.py
```diff
@@ -131,3 +131,289 @@ def test_copies_nxlog_children(nexus_with_nxlog_chain, tmp_path):
...
+            data='/entry/instrument/chopper1/transformations/location',
+        )
+        # Static geometry fields
+        slits = chop.create_dataset('slits', data=2)
+        chop.create_dataset(
+            'slit_edges', data=np.array([0.0, 90.0, 180.0, 270.0])
+        ).attrs['units'] = 'deg'
+        chop.create_dataset('slit_height', data=0.1).attrs['units'] = 'm'
+        chop.create_dataset('radius', data=0.35).attrs['units'] = 'm'
+        chop.create_dataset('delay', data=0.0).attrs['units'] = 's'
```
nvaytet: Why is the delay a static field? Isn't it basically the phase of the chopper? I expect users to play with the delay as livedata is running.

tests/make_geometry_nexus_test.py
```diff
@@ -131,3 +131,289 @@ def test_copies_nxlog_children(nexus_with_nxlog_chain, tmp_path):
...
+    write_minimal_geometry(src, output)
+
+    with h5py.File(output, 'r') as f:
+        # Outer NXtransformations preserved.
+        assert f['entry/instrument/stage/transformations'].attrs['NX_class'] == (
+            'NXtransformations'
+        )
+        # NXlog trimmed.
+        rx = f['entry/instrument/stage/transformations/rx']
+        assert rx.attrs['NX_class'] == 'NXlog'
```
nvaytet: Not sure what this is testing exactly?
The test below ensure the log is trimmed. Here we just check the NX_class. As far as I can see, `rx` is not used further down in the test?

tests/make_geometry_nexus_test.py
```diff
@@ -131,3 +131,289 @@ def test_copies_nxlog_children(nexus_with_nxlog_chain, tmp_path):
...
+    write_minimal_geometry(nexus_with_disk_chopper, output)
+
+    with h5py.File(output, 'r') as f:
+        tr = f['entry/instrument/chopper1/transformations/location']
+        assert tr[()] == pytest.approx(15.0)
+        assert tr.attrs['transformation_type'] == 'translation'
+
+
+@pytest.fixture
+def nexus_with_populated_chopper_logs(tmp_path):
```
nvaytet: Why is this written as a fixture, and then only used in a single test below?
Other tests create the nexus inside the test body.

Can we use this fixture for multiple tests? Alternatively, make it consistent with the other tests and move this inside the test body?
It makes it hard to follow the tests that this fixture is here, and the test that uses it is quite a bit further down.

tests/make_geometry_nexus_test.py
```diff
@@ -131,3 +131,289 @@ def test_copies_nxlog_children(nexus_with_nxlog_chain, tmp_path):
...
+        log = f['entry/instrument/detector_0/transformations/rotation']
+        assert log.attrs['NX_class'] == 'NXlog'
+        assert log.attrs['transformation_type'] == 'rotation'
+        assert log['time'].shape == (0,)
+        assert log['value'].shape == (0,)
+        assert log['value'].attrs['units'] == 'deg'
+
+
+@pytest.fixture
+def nexus_with_populated_motor_log(tmp_path):
```
nvaytet: Same here: fixture used only once?

src/ess/livedata/scripts/make_geometry_nexus.py
SimonHeybrock: If you think of an NXlog, it writes values one by one over time. So I would expect that `time` is the outer (first) index, right?

tests/make_geometry_nexus_test.py
SimonHeybrock: This test here is a about a special case of nesting encountered in ESTIA (see docstring), which caused issues in the first iteration of this branch.

src/ess/livedata/scripts/make_geometry_nexus.py
SimonHeybrock: (Simon already replied above — NXlog convention has `time` as the outer index. No code change.)

src/ess/livedata/scripts/make_geometry_nexus.py
SimonHeybrock: Extracted `_get_or_create_dst` in 7e4a24fb — every NX_class branch now uses it.

src/ess/livedata/scripts/make_geometry_nexus.py
SimonHeybrock: Fair — it was a smell. In 7e4a24fb `_copy_child` recurses through non-NXlog groups, so any NXlog descendant gets trimmed wherever it sits. The skip is gone; the NXdisk_chopper handler reduces to the same two-line body as NXtransformations.

tests/make_geometry_nexus_test.py
SimonHeybrock: You're right — verified `delay` is an NXlog in production LOKI/ESTIA files. Switched the fixture to NXlog in 7e4a24fb.

tests/make_geometry_nexus_test.py
SimonHeybrock: The thin `rx` assertion next to it isn't pulling its weight — expanded in 7e4a24fb to also check trim, dtype, and units, which is what the test is really about.

tests/make_geometry_nexus_test.py
SimonHeybrock: Inlined in 7e4a24fb.

tests/make_geometry_nexus_test.py
SimonHeybrock: Inlined in 7e4a24fb.

tests/make_geometry_nexus_test.py
```diff
@@ -131,3 +131,281 @@ def test_copies_nxlog_children(nexus_with_nxlog_chain, tmp_path):
...
+    write_minimal_geometry(nexus_with_disk_chopper, output)
+
+    with h5py.File(output, 'r') as f:
+        tr = f['entry/instrument/chopper1/transformations/location']
+        assert tr[()] == pytest.approx(15.0)
+        assert tr.attrs['transformation_type'] == 'translation'
+
+
+def test_nested_nxtransformations_inside_nxlog_does_not_conflict(tmp_path):
+    """A real ESTIA structure: outer NXtransformations contains an NXlog
```
nvaytet: Why is this specific to ESTIA? It's also not very 'real'... seems more 'fake' than anything else?
```suggestion
    """Case where an outer NXtransformations contains an NXlog
```

src/ess/livedata/scripts/make_geometry_nexus.py
```diff
@@ -113,51 +179,24 @@ def write_minimal_geometry(
     """Create minimal geometry file with only detector positions and transformations."""
     with h5py.File(input_filename, 'r') as fin, h5py.File(output_filename, 'w') as fout:
 
-        def ensure_parent_groups(name: str) -> None:
-            parts: list[str] = name.split('/')
-            current_path: str = ''
-            for part in parts[:-1]:  # Skip the last part (current group name)
-                if not part:
-                    continue
```
nvaytet, line 121: I didn't entirely follow where this part went in the latest refactor...

src/ess/livedata/scripts/make_geometry_nexus.py
```diff
@@ -113,51 +179,24 @@ def write_minimal_geometry(
...
+            elif nx_class == 'NXmonitor':
+                dst = _get_or_create_dst(fin, fout, name, obj)
+                _copy_monitor_fields(obj, dst)
+            elif nx_class in ('NXsource', 'NXsample'):
+                dst = _get_or_create_dst(fin, fout, name, obj)
+                obj.copy('depends_on', dst)
+            elif nx_class in ('NXtransformations', 'NXdisk_chopper'):
+                dst = _get_or_create_dst(fin, fout, name, obj)
+                for key in obj:
+                    _copy_child(obj, key, dst)
```
nvaytet, line 207:
```suggestion
            dst = _get_or_create_dst(fin, fout, name, obj)
            if nx_class == 'NXdetector':
                _copy_detector_fields(obj, dst, use_pixel_shape=use_pixel_shape)
            elif nx_class == 'NXmonitor':
                _copy_monitor_fields(obj, dst)
            elif nx_class in ('NXsource', 'NXsample'):
                obj.copy('depends_on', dst)
            elif nx_class in ('NXtransformations', 'NXdisk_chopper'):
                for key in obj:
                    _copy_child(obj, key, dst)
```
or something like that?

tests/make_geometry_nexus_test.py
SimonHeybrock: Good point — adopted your suggestion in the next push.

src/ess/livedata/scripts/make_geometry_nexus.py
SimonHeybrock: Hoisted to module-level `_ensure_parent_groups` (line 125), and now called via the new `_get_or_create_dst` helper (line 138). Same logic — moved out of the closure so the other top-level helpers can share it.

src/ess/livedata/scripts/make_geometry_nexus.py
SimonHeybrock: Agreed — applied with one tweak: an explicit `_HANDLED_NX_CLASSES` guard before the `_get_or_create_dst` call. Without it the unconditional call would create empty output groups for every other NX_class in the input (NXentry, NXinstrument, NXuser, …); right now those are silently skipped.

tests/make_geometry_nexus_test.py
SimonHeybrock: ... but not sure why you call it "fake?

nvaytet: I think I meant in terms that it was not read from a real data file, the values were realistic but at the same time made up.
Anyway, it doesn't really matter anymore, as the comment no longer mentions Estia nor real.

nvaytet (approved):

PR #918, opened 2026-05-06
Add display titles for coord/dim axis labels with per-instrument overrides

### LINKED ISSUE #764: Coord/dimension titles?
We have "titles" most/many things, used for UI display. This also includes certain labels of plot axes. However, for many default plots we simply use the coord/dimension names, such as `time_of_arrival`. It might be useful (and consistent) to extend the our `title` (and `description`) mechanism also to dimensions and coords (probably the same for this purpose).

SimonHeybrock:
## Summary
Plot axis labels were locked to the raw coord/dim names, so the only way to display a dim differently (e.g. `detector_number` as "Pixel ID", or `wavelength` as λ) was to rewrite the data with per-view transforms. This adds a configurable display title for coord/dim names, mirroring the existing `source_metadata` pattern. A global `DEFAULT_DIM_TITLES` map (λ, 2θ, d-spacing, time of arrival, …) is merged with optional per-instrument `Instrument.dim_titles` overrides, threaded through `TitleResolver` to the HoloViews converters via a `dim_label` callback. LOKI now uses it to display `detector_number` as "Pixel ID".

Closes #764 - note that we are not adding a `description` for now as it did not seems very useful. Should be easy to extend later if we find a use.

## What this is *not*

This does *not* add a runtime user-facing config options for per-plot titles. Titles defined here are set in the package and are thus fixed (code change required to change title). Adding a plotter config option is a potential future enhancement.

## Test plan
- [x] Open a LOKI plot using the `detector_number` dim (e.g. `monitor_counts_per_pixel`) and verify the X-axis label reads "Pixel ID".
- [x] Open a plot with a `wavelength` axis and verify the label reads "λ".
- [x] Open a plot using a dim with no entry in `DEFAULT_DIM_TITLES` or per-instrument overrides and verify it still falls back to the raw dim name.


SimonHeybrock:
LGTM

PR #919, opened 2026-05-06
Drop ConfigKey envelope on livedata_commands topic

### LINKED ISSUE #853: Reconsider use of Kafka message key for ConfigUpdate on livedata_commands topic
## Context

`CommandSerializer` (`src/ess/livedata/kafka/sink_serializers.py`) and `CommandsAdapter` (`src/ess/livedata/kafka/message_adapter.py`) use a non-standard wire format for the `livedata_commands` topic:

- **Kafka message key** = encoded string form of `ConfigKey` (`source_name/service_name/key`)
- **Kafka message value** = payload JSON only (not the full `ConfigUpdate`)

The consumer reconstructs `ConfigUpdate.config_key` by parsing the Kafka key in `ConfigUpdate.from_raw` (`handlers/config_handler.py`).

This was originally tied to a compacted topic design. The topic is now created with `cleanup.policy=delete` (see `scripts/setup-kafka-topics.sh`), so the original motivation no longer applies. The stale comments were fixed in a follow-up commit, but the wire format itself was not changed.

## Question

Should we fold `config_key` into the value and drop the custom key handling?

### Pros of dropping the key

- Symmetric with every other stream in the system (value-only serialization).
- Removes a special case from both `CommandSerializer` and `CommandsAdapter`.
- `RawConfigItem` would no longer need a separate `key` field.
- Consumer no longer has to parse a structured identifier out of bytes.

### Open considerations

- **Ordering**: `livedata_commands` likely runs with a single partition (partition count is not set explicitly in `setup-kafka-topics.sh`), so per-key ordering via the Kafka key is not actually relied upon. If we ever wanted multi-partition, the key would restore per-`ConfigKey` ordering — but config updates are user-initiated and unlikely to race.
- **Wire-format break**: the topic has 2-day retention and holds no long-lived state, so a coordinated producer+consumer deploy should be safe.
- **Config keys in use**: in practice only `workflow_config` and `JobCommand.key` are dispatched (`ConfigProcessor._actions`). Worth confirming no external producers rely on the current key scheme.

## Proposed direction (to discuss)

Serialize the full `ConfigUpdate` as JSON in the value, drop the Kafka message key entirely, and simplify `RawConfigItem` / `CommandsAdapter` / `CommandSerializer` accordingly.

SimonHeybrock:
Closes #853.

## Motivation

The `livedata_commands` topic used a non-standard, asymmetric wire format: the Kafka message **key** carried an encoded `ConfigKey` (`source_name/service_name/key`), and the **value** carried only the payload JSON. Every other stream in the system uses value-only serialization. The original motivation was that the topic was log-compacted, so the Kafka key needed to be the deduplication key. The topic now uses `cleanup.policy=delete` so that motivation is gone.

But once we looked closely, the asymmetry was only the visible symptom. Most of `ConfigKey` had quietly become dead infrastructure:

- **`service_name` is fully unused.** Both registered specs in `keys.py` set it to `None`, the dispatcher never reads it, and no test or producer ever sets it to a non-None value. Service affinity moved into `WorkflowId.instrument` plus the registry-based `factory.get_service(workflow_id)` filter when `WorkflowId.namespace` was dropped in b6bb1e83.
- **`source_name` for `JobCommand` is redundant.** `JobCommand.job_id` already contains the source. The consumer-side adapter explicitly discards it (`_ = source_name  # Legacy, not used.`).
- **`source_name` for `WorkflowConfig` is the only meaningful use** — but only because `WorkflowConfig` itself didn't carry the source. It was smuggled via the Kafka key purely to satisfy `JobManager.schedule_job`, which then immediately reassembled `JobId(source_name, job_number)` from the envelope plus the payload's `job_number`.
- **The dedup wildcard branch** (`source_name=None overrides all previous source-specific updates`) has no producer in the current dashboard. Confirmed dead - this is a leftover from when the dashboard contained several widgets that controlled running workflows via config values. The mechanism was phased out last summer, moving to the pydantic-model based "config at workflow start" approach.
- **The runtime `SchemaRegistry` API** (`get_model`, `list_specs`, `get_produced_keys`) has zero callers outside the module itself. Only `WORKFLOW_CONFIG.create_key()` is used, and only in one call site that disappears here.

So the fix wasn't just "fold `config_key` into the value" — it was to recognize that `ConfigKey` no longer named anything meaningful, and that the cleanest end state is for each command type to carry its own routing information natively.

## Approach

Each command type now carries the identity it needs:

- **`JobCommand`** already had `job_id: JobId` — no change.
- **`WorkflowConfig`** gains `job_id: JobId` (required), replacing `job_number: JobNumber | None`. `JobId` packages `(source_name, job_number)` as a single field and is exactly what `JobManager.schedule_job` needs to construct anyway. The historical `None` / UUID-fallback was a leak from the old envelope design (where the producer couldn't fully construct a `JobId` because it owned `source_name` but not `job_number` until late) — it is gone now that producers always supply both halves.

The wire type becomes `Command = WorkflowConfig | JobCommand`, a Pydantic discriminated union on a `kind` literal field. The Kafka key is dropped entirely. With that, `ConfigKey`, `ConfigUpdate`, `RawConfigItem`, `keys.py`, and `schema_registry.py` all become unreferenced and are deleted in the same commit — keeping them as thin shims would only hide the simplification we are after.

The dispatch in `ConfigProcessor` switches from `actions[parsed_string_key](source_name, raw_value)` to a plain `match command` on the union variants, removing the parse-then-revalidate round trip on the consumer side. `JobManagerAdapter`'s methods take typed commands directly.

`JobId` was already embedded in Pydantic models elsewhere (`JobCommand.job_id`, `ServiceId.job_id`), so no conversion to `BaseModel` was needed — Pydantic v2 handles frozen dataclasses transparently. Verified with a round-trip test before touching anything.

## Decisions worth flagging

- **Discriminated union vs. `Any`-typed value.** Tightening `Command` to a closed union locks the wire format to two types. Adding a third command type already requires touching the dispatcher and adapter, so the union doesn't add coupling — it just documents what's actually on the wire.
- **`job_id` is required, not `JobId | None`.** The old `None` fallback existed only because `WorkflowConfig` couldn't express its own source. Once it can, "backend assigns" is no longer a meaningful state — the producer always knows both halves at construction time.
- **`SchemaRegistry` cleanup folded in, not deferred.** The whole module's only remaining caller (`keys.py`) goes away here. Keeping it as a separate PR would mean shipping orphaned code.
- **No backwards-compatibility shim for `ConfigKey`.** The wire-format break already forces a coordinated deploy; a stub class would only mislead future readers into thinking it still does something.

## Deployment

Wire-format break. Producer (dashboard) and consumer (backend services) must restart together. The topic has 2-day retention and no persistent state, so a coordinated deploy is sufficient — no migration window needed.

## Test plan

- [x] Full unit + slow test suite passes (3350 passed, 50 skipped).
- [x] End-to-end smoke test on a dev instrument: dashboard starts a workflow, backend processes data, dashboard issues stop/reset commands and observes ACKs.


src/ess/livedata/handlers/config_handler.py
```diff
@@ -2,134 +2,50 @@
...
-                    value=value,
-                )
-
-                if source_name is None:
-                    # source_name=None overrides all previous source-specific updates
-                    latest_updates[config_key].clear()
-
-                latest_updates[config_key][source_name] = update
-
+                if isinstance(command, WorkflowConfig):
```
jl-wynen, line 37: Nitpick: Since you used `kind` as the discriminator, does it make sense to always use that instead of using `isinstance`?

jl-wynen (commented):
I think this makes sense based on the description and the changes I see.

src/ess/livedata/handlers/config_handler.py
SimonHeybrock: Not that we are strict about typing here, but it would mean the `command` passed to the methods in the branches below would violate `mypy` and friends.

jl-wynen: True

jl-wynen (approved):

PR #921, opened 2026-05-07
Make stream statistics sticky in ServiceRegistry

SimonHeybrock:
## Summary

Stream statistics in the System Status tab's "Backend Workers" section took up to ~30 s to appear in a new session, even though the dashboard process is persistent and continuously consuming `ServiceStatus` messages.

Backend services publish `ServiceStatus` heartbeats every ~2 s but only attach a fresh `stream_stats` snapshot once per `_metrics_interval` (~30 s) — `_get_service_status` reads `_pending_stream_stats` and resets it to `None`. So roughly 14 of every 15 heartbeats arrive with `stream_stats=None`, and `ServiceRegistry.status_updated` overwrote the stored entry on each one. A session opening between drains read a `stream_stats=None` entry from the registry and showed "No stream data" until the next drain.

Fix: keep `stream_stats` sticky on the dashboard side. When an incoming status has `stream_stats=None` and the previously stored status has stats, preserve them. The per-row cache in `WorkerStatusRow` is now redundant and removed.

## Test plan

- [x] Unit tests for sticky preservation and fresh-snapshot overwrite in `service_registry_test.py`
- [x] Full dashboard test suite (1341 tests) passes
- [x] Open a fresh session against a running backend; "Backend Workers" stream stats render immediately

### DIFF
--- a/src/ess/livedata/dashboard/service_registry.py
+++ b/src/ess/livedata/dashboard/service_registry.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import time
+from dataclasses import replace
 
 import structlog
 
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
LGTM

Issue #780, opened 2026-03-11
[LOKI] beam_monitor_m4: NXlog transformation without value breaks monitor workflow


SimonHeybrock:
## Problem

The LOKI geometry file (from `coda_loki_999999_00021677.hdf`) contains `beam_monitor_m4` whose `depends_on` points to `transformations/trans_20`, which is an NXlog group with transformation attributes (`depends_on`, `transformation_type`, `vector`) but **no `value` dataset**.

ScippNexus reads this as a `DataGroup` instead of a transformation, causing:

```
AttributeError: 'DataGroup' object has no attribute 'depends_on'
```

The same `trans_20` path for `beam_monitor_4` has the NXlog structure with logged position values, suggesting this is a dynamic (positioner-driven) transformation that needs a position stream.

## Affected monitors

- `beam_monitor_m4` — empty NXlog, crashes immediately
- `beam_monitor_4` — has logged values, may also fail depending on how ScippNexus handles NXlog transformations at workflow init time

## Expected solution

The monitor workflow likely needs to accept a position stream for monitors with NXlog-based transformations, similar to how detector carriage positions are handled. The geometry extraction script (`make_geometry_nexus`) may also need to handle NXLog-valued transformations explicitly.

SimonHeybrock:
The structure this issue describes no longer exists, but the monitor is still broken — in a new shape. Checked against `coda_loki_999999_00020177.hdf`, a run of 2026-09-07. (Its run counter is *lower* than the `00021677` in the body, but that file predates this issue in March, so the counters do not order by time; going by the run timestamps this file is the newer description.)

**What changed.** `trans_20` is gone from the file entirely. `beam_monitor_m4/depends_on` now points at `transformations/x_translation`, a proper static dataset (`0.0 m`, `vector=[1,0,0]`), and the chain terminates cleanly at `'.'`:

```
x_translation -> beamstop_z1/transformations/motion   (NXlog, 0 samples)
              -> beamstop_y/transformations/motion    (NXlog, 0 samples)
              -> beamstop_x/transformations/motion    (NXlog, 0 samples)
              -> three static zeros in beamstop_x/transformations -> '.'
```

So one empty NXlog became three, and the monitor is on the **beamstop**, not the detector carriage as `specs.py` describes it.

[...]

PR #923, opened 2026-05-08
Centralise dynamic NeXus transforms for f144-driven geometry

### LINKED ISSUE #922: LOKI i_of_q workflow fails to start with new geometry artifact (length-0 NXlogs)
The trim-NXlogs-to-length-0 behaviour added to `make_geometry_nexus.py` (commit 45af6019) interacts badly with `_i_of_q_factory`: the `LokiWorkflow` graph is evaluated at construction time, hits the empty `detector_carriage` NXlog through the rear detector's depends_on chain, and `ess.reduce.nexus.workflow.reject_time_dependent_transform` raises because no filter is installed. This blocks the workflow from starting — not just in tests; production starts the same way.

This was masked while the registry pointed at `geometry-loki-2026-04-13.nxs` (predates the trim change, has `detector_carriage/value/value` shape `(1,)`). With `geometry-loki-2026-05-08.nxs` (chopper-aware artifact, shape `(0,)`), `tests/config/registered_workflow_factories_test.py::test_workflow_roundtrip[loki/i_of_q/1]` fails. Default `pytest` filters out the slow marker so the failure only surfaces under `tox` / CI.

### Repro
```sh
python -m pytest tests/config/registered_workflow_factories_test.py::test_workflow_roundtrip -k 'loki and i_of_q' -m ''
```

Stack ends in `ess/reduce/nexus/workflow.py:298 reject_time_dependent_transform`.

### Fix sketch
`DetectorViewFactory` already handles this correctly for the rear bank XY projection (`detector_view/factory.py:262`):

```python
value_stream = self._dynamic_transforms.get(source_name)
if value_stream is not None:
    add_dynamic_transform(workflow, transform_name=value_stream.transform_name)
    context_keys[value_stream.aux_stream] = TransformValueLog
```

`_i_of_q_factory` (`config/instruments/loki/factories.py`) needs the equivalent wiring: install `add_dynamic_transform` for `/entry/instrument/detector_carriage/value` on `loki_detector_0` and add the `detector_carriage` aux stream as a context key. This aligns with the `dynamic-detector-geometry-loki-carriage.md` plan; xy_projection just got there first.

### Notes
- Don't revert to the old artifact — the new one is required for the wavelength-LUT chopper workflow.
- Alternative: have `make_geometry_nexus.py` preserve a single sample per NXlog (as a "current state" snapshot) instead of trimming to 0. Less clean — the artifact is meant to be schema-only — but cheaper than touching every workflow that walks depends_on through f144 nodes.

SimonHeybrock:
## Why

Fixes #922. LOKI's `i_of_q` workflow crashes at construction against the chopper-aware geometry artifact: the rear-bank `depends_on` chain walks through `/entry/instrument/detector_carriage/value`, an empty NXlog placeholder, and essreduce's `reject_time_dependent_transform` filter fires because no live-stream override is installed.

The narrow fix would have copied `DetectorViewFactory`'s existing per-source `dynamic_transforms` wiring into `_i_of_q_factory`. We chose not to. The same pattern was already silently missing from `beam_monitor_m4` (the unfinished tail of #780/#782) and would need to be re-applied to every future SANS or monitor variant whose chain walks the same node. The mechanism's failure mode — "if you forget the wiring, the workflow crashes only when CI runs against the new artifact" — does not scale. This PR replaces it with a per-instrument registry and a single helper, so the project's "no silent wrong values" property covers dynamic geometry uniformly.

The fix is artifact-version-independent: the patched chain provider depends on a context key, so Sciline keeps it out of `compute(last_static)` at construction time. Neither the old length-1 NXlog nor the new length-0 NXlog crashes there. Once the chopper-workflow branch lands, the existing roundtrip test exercises the formerly-failing path automatically.

## Design choices

**Bindings keyed by NXlog path, not by source name.** One NXlog can sit on the `depends_on` chain of many components — the carriage node drives multiple positions in LOKI today, and more elsewhere tomorrow. Path-keying with an explicit consumer set mirrors the `depends_on` graph and avoids each component re-declaring the same routing.

**One author-declared `TransformLog` subclass per binding.** The original `NewType('TransformValueLog', sc.DataArray | None)` was a typing-spec misuse (pylint/mypy grumble) and lost the per-binding distinction Sciline needs when multiple dynamic transforms apply in one workflow. Short subclasses of a shared dataclass are grep-able, IDE-visible, type-safe, and give Sciline distinct keys.

**Apply patches per component type with a fused single-step provider.** Sciline replaces providers by return type, so a patched provider for `NeXusTransformationChain[T, SampleRun]` cleanly displaces essreduce's generic implementation for that instantiation; multiple component types compose. Fusing the prior two-step `TransformValueLog → TransformValue → chain` into one provider drops a parallel hierarchy whose intermediate had a single consumer per binding; splitting back is a binding-local refactor if ever needed.

**Aux-source merging centralised in `Instrument.register_spec`.** The old design threaded `dynamic_transforms=…` through every factory and required each callsite to remember it. Centralising the merge means all specs benefit, including those that call `register_spec` directly (e.g. `_i_of_q_factory`). Scoping by the binding's consumer set keeps unrelated specs free of phantom aux inputs.

**Registry sits on `Instrument`; helper is a method.** Binding-matching is pure data on the instrument; only the `workflow.insert(...)` step is Sciline-aware. Exposing it as `Instrument.apply_dynamic_transforms` keeps the registry encapsulated and matches the spirit of `register_spec`. Callers pass an explicit `source_name → component_type` mapping, which doubles as documentation of which loaded components are subject to dynamic geometry.

**Trust CI, not artifact introspection at runtime.** A registry-validator test (per instrument) walks each binding's `nxlog_path` from every declared consumer in the actual artifact and scans for orphan empty NXlogs. With that in place, the runtime helper does pure registry matching — no file I/O, no h5py walker on the construction hot path. The scippnexus-based chain walker lives only in the test.

**`Instrument` becomes a required argument of `DetectorViewFactory` and `create_monitor_workflow`.** Every production callsite already had an `Instrument` in scope; optionality only allowed misuse to compile. With the argument required, `InstrumentDetectorSource` — which just wrapped the instrument to pass it through `data_source` — became redundant and is removed; `data_source` is optional and defaults to instrument-derived geometry. Also tightens the wavelength-mode guard so the instrument-derived path is rejected explicitly rather than failing deep in essreduce. This is a small detour from the issue's scope, but the asymmetry would otherwise have remained as a confusing artifact of the rollout order.

**Loud failures everywhere.** Forgetting to call the helper: `reject_time_dependent_transform` raises at construction. Helper called but routing not wired: `UnsatisfiedRequirement` at construction. Both wired but no f144 message yet: "no samples yet" at first `finalize`. Registry inconsistent with artifact: CI test fails. Empty NXlog on a consumer's chain that no binding covers: CI test fails. No path silently substitutes a default.

## Deferred

- **`beam_monitor_m4` wavelength.** Its position is represented by a separate empty NXlog (`trans_20`) rather than the carriage node. Resolving it needs either a `make_geometry_nexus.py` change so m4 shares the carriage chain (a one-line addition to the binding's consumers) or registering an independent m4-position f144 stream. Tracked in the CI validator's known-orphan list; the test trips the moment the entry is removed.
- **Readiness gate at job start.** The closure already raises a clear "no samples yet" error at first `finalize`. Converting that into an actionable error at job start is orthogonal to the correctness fix and worth its own work.

## Test plan

- [ ] CI passes against `main`.
- [ ] After the chopper-workflow branch lands and registers the new geometry artifact, confirm `test_workflow_roundtrip[loki/i_of_q/1]` exercises the formerly-failing path under `tox`.
- [ ] Run LOKI `i_of_q` in dev mode against the new artifact and confirm the rear-bank workflow produces I(Q) once a `detector_carriage` f144 message arrives.
- [ ] Drop the `detector_carriage` f144 producer mid-run and confirm the workflow surfaces the "no samples yet" error rather than producing stale results.


SimonHeybrock:
Superseded

PR #925, opened 2026-05-12
fix: set estia monitor stream name

jokasimr:
Fixes the monitor naming to match the name ECDC gave the monitor.

It's also a bit nicer to only show the single monitor in the UI instead of the 9 generic ones.

### DIFF
--- a/src/ess/livedata/config/instruments/estia/specs.py
+++ b/src/ess/livedata/config/instruments/estia/specs.py
@@ -198,7 +198,7 @@ f144_log_streams = {
 instrument = Instrument(
     name='estia',
     detector_names=detector_names,
-    monitors=['cbm'],
+    monitors=['cbm1'],
     f144_attribute_registry={
         name: {'units': info['units']} for name, info in f144_log_streams.items()
     },
--- a/src/ess/livedata/config/instruments/estia/specs.py
+++ b/src/ess/livedata/config/instruments/estia/specs.py
@@ -16,7 +16,6 @@ from ess.livedata.handlers.monitor_workflow_specs import (
     register_monitor_workflow_specs,
 )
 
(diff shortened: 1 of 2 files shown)


src/ess/livedata/config/instruments/estia/specs.py
```diff
@@ -199,7 +198,7 @@ class EstiaReflectometryReductionOutputs(WorkflowOutputsBase):
 instrument = Instrument(
     name='estia',
     detector_names=detector_names,
-    monitors=list(GENERIC_CBM_MONITORS),
+    monitors=['cbm'],
```
SimonHeybrock: Consider also given this a user-facing title (and description) via `SourceMetadata` (see below and other instruments for example).

SimonHeybrock:
```suggestion
    monitors=['cbm1'],
```

src/ess/livedata/config/instruments/estia/streams.py
```diff
@@ -38,12 +40,17 @@ def _make_estia_logs() -> StreamLUT:
...
-            monitor_names=instrument.monitors,
-            cbm_start=0,
-        ),
+        monitors={
+            InputStreamKey(
+                topic=stream_kind_to_topic(
+                    instrument='estia', kind=StreamKind.MONITOR_EVENTS
+                ),
+                source_name='cbm',
+            ): 'cbm'
```
SimonHeybrock:
```suggestion
                source_name='cbm1',
            ): 'cbm1'
```

SimonHeybrock (commented):
cbm->cbm1 as now renamed by upstream

jokasimr:
> cbm->cbm1 as now renamed by upstream

That means this can be closed right?

Issue #933, opened 2026-05-13
Histogram (window mode) does not refresh when there are no events in the monitor


jokasimr:
If a monitor histogram is displayed in "window" mode with a 1s interval and the intensity the monitor sees goes down so that the monitor sees zero events, it will emit zero messages to kafka.
Then the histogram will remain whatever it was before the source went down, because it seems to only update when it actually receives messages.

Users will expect to see whatever the intensity was the last window interval period, so they would expect to see zero intensity.

If we want to depend on the arrival of messages to update figures we need to verify that we will always receive messages, even if the devices don't see any intensities etc.

SimonHeybrock:
This is by design and I am not sure we can do much about it, other than maybe improving the indicators that the data is stale. Upstream does not distinguish between "there were no events" and "message got delayed" - if there are not events than no `ev44` message is sent, but ESSlivedata cannot tell what that means.

jokasimr:
Not sure what you mean.

Here's how I understand it:

In normal circumstances (when some messages are arriving) esslivedata assumes that the messages that arrived in the interval were the full set of messages that should have arrived, that is, it assumes no messages were significantly delayed, and it displays the intensity from those messages.

But when no messages are arriving, esslivedata changes the assumption, and assumes that no messages arrived because they might be late. This seems to me a discrepancy we could address. But I might be missing something?

SimonHeybrock:
When no message arrives the backend does not call the Sciline workflow, and therefore it does not produce and all-zero output. So your options are one of:
1. Backend somehow synthesizes empty messages if the stream is silent -> workflow -> empty result (or unchanged cumulative)
2. Dashboard flags visibly that the stream is stale.

jokasimr:
I think if there are no events in the monitor / detector then our figures in esslivedata should reflect that by showing zero intensity. If the only way of achieving that is option (1) above then I guess that is what we should do, but maybe there's a better way? I'm not familiar enough with esslivedata to tell.

I got a bit more information from ECDC about this:

1. They say that for **histogram** data we should expect to see one message per pulse regardless of if there was any intensity in the monitor or not (note that this seems to currently not be the case, might be a bug). If we don't see any message we can assume something is wrong and inform the user about that. That makes the current mechanism here robust. If we don't see any messages we can mark the stream as stale.

[...]

PR #935, opened 2026-05-13
Unify cumulative/current outputs behind user-facing views

### LINKED ISSUE #929: Window aggregation interval selector allows selecting windows <1s, despite the minimum window being 1s.
It would be more clear if we capped the minimum value of the window size selector widget to 1s.

### LINKED ISSUE #932: "Latest" in the monitor workflow does not correspond to the latest monitor histogram in kafka
"Latest" mode in the monitor time-of-arrival histogram figures corresponds to a sum over the last 1s worth of data.

In most cases users probably expected to see the latest monitor histogram message, whatever period that covers.

Explicit time period coverage can be set by using the `window` mode instead of `latest`.

SimonHeybrock:
## Summary

Closes #932.

- Replace the `(cumulative)` / `(current)` output split with one user-facing view per quantity.
- Move the cumulative-vs-update choice into the window-mode plot parameter as a new `since_start` mode; the mode now drives which backend stream the plot subscribes to.
- Drop the parenthetical qualifier from plot titles and axis labels; window state is shown next to the title.
- Closes #929 by changing the meaning of "window" at setting to lower bound to 0.

## Design

- `OutputView` (in `workflow_spec.py`) binds one user-facing title to 1..N pydantic field names, keyed by `StreamRole` (`since_start`, `per_update`). Outputs classes declare views via an `output_views` ClassVar; classes that don't declare any get one view per field via fallback. Reduction outputs (LOKI, Estia, Bifrost, Dummy) need no annotation.
- `WindowMode.since_start` added. Mode availability per view: `since_start` requires the corresponding stream role; `window` requires `per_update`. For cumulative-only views the window widget collapses.
- `DataSourceConfig.output_name` → `view_name`. `PlotOrchestrator` resolves `(view_name, window_mode) → backend_field` at subscribe time. The wire-format `ResultKey.output_name` and all pydantic field names are unchanged.
- Plot title resolver maps backend field names back to view titles for legend / Y-axis labels.
- Window duration=0 means "latest"
- Added extended description to plot config

### Before

workflow output list
output selection for plot

### After

workflow output list
output selection for plot
window mode selection with extended description

## Migration

- The 5 checked-in grid templates have been migrated.
- User-saved layouts will be partially broken, plots need to be recreated.

## Test plan

- [x] `pytest tests/` (3324 passed locally)
- [x] Smoke-test the dashboard against a running backend, verifying:
  - [x] Each output appears once in the wizard with no `(cumulative)` / `(current)` qualifier
  - [x] Switching window mode between `since_start` / `latest` / `window` produces the correct data
  - [x] Cumulative-only outputs (e.g. `I(Q)`) show no window widget
  - [x] Migrated grid templates load and render correctly


jokasimr:
What does the "latest" mode do? Is it different from a window of 1 second? In what context would a user use it?

Edit: I guess this explains what the `latest` mode does:

https://github.com/scipp/esslivedata/pull/935/changes#diff-5f08cc55c4ff23709ab6621a8f193f55c15e68b832ae70a95e5fb11784c8056fR203-R209

However I don't think that is the best interface.

Can't we:

1. Remove `latest`.
2. Limit the window "length" to be at least the update period (typically 1 sec?).

Then the users get an interface with fewer options (no "latest"), and we remove the possibility of selecting a window length that is shorter than the update period, and viewing the "latest" update becomes just selecting the shortest available window length.

jokasimr:
I think the new interface looks nice! A bit cleaner to not have two versions of each view.

src/ess/livedata/config/workflow_spec.py
```diff
@@ -529,15 +532,38 @@ def _is_timeseries_output(da: sc.DataArray) -> bool:
...
-    Find all timeseries outputs in the workflow registry.
+    Find all timeseries output views in the workflow registry.
 
-    A timeseries output is a 0-D DataArray with a 'time' coordinate.
-    This is determined by checking the default_factory template of each
-    output field in the workflow spec.
+    A timeseries output is a 0-D DataArray with a 'time' coordinate. The
+    backing field for each ``per_update`` (or ``since_start``) stream of an
+    output view is inspected; views whose backing field templates match are
+    reported by view name.
```
jokasimr, line 566: What do time series have to do with this change?

I was under the impression that this only touched "accumulations".

jokasimr (commented):
Looks good to me. I have just a few questions about the interface and if it can be simplified a bit more.

SimonHeybrock:
> What does the "latest" mode do? Is it different from a window of 1 second? In what context would a user use it?
> 
> Edit: I guess this explains what the `latest` mode does:
> 
> https://github.com/scipp/esslivedata/pull/935/changes#diff-5f08cc55c4ff23709ab6621a8f193f55c15e68b832ae70a95e5fb11784c8056fR203-R209
> 
> However I don't think that is the best interface.
> 
> Can't we:
> 
>     1. Remove `latest`.
> 
>     2. Limit the window "length" to be at least the update period (typically 1 sec?).

The problem is the "typical":
1. There might not be raw data every second (at least I have seen no such guarantees)
2. The backend may dynamically degrade by increasing the batch length.
3. For workflows with large outputs we might in principle choose to not publish every second (there is no such case currently).

> Then the users get an interface with fewer options (no "latest"), and we remove the possibility of selecting a window length that is shorter than the update period, and viewing the "latest" update becomes just selecting the shortest available window length.

We don't strictly know what the update period is. I don't know if having the "1 second window" just mean "display the last value" is correct. Should "1 second window" (a) select all the result messages in a one second window and potentially result in zeros or (b) select at least the latest message? The two are equivalent if we define "current time = time of last message in this stream".

src/ess/livedata/config/workflow_spec.py
SimonHeybrock: Timeseries in this sense are series of messages that are collected in the frontend. They are made up of individual result messages. This can be done for any output that is not "cumulative". Since we moved the "cumulative" and "current value" output fields into views the detection logic is also changed. It is conceptually not different, stuff just got moved around since we need to look for the inputs inside the views..

SimonHeybrock:
> Should "1 second window" (a) select all the result messages in a one second window and potentially result in zeros or (b) select at least the latest message? The two are equivalent if we define "current time = time of last message in this stream".

@jokasimr This may also partially affects what we can do about #933.

jokasimr:
> There might not be raw data every second (at least I have seen no such guarantees)
> For workflows with large outputs we might in principle choose to not publish every second (there is no such case currently).

The 1 second that I mentioned was not meant to be a default value. I meant that *whatever* the minimum reasonable window length is, it would be good if that was also the minimum value in the widget, because otherwise it looks to the users like they can select shorter window times than the workflow allows, and that will be confusing.

> The backend may dynamically degrade by increasing the batch length.

How does that interact with windowing? If the batch length suddenly goes over the selected window length, will the figures start showing wrong results?

> I don't know if having the "1 second window" just mean "display the last value" is correct. Should "1 second window" (a) select all the result messages in a one second window and potentially result in zeros or (b) select at least the latest message? The two are equivalent if we define "current time = time of last message in this stream".

I didn't mean we should redefine the meaning of "1 second window". In my opinion the answer to the question is (a), if the user selected 1 second window we should always show the data from the last 1 second window.

What I meant was that if the batch length happens to be equal to the selected window length, then they are the same, right?

SimonHeybrock:
> > There might not be raw data every second (at least I have seen no such guarantees)
> > For workflows with large outputs we might in principle choose to not publish every second (there is no such case currently).
> 
> The 1 second that I mentioned was not meant to be a default value. I meant that _whatever_ the minimum reasonable window length is, it would be good if that was also the minimum value in the widget, because otherwise it looks to the users like they can select shorter window times than the workflow allows, and that will be confusing.

That would be difficult to implement. It can change on the fly, what would you expect existing plots do to, that have a smaller window length?


> > The backend may dynamically degrade by increasing the batch length.
> 
> How does that interact with windowing? If the batch length suddenly goes over the selected window length, will the figures start showing wrong results?

Maybe not wrong, but unexpected.

> > I don't know if having the "1 second window" just mean "display the last value" is correct. Should "1 second window" (a) select all the result messages in a one second window and potentially result in zeros or (b) select at least the latest message? The two are equivalent if we define "current time = time of last message in this stream".
> 
> I didn't mean we should redefine the meaning of "1 second window". In my opinion the answer to the question is (a), if the user selected 1 second window we should always show the data from the last 1 second window.

Right, and I am saying that that is not possible.

> What I meant was that if the batch length happens to be equal to the selected window length, then they are the same, right?

Correct, if we move the window only when messages arrive, i.e., not tied to wallclock or time in messages from other streams.

jokasimr:
> What does the "latest" mode do? **In what context would a user need it?**

I still don't understand what the answer to the highlighted question is.

> >   I didn't mean we should redefine the meaning of "1 second window". In my opinion the answer to the question is (a), if the user selected 1 second window we should always show the data from the last 1 second window.

> Right, and I am saying that that is not possible.

What is impossible? I think there's a misunderstanding here. What should the user expect to see in the figure if they select a 1 second window time?

SimonHeybrock:
> What is impossible? I think there's a misunderstanding here. What should the user expect to see in the figure if they select a 1 second window time?

It is impossible to show "data from a 1 second time window". What we can do is "show result messages from a 1 second time window", but each result message has an a-priori unknown internal "window length" (which *typically* is 1 second, but there is no guarantee). If we are fine with that, we could do the following:

#### Proposal: drop `latest`, unify into `window` with `ge=0`, default `0`

**Modes**: `since_start`, `window`. (No more `latest`.)

**Mental model for `window`**: *"the latest update, plus up to N seconds of prior history."* The latest update is the implicit floor; the duration knob controls additional lookback only.

Under this framing:

- `window = 0s` → just the latest update. Coherent: 0s of *extra* history.
- `window = 5s` → latest update plus updates from the prior 5s.
- If the requested lookback is shorter than the backend cadence, the extra-history window is empty and you simply get the latest update. Not a contradiction — you asked for N seconds of additional history, there wasn't any.

**Defaults / bounds**: `ge=0`, `default=0`. This preserves the current default behaviour (today's default is `latest`) and keeps `LatestValueExtractor` + `SingleValueBuffer` available for the duration=0 case, which matters for ND detector arrays.

**Internal mapping**: `duration == 0` constructs `LatestValueExtractor`; `duration > 0` constructs `WindowAggregatingExtractor`. The UI sees one mode; the buffer/extractor choice is an implementation detail.

**Header label**: stop reading `window_duration_seconds`. Display the actual time range from the data's `start_time`/`end_time` coords. The configured duration is a request; the header reports reality. This also resolves the existing inconsistency where `_format_window_info` can disagree with `_compute_time_info`.

**On a 1s minimum (re: #929)**: Backend cadence isn't a UI constant — encoding it in the dashboard parameter bound couples the dashboard to deployment-specific timing, and forbidding `0` removes the cheap path for ND arrays. If the concern is users picking sub-cadence values: the honest header makes the consequence visible, and `default=0` plus a tooltip ("latest update + N seconds of prior history; 0 = latest only") should be sufficient guidance.

jokasimr:
I think the trade-off to think about here is: Do we care mostly about the "freshness" of the data in the figure, or do we care mostly about the accuracy of the time window that the data in the figure covers?

Right now the system degrades the accuracy of the time window to keep the figure update frequency (freshness) high. I think the better alternative would be to keep the accuracy of the time window constant, and instead update the figures less frequently. But I imagine changing that would be an invasive change, and it might not be worth the effort.

The suggestion in your previous comment seems reasonable to me. It's a bit unusual, but it's coherent, and more explicit than the current interface. :+1:

SimonHeybrock:
> I think the trade-off to think about here is: Do we care mostly about the "freshness" of the data in the figure, or do we care mostly about the accuracy of the time window that the data in the figure covers?
> 
> Right now the system degrades the accuracy of the time window to keep the figure update frequency high. I think the better alternative would be to keep the accuracy of the time window constant, and instead update the figures less frequently.

I have my doubts if scientist would agree. No one wants a, say, 2-second or 4-second window just because it is our conservative estimate of how long the backend might take in the "worst" case.

> But I imagine changing that would be an invasive change, and it might not be worth the effort.
> 
> The suggestion in your previous comment seems reasonable to me. It's a bit unusual, but it's coherent, and more explicit than the current interface.

I'll update the PR description and screenshots.

jokasimr:
> I have my doubts if scientist would agree. No one wants a, say, 2-second or 4-second window just because it is our conservative estimate of how long the backend might take in the "worst" case.

And you may well be right about that. It probably also depends on the kind of visualization. For quantities that are normalized by the size of the time window, the exact width of that window probably matters less. But for figures showing a "raw" intensity it might matter more.

SimonHeybrock:
> > I have my doubts if scientist would agree. No one wants a, say, 2-second or 4-second window just because it is our conservative estimate of how long the backend might take in the "worst" case.
> 
> And you may well be right about that. It probably also depends on the kind of visualization. For quantities that are normalized by the size of the time window, the exact width of that window probably matters less. But for figures showing a "raw" intensity it might matter more.

The dashboard has the option of normalizing (counts outputs) by time, so the exact window does not matter, even if the workflow does not perform such normalization.

jokasimr:
> The dashboard has the option of normalizing (counts outputs) by time, so the exact window does not matter, even if the workflow does not perform such normalization.

That's good to know!

src/ess/livedata/dashboard/plot_params.py
```diff
@@ -289,12 +289,28 @@ class RateNormalizationParams(pydantic.BaseModel):
...
+    "This setting only controls how many of those results the dashboard "
+    "combines for display; it does not change what or how often the "
+    "reduction emits."
+    "<br><br>"
+    "The duration is a target, not a guarantee: when it does not align "
+    "with the reduction's current cadence the aggregated time range "
+    "will differ from the requested value, and a duration shorter than "
+    "one cadence interval collapses to just the latest result. To "
+    "compare values across different cadences or window lengths, enable "
+    "'Counts Per Second' in the 'Rate' tab."
```
jokasimr: I find this is a bit fuzzy and hard to understand. 
Would it be possible to make it more concrete?

I have three suggestions:

1. Using the same word for the window parameter throughout the explanation. (Right now it's called "duration" and "length" and "value").
2. Saying explicitly what happens when the window duration is less than one cadence, instead of using the "collapsing" analogy.
3. Isn't "aggregation time range" more precise than "aggregated time range", because the interval we are talking about it the interval that is used for aggregation?

So, something like:
```
    "The window duration is a target, not a guarantee: when it does not align "
    "with the reduction's current cadence the aggregation time range "
    "1) will differ from the requested window duration, and 2) always covers at least one cadence interval. "
    "To compare values across different cadences or window durations, enable "
    "'Counts Per Second' in the 'Rate' tab."
```

That said, it's not a blocker, write it the way you think is most clear.

jokasimr (approved):

jokasimr:
(Removing `latest` remains to be done here, but I assume that is a rather mechanical change with the help of Claude.)

SimonHeybrock:
> (Removing `latest` remains to be done here, but I assume that is a rather mechanical change with the help of Claude.)

Can you clarify what remains to be done?

jokasimr:
> Can you clarify what remains to be done?

As I understood https://github.com/scipp/esslivedata/pull/935#issuecomment-4495132016 we don't need a `latest` mode in the window settings widget any more, because that case is covered by window + duration=0. But in the code it looks like there are still references to a `latest` mode. But maybe those are only internal, and the user interface does not have `latest` any more?

If the `latest` mode is still part of the UI then removing it remains to be done (unless we have to keep it for some reason?). Otherwise there are no blockers from my side.

Edit: I can see now (https://github.com/scipp/esslivedata/pull/935/changes#diff-5f08cc55c4ff23709ab6621a8f193f55c15e68b832ae70a95e5fb11784c8056fR13-R22) that the `latest` mode is removed. :+1:

SimonHeybrock:
> > Can you clarify what remains to be done?
> 
> As I understood [#935 (comment)](https://github.com/scipp/esslivedata/pull/935#issuecomment-4495132016) we don't need a `latest` mode in the window settings widget any more, because that case is covered by window + duration=0. But in the code it looks like there are still references to a `latest` mode. But maybe those are only internal, and the user interface does not have `latest` any more?
> 
> If the `latest` mode is still part of the UI then removing it remains to be done (unless we have to keep it for some reason?). Otherwise there are no blockers from my side.

Can you point me to the instances you mean? There is still a `LatestValueExtractor` as that is needed for efficiency, but `latest` should be gone from the user facing bits.

jokasimr:
> Can you point me to the instances you mean? There is still a LatestValueExtractor as that is needed for efficiency, but latest should be gone from the user facing bits.

I didn't look through it well enough. After looking again it looks to me `latest` is gone from the user facing bits.

PR #936, opened 2026-05-13
Unified per-instrument stream registry + LogContextBinding

SimonHeybrock:
Each instrument used to declare f144 streams in three parallel places — a per-instrument `f144_log_streams` dict, an `Instrument.f144_attribute_registry`, and the `StreamLUT[logs]` topic/source map — kept in sync by hand. Wiring a stream into a Sciline workflow added a fourth declaration in the consuming factory. The shape did not survive adoption pressure from two upcoming efforts: dynamic detector transforms (LOKI carriage) and the chopper workflow, both of which need stream → Sciline-key routing for O(10) and O(100) f144 streams respectively.

This PR collapses the parallel structures into one source of truth per instrument and adds a thin binding layer for the routing.

## What changed

- **`Stream` / `F144Stream` records** in `config.stream`. One frozen dataclass per streaming group carrying NeXus identity (`nexus_path`, `nx_class`, `parent_nx_class`), wire identity (`topic`, `source`), and writer-module-specific overlay (`units` for f144). Synthesised streams are records with `topic=None`.
- **`Instrument.streams: dict[str, Stream]`** replaces `f144_log_streams` and `f144_attribute_registry`. `StreamLUT[logs]` and timeseries auto-registration both derive from it; factories look up units via `Instrument.streams[name]`.
- **`LogContextBinding`** declares `(stream_name, workflow_key, dependent_sources)`. Factories call `instrument.get_context_keys(source_name)` to get the `StreamProcessorWorkflow.context_keys` map filtered by `dependent_sources`. Bifrost's cut workflow migrates as the demonstration consumer; LOKI dynamic-transforms and the chopper workflow can adopt the same mechanism in focused follow-ups.
- **Codegen for NeXus-rich instruments.** `python -m ess.livedata.nexus_helpers <coda.hdf> --generate` emits an importable `streams_parsed.py` containing a path-keyed `dict[str, F144Stream]`. Each instrument's `specs.py` composes the final name-keyed registry via `name_streams(PARSED_STREAMS, rename=...)`. Auto-naming derives instrument-facing names from the NeXus path tail (filtering generic containers like `entry`/`instrument`/`sample`); `rename=` overrides individual entries by `nexus_path`.
- **All instruments rolled out** (except `dummy`, which stays hand-coded): bifrost (72 entries), dream (72), estia (62), loki (140), nmx (64), odin (278), tbl (68). Sourced from per-instrument `coda_<inst>_*.hdf` ground-truth files.

Design rationale — why these shapes and not others — lives in `docs/developer/design/stream-registry.md`. This PR replaces the work-in-progress plan that previously sat under `docs/developer/plans/`.

## Why now

Without unification, each added consumer (LOKI's dynamic-transforms `log_context_bindings` shim, the chopper workflow's `dynamic_transforms` dict) bolts on its own parallel structure and the four-place declaration drifts. This PR collapses them so the follow-ups become focused diffs that just *read* the registry and append to one binding list.

The expansion of registered streams (a handful per instrument → ~756 auto-registered timeseries workflows across instruments) is intended: the manual lists were incomplete; the dashboard now surfaces the full readback set the file writer actually publishes.

## Caveats

- `streams_parsed.py` is checked-in code, regenerated by hand when the geometry file changes. There is no automated drift check; staleness surfaces only when someone regenerates and notices a diff.
- Estia geometry uses the same PV (`ESTIA-DtRot:MC-RotZ01:Mtr.RBV`) under two NeXus paths; only the readback path is exposed via the rename map.

## Test plan

- [x] Smoke-test each backend service with one instrument (`fake_logdata` + `timeseries` + dashboard) and confirm timeseries workflows appear and accumulate.
- [ ] Verify Bifrost cut workflow still binds rotation aux inputs correctly (regression for the binding migration).
- [x] Verify LOKI dynamic-transforms still wires `detector_carriage` to the depends_on chain.


src/ess/livedata/config/stream.py
```diff
@@ -0,0 +1,167 @@
...
+    ``nexus_path`` may still be None for hand-coded entries that have not been
+    cross-referenced with a NeXus geometry file.
+    """
+
+    writer_module: str
+    nexus_path: str | None = None
+    topic: str | None = None
+    source: str | None = None
+    nx_class: str = ''
+    parent_nx_class: str | None = None
```
jl-wynen: Why the different default values? Should `nx_class` default to `None`?

src/ess/livedata/config/instrument.py
```diff
@@ -107,11 +109,63 @@ def __post_init__(self) -> None:
...
+        """
+        binding = LogContextBinding(
+            stream_name=stream_name,
+            workflow_key=workflow_key,
+            dependent_sources=frozenset(dependent_sources),
+        )
+        self._validate_binding_stream_name(binding)
+        self.log_context_bindings.append(binding)
+
+    def get_context_keys(self, source_name: str) -> dict[str, Any]:
```
jl-wynen, line 171: Does this really return all context keys? What about user-specified workflow parameters? Or does the workflow need to be reconstructed when those change?

SimonHeybrock: "context keys" here are only workflow inputs that are bound to *streams*. It does not include fixed params that a hard-coded or set from the values in the pydantic model the uses configures via the UI.

src/ess/livedata/config/stream.py
SimonHeybrock: I removed the unused `parent_nx_class` field.

jl-wynen (approved):

PR #937, opened 2026-05-18
Cap default source pre-selection in config modals

SimonHeybrock:
## Summary

Restores a cap on default source pre-selection in workflow and plot configuration modals, so that opening a fresh timeseries workflow (or any modal with many sources and no stored configuration) no longer arrives with 50+ source rows pre-selected. The original cap was added in d8c748768 and silently dropped during the single-source `ConfigurationState` refactor in 8e0ae4e54. The narrower replacement in ab904013d only covered 2D+ plot outputs.

`ConfigurationAdapter.initial_source_names` now returns an empty list once the source count exceeds `DEFAULT_SOURCE_PRESELECTION_CAP` (5), but only when `config_state is None`. When stored configuration exists (editing an existing plot, or gear-clicking a configured workflow group) the prior selection is preserved as-is — the cap must never silently truncate existing data.

`ConfigurationWidget` gains a small Select all / Select none button row above the source `MultiChoice`, always shown when the selector exists, providing a one-click escape hatch both for fresh-with-cap and for trimming a restored selection.

## Example

workflow config modal with added select-all and select-none buttons

## Test plan

- [x] Open a fresh timeseries workflow on an instrument with many log sources (e.g. DREAM): the gear modal opens with no sources selected, Select all populates the full list.
- [x] Stage a few sources, commit, gear-click the configured group: the prior selection is preserved unchanged.
- [x] Open an existing 1D plot with many selected sources: prior selection is preserved.
- [x] Open a fresh 1D plot on a workflow with many sources: no sources pre-selected.
- [x] Verify Select all / Select none buttons align with the right edge of the source MultiChoice.

### DIFF
--- a/src/ess/livedata/dashboard/configuration_adapter.py
+++ b/src/ess/livedata/dashboard/configuration_adapter.py
@@ -11,6 +11,15 @@ from ess.livedata.config.workflow_spec import AuxSources
 
 Model = TypeVar('Model')
 
+DEFAULT_SOURCE_PRESELECTION_CAP = 5
+"""Maximum number of sources to pre-select by default.
+
+Above this threshold, ``initial_source_names`` defaults to an empty list so
+the user is not faced with a long pre-selection that they have to trim. The
+configuration widget exposes "Select all" / "Select none" buttons so the
+user can opt into the full set with a single click.
+"""
+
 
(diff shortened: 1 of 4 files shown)


SimonHeybrock:
LGTM

PR #938, opened 2026-05-18
Merge VAL/DMOV/RBV substreams into per-device streams

SimonHeybrock:
## Summary

ESS instruments publish three independent f144 substreams per motorised device (RBV readback, VAL setpoint, DMOV idle flag) with non-synchronised timestamps. Until now only RBV reached any consumer; VAL and DMOV arrived from Kafka but went nowhere, so no workflow could (e.g.) filter readings taken while the motor was still moving.

This PR adds in-process synthesis of merged per-device streams, exposing RBV as data with `target` and `settled` as `time`-dim coords on a single `sc.DataArray`. Any consumer (timeseries plotting, Sciline workflows) gets the merged view without re-implementing cross-stream alignment. The change does not by itself introduce any consumer that filters on `settled` or overlays `target` — it unblocks those developments.

Design rationale (anchoring, timestamp policy, source-suffix detection, bootstrap, etc.) lives in [ADR-0001](docs/developer/adr/0001-in-process-device-stream-synthesis.md). This PR also introduces `docs/developer/adr/` as the home for load-bearing design records, modelled on [scipp's convention](https://github.com/scipp/scipp/tree/main/docs/development/adr).

Wiring:
- All four motion-consuming services (`timeseries`, `data_reduction`, `monitor_data`, `detector_data`) wrap their adapted message source with `DeviceSynthesizer`. Instruments that configure no devices fall through transparently.
- Bifrost's two `LogContextBinding`s (`InstrumentAngle`, `SampleAngle`) and loki's `detector_carriage` dynamic transform migrate to the merged Device names.
- The dev `LogProducerWidget` is extended to drive synthesised devices (animated readback ramp, DMOV transitions) so the merged stream can be exercised without live EPICS.

Bundled fixes surfaced by this work:
- Dashboard `TemporalBuffer` no longer wipes history when any time-dim coord changes — previously the new `target`/`settled` coords triggered a reset.
- Two-pass naming in `suggest_names` lets device names drop the redundant parent prefix, removing a per-instrument `rename={...}` workaround.

## Test plan

- [x] Bifrost reduction workflow verified end-to-end against live data with the migrated bindings; the extra `target`/`settled` coords ride along passively through `sc.lookup`.
- [x] LogProducerWidget exercised against LOKI; merged timeseries plot shows continuous history.


docs/developer/adr/0001-in-process-device-stream-synthesis.md
```diff
@@ -0,0 +1,85 @@
...
+
+Goal: make each device available as one logical stream and one DataArray — RBV as data, target/settled attached as time-dim coords on a shared `time` axis — so that any consumer (timeseries plotting, Sciline-based science workflows) can opt into using the extra fields without re-implementing cross-stream alignment. This change does not by itself introduce any consumer that filters on `settled` or overlays `target`; it unblocks those developments.
+
+## Decision
+
+Merge per-device substreams **in-process via a `MessageSource` decorator** (`DeviceSynthesizer`), matching the precedent of `ChopperSynthesizer`. The synthesizer:
+
+1. Maintains per-device state (last-seen value + time per substream).
+2. Emits `LogData` messages (with optional `target` / `settled` fields populated per device configuration) on a new `StreamKind.DEVICE` once bootstrap is complete (every configured substream observed at least once).
```
jl-wynen: Why 'settled' instead of the 'idle' as in the files? Deviating here makes it harder and more confusing to use this for files.

src/ess/livedata/config/instrument.py
```diff
@@ -125,7 +125,14 @@ def __post_init__(self) -> None:
...
             self._validate_binding_stream_name(binding)
-        timeseries_names = list(self.f144_streams)
+        # Device substreams are suppressed by DeviceSynthesizer; expose the
+        # merged Device stream instead of the per-PV substream plots.
+        devices = self.devices
+        suppressed = {n for d in devices.values() for n in d.substream_names}
+        timeseries_names = [
+            name for name in self.f144_streams if name not in suppressed
+        ]
+        timeseries_names.extend(devices)
```
jl-wynen:
```suggestion
        timeseries_names = set(self.f144_streams) - suppressed | set(devices)
```

jl-wynen: Also, can you extract this into a separate function so it doesn't leak `devices` and `suppressed`?

src/ess/livedata/config/stream.py
```diff
@@ -55,6 +55,35 @@ class F144Stream(Stream):
...
+    """Synthesised stream merging RBV/VAL/DMOV substreams of a motorised device.
+
+    A ``Device`` is not transported over Kafka; it is materialised in-process
+    by :class:`DeviceSynthesizer` from the substreams referenced by
+    :attr:`value`, :attr:`target`, and :attr:`settled` (each is a key into
+    :attr:`Instrument.streams`).
+
+    ``value`` is required and points to the RBV substream. ``target`` and
+    ``settled`` are optional: only RBV is required for the synthesizer to
+    emit a sample.
```
jl-wynen: The ADR says

> Substreams co-located under one NeXus parent group are then bundled into a `Device` if classified RBV is present **and** at least one of classified VAL / DMOV is present

which seems to contradict what is written here.

Where are devices actually identified?

src/ess/livedata/handlers/to_nxlog.py
```diff
@@ -66,9 +78,16 @@ def _ensure_capacity(self, data: sc.Variable) -> None:
...
-            self._timeseries = sc.DataArray(
-                values, coords={'time': self._start + times}
-            )
+            coords: dict[str, sc.Variable] = {'time': self._start + times}
+            if self._has_target:
+                coords['target'] = sc.zeros(
+                    dims=['time'], shape=[2], unit=self._unit, dtype='float64'
+                )
+            if self._has_settled:
+                # int32 (not bool) because da00 serialization rejects bool.
```
jl-wynen: Is this a limitation in ESSlivedata or ECDC software?

jl-wynen (approved):

PR #941, opened 2026-05-22
Fix shipped grid templates broken since namespace drop (26.5.0)

SimonHeybrock:
The bifrost and dummy grid templates have shipped with obsolete 4-part `workflow_id` strings (`instrument/group/name/version`) since `42837b33` dropped the namespace from `WorkflowId` identity. `WorkflowId.from_string` raises on them, `_parse_single_spec` catches the `ValueError`, and every affected template silently disappears from the dashboard's Template dropdown. Shipped broken in 26.5.0.

Rewrites every `workflow_id:` line in shipped templates to the 3-part form, plus a regression test that parses every `workflow_id` in every shipped template.

Out of scope: `_parse_single_spec` swallowing template errors at startup is also worth revisiting -- fail-fast on a misconfigured template would have surfaced this in CI or first-boot.

## Test plan

- [x] Bifrost dashboard: open the Template dropdown, confirm Detectors, Monitors, Spectrum View, and Bragg Peak Monitor are present.
- [x] Apply each template, confirm cells subscribe and render without errors in the browser console / backend logs.
- [x] Dummy dashboard: open Template dropdown, confirm Detectors template is present and applies.

### DIFF
--- a/src/ess/livedata/config/instruments/bifrost/grid_templates/bragg_peak_monitor.yaml
+++ b/src/ess/livedata/config/instruments/bifrost/grid_templates/bragg_peak_monitor.yaml
@@ -12,7 +12,7 @@ cells:
   layers:
   - data_sources:
       primary:
-        workflow_id: bifrost/monitor_data/monitor_histogram/1
+        workflow_id: bifrost/monitor_histogram/1
         view_name: total_in_range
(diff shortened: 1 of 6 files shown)


SimonHeybrock:
LGTM

PR #942, opened 2026-05-26
Downsample and throttle timeseries plots

### LINKED ISSUE #940: Long-running timeseries plots cause dashboard lag (server-side too, not just rendering)
## Summary

Long-running timeseries plots (≳1 day at 1 Hz, ~100k points) make the dashboard noticeably laggy. Symptoms reported by users include perceived freezes on browser connect, some tabs initially appearing empty, and reloading the page sometimes "fixing" things. Crucially, the underlying problem develops even while **no browser is connected** — connecting just exposes it.

This issue is the result of an analysis of the current data path; it does not propose a fix.

## What is going on

On every Kafka delta (≈1 / s for f144 scalar logs), the timeseries plot rebuilds the **entire** `hv.Curve` element from the **entire** buffered history, caches it, and — for each connected session — pushes the full element through a HoloViews `Pipe` on the per-session tick. There is no streaming, no patching, and no downsampling at any layer. Per-tick cost is O(N) along several axes.

Data flow:

```
ToNXlog (service-side, doubling buffer)
  → TimeseriesStreamProcessor.finalize() emits only the new slice ("delta")
    → Kafka da00
      → Orchestrator.update() on the dashboard's background ingestion thread
        → DataService.__setitem__
          → TemporalBuffer.add(delta)                                  (append)
          → notify subscribers
            → FullHistoryExtractor.extract(full buffer)                (O(N))
              → on_data → _run_compute → plotter.compute()             (O(N))
                → builds hv.Curve from full history
                → stores as _cached_state
                → marks presenters dirty (no-op if no session)
```

With a browser attached, additionally:

```
SessionUpdater periodic_callback (1 s, per session)
  → SessionLayer.update_pipe()
    → pipe.send(full element)
      → DynamicMap callback → HoloViews → Bokeh ColumnDataSource patch
        → WebSocket flush                                              (≈2.4 MB/s/plot at N=100k)
```

## Three angles the investigation looked at

> Is plotting 100k points the problem, or plotting 100k times (1, 2, ..., N)?

They are the same code path. Each tick rebuilds the full element with the current buffer, so total work over a day is roughly Σ k·c ≈ N²·c/2 of background-thread CPU per plot (where c is the per-point cost). With several plots this stacks linearly.

> Why does the dashboard get laggy even with no browser connected?

`_run_compute` runs on the background ingestion thread the moment a Kafka message arrives, independently of session state:

- `Orchestrator.update()` is driven by `dashboard_services._update_loop` every ~200 ms.
- Inside the `DataService.transaction`, every `__setitem__` triggers `subscriber.trigger(...)` → `_run_compute` → `plotter.compute()`.
- `plotter.compute()` does the full O(N) rebuild and stores it in `_cached_state` so the next browser session can pick it up.

So even idle, a day-running dashboard with K timeseries plots does ~K · (extract + autoscaler + hv.Curve) of numpy + HoloViews work each tick, and freshly allocates an N-element datetime coord each tick in `FullHistoryExtractor._to_local_datetime` (`data.assign_coords({dim: self._time_origin + data.coords[dim]})`). The same background thread also drives `JobOrchestrator`, status routing, and stale-session cleanup — anything slow here delays all of them.

> Why does connecting the browser make it worse?

1. **First paint.** Every visible cell calls `SessionComponents.create()`, which builds a `Pipe(data=plotter.get_cached_state())` and a `DynamicMap`. When Panel renders the active tab, Bokeh must build a model for an N-point `hv.Curve` and serialise its ColumnDataSource. Measured ~50–75 ms per plot on the HoloViews side alone; Bokeh + WebSocket adds more for N=100k. With several plots on the active tab this stacks into the perceived freeze.
2. **Empty / slow tabs.** `pn.Tabs(dynamic=True)` defers Bokeh model creation for inactive tabs, which is why they initially look empty. The same defer-and-pay-on-click behaviour explains slow tab switches.
3. **Steady state.** Each session does `pipe.send(full element)` every second. At N=100k that's ~9 ms HoloViews-side server work and ~2.4 MB on the WebSocket per plot. Multiple plots compound; the browser canvas redraw cost compounds again. Reloading "fixes" things by tearing down a heavy document model and starting from a smaller one — at least until the buffer fills back up.

## Measurements

Microbenchmarks (`HvConverter1d.curve()` + `Autoscaler.update_bounds` + `FullHistoryExtractor.extract` on a pre-filled `TemporalBuffer`, then `pipe.send` into a rendered `DynamicMap`). All times in milliseconds, on a single curve.

| N | `extract` | `autoscaler` | `hv.Curve` | `pipe.send` (HV side only) |
|---:|---:|---:|---:|---:|
| 1 000 | 0.05 | 0.14 | (warmup) | 6.7 |
| 10 000 | 0.06 | 0.14 | 1.3 | 6.6 |
| 50 000 | 0.13 | 0.28 | 1.5 | 8.2 |
| 100 000 | 0.15 | 0.31 | 1.7 | 9.4 |

Cumulative-growth run (append 1 sample, do one full tick, repeat) from 1 to 20 000 ticks cost ~32 s of CPU on the background thread; extrapolating to 86 400 ticks (1 day at 1 Hz) gives roughly ~150 s of CPU per plot per day for the build path alone. Linear in plot count.

These numbers do not include the Bokeh-side CDS patch and WebSocket serialisation that fire only with a live document; user-perceived browser lag will be larger than the table suggests.

## Specific code references

- `dashboard/plots.py` `Plotter.compute()` — full element rebuild from `data`; overwrites `_cached_state`. No incremental path.
- `dashboard/scipp_to_holoviews.py` `HvConverter1d.curve()` — constructs `hv.Curve(data=(x, y), …)` with full arrays each call.
- `dashboard/extractors.py` `FullHistoryExtractor.extract()` — allocates a fresh N-element datetime coord each tick via `assign_coords`. Time origin is cached, but the addition and copy are not.
- `dashboard/autoscaler.py` `_update_value_bounds` — `data.nanmin()` / `nanmax()` over full N each tick.
- `dashboard/session_layer.py` `SessionComponents.update_pipe` — `pipe.send(self.presenter.consume_update())`; ships the full element.
- `dashboard/temporal_buffers.py` `TemporalBuffer.DEFAULT_MAX_MEMORY = 20 MB` — never overridden. For scalar f144 logs that's ~625 k points (~7 days at 1 Hz) before ring-buffer trimming kicks in, so buffer growth itself is bounded and is **not** the dominant lag source.
- `dashboard/dashboard_services.py` `_update_interval = 0.2 s` — background thread that runs the orchestrator update (and therefore `_run_compute` for every layer with new data) is the same one driving job and service routing.

## Reproducer outline

1. Run the dashboard against a 1 Hz scalar f144 source (or the existing `fake_logdata` service) and create a single timeseries plot.
2. Let it run with the browser disconnected for long enough that the buffer reaches ~100 k samples.
3. Connect a browser. Observe initial freeze and that the rest of the dashboard (status, job widgets) is also delayed.
4. Note the per-second WebSocket payload while connected.

The per-stage costs above were obtained by running the same component pipeline (`TemporalBuffer.add` → `FullHistoryExtractor.extract` → `Autoscaler.update_bounds` → `HvConverter1d.curve`) in isolation against a synthetic 1 Hz scalar feed, and timing it for buffer sizes 1 k–100 k as well as cumulative 1 → 20 000.

## Why this matters operationally

f144 scalar logs are exactly the long-lived, slow-cadence stream we expect dashboards to display for a full run (hours to days). The cost being O(N) per tick, on a thread shared with job orchestration, means a dashboard left running through a run is increasingly likely to feel laggy and to delay non-plot work, and that the first user to open a browser pays a one-shot cost proportional to total points × visible plots.

Filing this as an issue so we can decide together what to do about it; not proposing a fix here.

SimonHeybrock:
## Summary

Closes #940.

Long-running timeseries plots are laggy because every Kafka delta rebuilds the full `hv.Curve` from the entire buffered history and ships it through `pipe.send`. The timeseries plotter now exposes three time-based knobs — **Period**, **Recent Window**, **Floor Period** — that together bound the displayed point count and throttle plot updates to the chosen period.

Defaults (1 s / 1 h / 5 min) keep a multi-day 1 Hz run near ~4 000 points indefinitely, regardless of buffer size.

By setting "Floor Period" to zero the user can achieve display of a moving window such as the last hour.

## Example

Downsampling settings in plot config
Plot showing transition from downsampled floor rate to downsampled rate in recent window

## Design notes

- Downsampling lives at the plotter, not the extractor. The subscription stays a plain full-history pull; per-plot config can change without re-subscribing.
- The throttle short-circuits `compute()` when no new data has crossed `Period`, which means no autoscaler run, no `hv.Curve` build, no `_set_cached_state`, no presenter dirty bit, and crucially no `pipe.send` / Bokeh patch / WebSocket flush / browser repaint.
- The extractor's `_to_local_datetime` O(N) work is **not** skipped — that would need a gate in `DataService`. The skipped portion is the dominant cost (~100 ms `pipe.send` at N=100 k per the issue's measurements).
- Knobs are time-based, not point-based: users reason in time, not in budgets. `max_points` was considered and dropped.
- Bucket boundaries are anchored to the epoch (fixed time grid), so kept samples don't slide as new data arrives. The recent-window cutoff is quantized to the floor period; actual recent length runs from `recent_seconds` to `recent_seconds + floor_period_seconds` (soft lower bound).


src/ess/livedata/dashboard/plot_params.py
```diff
@@ -305,6 +305,67 @@ class RateNormalizationParams(pydantic.BaseModel):
...
+        title="Recent Window (s)",
+        ge=0.0,
+    )
+    floor_period_seconds: float = pydantic.Field(
+        default=300.0,
+        description=(
+            "Coarser sample period for data older than the Recent Window. "
+            "Set to 0 to drop older data entirely."
+        ),
+        title="Floor Period (s)",
```
nvaytet: I was a little confused by the name 'Floor Period' as to what it meant until I read the description.
Maybe it's just a term that I am not aware of.
I imagined it could be called something like 'Coarse sampling period', which to me wouold make more immediate sense, but maybe I'm the ignorant one here?

src/ess/livedata/dashboard/plots.py
```diff
@@ -661,6 +680,71 @@ def from_params(cls, params: PlotParams1d):
             params, normalize_to_rate=params.rate.normalize_to_rate
         )
 
+    @classmethod
+    def from_timeseries_params(cls, params: PlotParamsTimeseries):
+        """Create LinePlotter for the timeseries plotter, with downsampling on.
+
+        Downsampling and update throttling live at the plotter rather than at
+        the extractor: the subscription stays a plain full-history pull, and
```
nvaytet:
```suggestion
        the extractor: the subscription still pulls the full-history, and
```

src/ess/livedata/dashboard/plots.py
```diff
@@ -661,6 +680,71 @@ def from_params(cls, params: PlotParams1d):
...
+        standard plot path.
+        """
+        if self._downsampling is None:
+            super().compute(data, title_resolver=title_resolver, **kwargs)
+            return
+        primary = data.get(PRIMARY, {})
+        latest_ns = _latest_time_ns(primary)
+        if self._should_skip_for_throttle(latest_ns):
+            return
+        ds = self._downsampling
```
nvaytet: The `ds` name can be confused with Dataset in the Scipp world (even though I know Dataset is no longer used very much).
Is the declaration of this local variable here only to make lines shorter below? Is the readability so much worse if we just use `self._downsampling` everywhere?

src/ess/livedata/dashboard/timeseries_downsample.py
```diff
@@ -0,0 +1,98 @@
...
+    Returns
+    -------
+    :
+        New DataArray with the same dims/coords structure but fewer points.
+    """
+    if concat_dim not in data.dims:
+        return data
+    n = data.sizes[concat_dim]
+    if n < 2 or concat_dim not in data.coords:
+        return data
```
nvaytet:
```suggestion
    if concat_dim not in (data.dims + tuple(data.coords)):
        return data
    if data.sizes[concat_dim] < 2:
        return data
```

src/ess/livedata/dashboard/timeseries_downsample.py
```diff
@@ -0,0 +1,98 @@
...
+    trailing_true = sc.array(dims=[concat_dim], values=np.array([True]))
+    keep_mask = sc.concat(
+        [bucket_changes | band_changes, trailing_true], dim=concat_dim
+    )
+    if drop_older:
+        keep_mask = keep_mask & is_recent
+
+    if keep_mask.values.all():
+        return data
+    return data[keep_mask]
```
nvaytet, line 97: So if I understand correctly, in the coarse region we only keep one out of X data points.
Would it be better (but maybe more complicated) to average the values inside the coarse bins instead of selecting just one value (so that less information is lost)?

tests/dashboard/plots_test.py
```diff
@@ -2227,3 +2227,110 @@ def test_bars_plotter_normalizes_when_enabled(self, data_key):
...
+        params = PlotParamsTimeseries(
+            downsampling=TimeseriesDownsamplingParams(period_seconds=10.0)
+        )
+        plotter = plots.LinePlotter.from_timeseries_params(params)
+
+        plotter.compute({'primary': {data_key: _make_timeseries_data_array(100)}})
+        # Calling compute() with the same data many times must not recompute.
+        first_state = plotter.get_cached_state()
+        for _ in range(10):
+            plotter.compute({'primary': {data_key: _make_timeseries_data_array(100)}})
```
nvaytet: Do we need to re-call `_make_timeseries_data_array` inside the loop, or can we call it once and store in a local variable?

src/ess/livedata/dashboard/plot_params.py
SimonHeybrock: "Floor" as in "nothing below this". But I am happy to rename. I don't know if "Coarse" is better though?

src/ess/livedata/dashboard/timeseries_downsample.py
SimonHeybrock: I had considered that but thought it would lead to an odd effect by reducing the spread of points in this region.

src/ess/livedata/dashboard/timeseries_downsample.py
SimonHeybrock: Thanks, agreed the two checks were worth tidying. I kept them as separate guards rather than the `data.dims + tuple(data.coords)` form though: that concatenation passes when `concat_dim` is in *either*, so a dim without a matching coord (or vice versa) would slip through and then blow up at `data.coords[concat_dim]` below. So it stays an AND, just split into a dims/coords guard and a size guard for readability.

src/ess/livedata/dashboard/plot_params.py
SimonHeybrock: Renamed to "Coarse" now! Floor was anyway inverted logic (was thinking of frequency)

nvaytet (approved):

PR #943, opened 2026-05-26
Gate workflows on context-stream readiness; unify context declaration model

### LINKED ISSUE #922: LOKI i_of_q workflow fails to start with new geometry artifact (length-0 NXlogs)

SimonHeybrock:
## Why

Sciline-based reduction workflows parametrise their graph on *context* values — motor positions, sample geometry, choppers — that arrive as their own Kafka streams alongside detector data. `StreamProcessor` holds a hidden `None` seed for every such input. If a workflow runs before its context stream has produced a value, the result is one of two production-observed failures: a hard crash (bifrost reads rotation inside `accumulate`, raising `'NoneType' object has no attribute 'ndim'`) or, worse, detector data silently attributed to a default geometry.

Separately, LOKI `i_of_q` does not even construct against the chopper-aware geometry artifact: that artifact represents dynamic geometry as length-0 NXlog placeholders along `depends_on` chains, and essreduce's `reject_time_dependent_transform` raises at workflow build unless the placeholder is replaced with a live value (issue #922).

The pre-existing wiring spread context across mechanisms that had to agree by convention: `AuxSources` doubled as both the user-facing dynamic-aux selector and a context-routing primitive; LOKI's `detector_carriage` was declared twice (once for routing, once for graph wiring) with nothing enforcing the names match; and `Workflow.context_keys` leaked a configuration concern into the runtime protocol — every `Workflow` implementation had to expose it, though only job creation consulted it. On top of that, dynamic aux was keyed by role while primary and context were keyed by on-disk stream name, forcing a role↔stream remap onto the data path.

## What

Three decisions, recorded as ADRs under `docs/developer/adr/`.

**ADR 0002 — gate at the JobManager.** A `Job` carries `gating_streams`; the JobManager keeps a *sticky* per-job set of seen context streams (context arrives intermittently — a motor position is published once, then the stream is quiet — so a value seen once stays available). Each tick folds the available streams into that set and gates any job still missing context into a dedicated `JobState.pending_context` with a warning, clearing it back to `active` once the value is seen. Time-based scheduling is untouched; the gate is layered on top.

The gating axis is *"does this input tolerate absence?"*, not `set_context` vs `accumulate`. Dynamic aux (LOKI monitors) accumulates over time and is never gated. ROI maps a missing value to a well-defined "no ROI" result, so it is routed as ordinary aux, not gated. Only motion/geometry context — whose absence is unrepresentable — is gated. There is no seeding: the gate stays closed until the producer publishes, and primary data arriving while gated is dropped, since detector events with unknown geometry cannot be meaningfully reduced.

**ADR 0003 — one declaration record.** `ContextBinding(stream_name, workflow_key, dependent_sources)` is a single concrete dataclass declarable at two scopes:

- `Instrument.add_context_binding` — source-scoped, for source properties (motion/geometry, and chain patches).
- `SpecHandle.add_context_binding` — spec-scoped, late-bound from `factories.py` so `specs.py` stays free of workflow-key imports (the dashboard imports `specs.py` to render UI).

Two flavours share the record, discriminated by the type of `workflow_key`: an arbitrary Sciline key is fed straight to `set_context`; a `ValueLog` subclass is patched into the NXtransformations chain instead. The chain entry to patch is *derived* from `stream_name` via `Instrument.chain_patch_path`, so the parsed stream definition is the single source of truth and a binding cannot disagree with the stream it patches. `wire_dynamic_transforms` inserts one fused per-component provider that replaces essreduce's `NeXusTransformationChain` provider and writes the latest `ValueLog` sample into the chain — the same seam works whether the factory builds the workflow itself (detector view) or receives a pre-built `LokiWorkflow()`, which is what closes #922. Each chain patch declares its own `ValueLog` subclass so multiple dynamic transforms coexist on one workflow without colliding on a shared key.

`Instrument.resolve_context_keys` merges instrument- and spec-scope bindings by source membership into `context_keys` (→ `set_context`), whose key set becomes the job's `gating_streams` (→ the gate). Bindings live at instrument scope by default; a spec that shares the source but does not consume the geometry value opts out via `SpecHandle.skip_instrument_contexts()`. Instrument-scope-by-default with a noisy opt-out is deliberate: spec-scope-by-default would make a forgotten declaration on a new consumer *silently wrong* (stale geometry), whereas a forgotten opt-out merely gates a workflow on a stream it never reads — visibly never running rather than producing wrong output.

Context injection is folded into `SupportsContext.build`, called once at job creation so graph validation and precompute happen up front. `Workflow.context_keys` is removed from the protocol — the routing layer injects context keys via `build`, the other implementations drop their empty stub, and the protocol reduces to `accumulate`/`finalize`/`clear`. `AuxSources` is slimmed to dynamic, user-selectable aux only.

Routing pickup is a co-requirement, not a follow-up: `gather_source_names` includes context bindings so the namespace actually subscribes to motion streams — without it the gate would never open. Registration-time validators enforce the record's invariants (chain-patch uniqueness, wire-name collisions across scopes and against aux fields, `dependent_sources` against registered source names), and a test walks each chain-patch binding's derived path against the registered NeXus geometry artifact, keeping wrong-path out of the silent-failure class without putting NeXus introspection on the construction hot path.

**ADR 0004 — uniform on-disk stream-name keying.** Every workflow input — primary, aux, context — is keyed on the data path by its on-disk stream name. The role a user and author reason in ("the incident monitor") is resolved to a stream name once, at workflow construction in the instrument factory, and never enters the running workflow. `Job` therefore carries plain stream-name sets (`input_streams`, `gating_streams`) and forwards data untouched; the only `stream_name`→`sciline.Key` translation is the dict lookup inside `accumulate`. This drops the former role↔stream remap from both `Job` and `wire_dynamic_transforms`, and removes one-stream-to-many-role multiplexing (unused, and incompatible with on-disk keying). The naming model and its three translation points are illustrated in the Stream Keying design page.

## Test plan

- [ ] Simon: run LOKI `i_of_q` in dev mode and confirm the rear-bank workflow produces I(Q) after a `detector_carriage` f144 message arrives, and sits in `pending_context` (not stale results) when the producer is missing.
- [ ] Simon: run bifrost `qmap` in dev mode and confirm the gate drops events before rotation arrives (no crash) and opens once both motion devices have published.


docs/developer/adr/0002-context-stream-gating-at-jobmanager.md
```diff
@@ -0,0 +1,152 @@
...
+available streams into each active job's sticky set, then checks
+`job.missing_context(seen)`:
+
+- non-empty → the job is skipped for this tick, its state set to
+  `JobState.pending_context`, and a `pending_context_warning(missing)` recorded.
+- empty, and the job was previously `pending_context` → the warning is cleared and
+  the state returns to `active`, so the job's own processing outcome drives its
+  state from there.
+
+Time-based activation is preserved: `should_start(current_time)` continues to drive
```
jl-wynen: How does the scheduler handle unavailable jobs? They complete very quickly, so does that make the scheduler run them at a high frequency for every input update or does it still batch the inputs as normal?

docs/developer/adr/0003-context-input-declaration-model.md
```diff
@@ -0,0 +1,241 @@
...
+   happened to ignore. A routing primitive doubled as a workflow-execution
+   primitive.
+
+2. **Two parallel declarations for the same stream.** LOKI's `detector_carriage`
+   was declared once for routing and once for graph wiring, with nothing enforcing
+   that the names agree.
+
+3. **`Workflow.context_keys` was a protocol leak.** Only `JobFactory.create`
+   consulted it, only at job-creation time, yet every `Workflow` implementation had
+   to expose it (the actual keys for SPW, empty stubs for `AreaDetectorView` and
```
jl-wynen:
```suggestion
   to expose it (the actual keys for `StreamProcessorWorkflow`, empty stubs for `AreaDetectorView` and
```
right?

docs/developer/adr/0003-context-input-declaration-model.md
```diff
@@ -0,0 +1,241 @@
...
+position). We accept the over-gate: an instrument wiring a stream as context must keep
+the producer healthy. If "plot raw detector with a broken motion controller" becomes a
+real operational need, the answer is to split the spec into a TOA-only and a wavelength
+variant, not to introduce param-dependent gating. Today this is hypothetical; YAGNI.
+
+## Decision
+
+One declaration record, `ContextBinding`, declarable at two scopes, feeding both the
+factory's `context_keys` (wired into `set_context`) and the job's gating set (ADR
+0002).
```
jl-wynen, line 52: I don't understand what this means. And it is not a valid sentence.

docs/developer/adr/0003-context-input-declaration-model.md
```diff
@@ -0,0 +1,241 @@
...
+
+```python
+@dataclass(frozen=True, slots=True, kw_only=True)
+class ContextBinding:
+    stream_name: str                 # internal stream name; also the workflow-facing key
+    workflow_key: Any                # opaque Sciline key, consumed by the factory
```
jl-wynen: What is a 'workflow-facing key' if not `workflow_key`?

docs/developer/adr/0003-context-input-declaration-model.md
```diff
@@ -0,0 +1,241 @@
...
+  `issubclass(workflow_key, ValueLog)`. The chain entry to patch is *derived* from
+  `stream_name` (the f144 substream's `nexus_path`) via `Instrument.chain_patch_path`
+  — there is no separate `transform_path` field, keeping a single source of truth in
+  the parsed stream definitions.
+
+### Two declaration sites
+
+1. **`Instrument.add_context_binding(*, stream_name, dependent_sources, workflow_key)`**
+   — source-scoped. Stored on `Instrument.context_bindings`. Used for instrument-property
+   context (motion/geometry). Chain-patch bindings live at instrument scope only.
```
jl-wynen, line 80: Does this mean that all transformation chain streams have to be bound regardless of whether there is any workflow that needs them? Why can't this be part of the second case below?

Below it explains `skip_instrument_contexts` but that seems to always skip all motion/geometry keys. So does this mean that, e.g., a workflow processing a static loki detector still depends on the moving detector carriage?

docs/developer/adr/0003-context-input-declaration-model.md
```diff
@@ -0,0 +1,241 @@
...
+
+Routing (the Kafka subscription set) is decided per *namespace* at service startup,
+ahead of any specific job, and may over-subscribe; the bandwidth cost of an unused
+f144 stream is negligible. Gating is decided per *job* at `JobFactory.create` and must
+be precise. `config/route_derivation.py:gather_source_names` therefore includes context
+bindings in the subscription set: spec-scope entries by stream name, instrument-scope
+entries when a spec in the namespace shares a source with the binding's
+`dependent_sources`. This pickup is a co-requirement of this ADR, not a follow-up:
+without it the namespace never subscribes to motion streams and the gate stays closed
+indefinitely.
```
jl-wynen, line 166: Why isn't it a sciline build failure to miss a binding? Does livedata insert providers for all inputs regardless of stream bindings?

docs/developer/adr/0003-context-input-declaration-model.md
jl-wynen, line 1: I don't understand where it is decided which streams are gated. ADR0002 says, this is defined by the context binding. But there is no explanation in ADR0003. How is this done?

docs/developer/adr/0004-uniform-stream-name-keying.md
```diff
@@ -0,0 +1,128 @@
...
+## Context
+
+A message's identifier passes through four naming schemes between Kafka and a
+Sciline provider:
+
+- `(topic, source_name)` — raw Kafka identity.
+- **role** — the logical input a workflow author and a user reason in for
+  auxiliary data ("the incident monitor used for normalisation"), which may be
+  filled by any of several physical streams.
+- **on-disk `stream_name`** — the canonical instrument-config name a stream is
```
jl-wynen: Why 'on-disk'? Does livedata write this to disk or read it from a file?

docs/developer/adr/0002-context-stream-gating-at-jobmanager.md
SimonHeybrock: Can you clarify what you by "unavailable"?

In general, the jobs are not fed at more than 1 Hz as the message batcher operates at that rate, but I don't think that was your question?

docs/developer/adr/0003-context-input-declaration-model.md
SimonHeybrock: Confusing terminology, ESSlivedata workflow vs Sciline workflow. Maybe `workflow_key` should be `pipeline_key`.

docs/developer/adr/0003-context-input-declaration-model.md
SimonHeybrock: The `ContextBinding` class declares "dependent sources" (usually detector or monitor). So all bindings may be declared but if no workflow for a dependent source is running then the binding never takes effect, i.e., the stream is just dropped at runtime. All motion streams are on the same Kafka topic so there is not really any overhead that could be avoided.

docs/developer/adr/0003-context-input-declaration-model.md
SimonHeybrock: The Sciline workflow might of course fail later if it gets calles. But the point here is that with ADR 0002 we are not calling Sciline before we have values for all f144 streams it needs (and only those, else a broken motion in bank1 would block jobs for bank2).

docs/developer/adr/0004-uniform-stream-name-keying.md
SimonHeybrock: What is meant here is our convention that we use the NeXus group name as `stream_name`. Will clarify.

docs/developer/adr/0003-context-input-declaration-model.md
SimonHeybrock: Good catch — it was buried. The gated set is not declared separately: `resolve_context_keys` merges the `(spec, source)` context bindings into `{stream_name: workflow_key}`, and that mapping's key set *is* `Job.gating_streams`. So "which streams are gated" == "the resolved context bindings for the job". Made this explicit in the "Resolution at job creation" section.

docs/developer/adr/0002-context-stream-gating-at-jobmanager.md
jl-wynen: It basically was my question. When a job doesn't run because some of its inputs gates are closed, how often does it check whether the job can run?

docs/developer/adr/0003-context-input-declaration-model.md
```diff
@@ -101,10 +101,14 @@ return {binding.stream_name: binding.workflow_key
...
-stream names). The merge lives on `Instrument` rather than inlined in `JobFactory` so
-workflow visualization can resolve the same keys without constructing a job.
+`JobFactory.create` calls it once per job. The resolved `{stream_name: workflow_key}`
+mapping is the single source of the gate ADR 0002 applies: it goes to the factory as
+`context_keys`, and its key set becomes `Job.gating_streams` — the streams whose values
+must be present before the job runs (context wire names equal their stream names). This
+is where "which streams are gated" is decided: the gated set *is* the set of resolved
+context bindings for the `(spec, source)`, nothing more. The merge lives on `Instrument`
+rather than inlined in `JobFactory` so workflow visualization can resolve the same keys
+without constructing a job.
```
jl-wynen, line 111: But where does it exclude specific streams from gating? This now sounds like all contexts are gated.

SimonHeybrock: Yes, all context is gated: We assume that Sciline workflows do not set defaults for any inputs (major risk of correctness otherwise, e.g., if a workflow would read a stale position from a NeXus file and uses that if the stream value is not available). Unless all context consumed by a workflow is available, the job will not run.

I clarified the mechanism a bit more in the latest commit (3818db06).

jl-wynen: How are cases handled that should not be gated? See list in ADR2. Or is the distinction that all aux streams are gated but inputs that arrive by other means are not?

src/ess/livedata/config/instruments/bifrost/factories.py
```diff
@@ -69,18 +69,22 @@ def setup_factories(instrument: Instrument) -> None:
...
-    # on the cut-workflow graph. Routed via context_keys; not NeXus components
-    # loaded by name.
-    instrument.add_log_context_binding(
+    # on the cut-workflow graph. Direct-bind (no chain patch) — declared at
+    # instrument scope so every spec consuming ``unified_detector`` picks them
+    # up. Non-consumers opt out below, co-located with the bindings they negate:
+    # the detector view sums over banks and the ratemeter is counts-only.
+    instrument.add_context_binding(
         stream_name='detector_tank_angle_r0',
+        dependent_sources={'unified_detector'},
```
jl-wynen, line 78: Is this enough to also make the analyzer positions depend on the tank angle?

docs/developer/adr/0003-context-input-declaration-model.md
SimonHeybrock: *All* context streams are gated. Aux streams comprise secondary data streams (monitors in data reduction) and context, so not all aux streams are gated.

src/ess/livedata/config/instruments/loki/factories.py
```diff
@@ -3,11 +3,21 @@
 """LOKI instrument factory implementations."""
 
 from ess.livedata.config import Instrument
+from ess.livedata.config.value_log import ValueLog
 
 from . import specs
 from .specs import SansWorkflowParams, TransmissionMode
 
 
+class DetectorCarriageLog(ValueLog):
```
jl-wynen, line 12: Shouldn't this be in ESSsans?

jl-wynen (commented):
Partial review. Waiting for update

src/ess/livedata/config/instruments/bifrost/factories.py
SimonHeybrock: What this says is: "If the workflow is processing 'unified_detector' then it needs to also consume 'detector_tank_angle_r0'". So if you ever implement a workflow that processes *only* the analyzer positions then it is not enough. But in practice I presume that is unlikely. The mechanism here does not imply any dependency or not without in workflow graph.

During design I did consider whether these dependencies could or should be derived from graph-inspection, but decided against it.

src/ess/livedata/config/instruments/loki/factories.py
SimonHeybrock: You mean something like moving `ValueLog` into `ess.reduce` such that subclasses can be defined in instruments?

jl-wynen: Yes

SimonHeybrock: Probably, if this approach is deemed generally useful even for the file-based workflows. But since file-based workflows load the chain directly this may not be relevant? There is no need to patch individual logs into the workflows, unless we change the overall workflow structure such that the individual chain components (and value logs) are nodes in the Sciline graph.

SimonHeybrock: I don't know if it helps, but this shows how the mechanism in this branch patches the motion logs into the workflow:
image

jl-wynen (commented):
This is too much code for me to meaningfully review all details. But judging by the ADRs and docstrings, this looks ok. Is there something specific you want me to look at?

SimonHeybrock:
> This is too much code for me to meaningfully review all details. But judging by the ADRs and docstrings, this looks ok. Is there something specific you want me to look at?

I think reviewing the ADRs was useful. Given the follow-up work(s) I did, the mechanism seems to work as designed/planned, so I cannot think of anything the requires particular attention.

jl-wynen (approved):

PR #946, opened 2026-05-26
Gate plotter.compute on viewer interest

SimonHeybrock:
## Motivation

`plotter.compute()` runs on every Kafka delta for every layer, regardless of whether anyone is watching. Hidden tabs (Panel `Tabs(dynamic=True)`), backgrounded sessions, and a running backend with no browser attached all still pay the full build cost on every update. For expensive plotters this can consume significant CPU, especially when many plot tabs are used. The motivating trigger for this change was #942, which adds downsampling with significant cost to timeseries plotting.

## Approach

Add an interest gate on `LayerStateMachine`. A layer's plotter only builds when at least one viewer holds an interest token; otherwise the latest Kafka input is stashed and collapsed (last-writer-wins) until a viewer arrives. On a 0→1 token transition with pending input, the build runs synchronously so the same polling pass's component rebuild sees fresh `has_cached_state`.

Tokens are ref-counted via `LayerStateMachine.set_active(token, active)`. The polling loop drives `PlotOrchestrator.activate_layer(layer_id, session_layer, is_active)` per (grid, layer): the currently-visible grid's layers hold tokens; everything else is released. Cleanup of dropped sessions is automatic via `weakref.finalize`, so a session disposed without `PlotGridTabs.shutdown` running cannot leak interest.

Visibility is **orthogonal** to the layer FSM. It is per-(session, layer) — interest is held by SessionLayer instances, not by the layer itself — so it cannot fit into the layer-singular `WAITING_FOR_JOB → … → READY` chain. The gate lives on `LayerStateMachine` because that object already owns the plotter reference and is shared across sessions, but it is decoupled from the lifecycle states.

## Threading

- `stash_pending` runs on the **bg ingestion thread** (data-subscriber callback).
- `set_active` runs on the per-session **polling thread** (Bokeh main).
- Both return a `ComputeTask` when a build is due; the caller invokes `task.run()` on its own thread, outside the gate lock.
- The 0→1 activation flush therefore runs on the polling thread — deliberately, so the same poll pass sees fresh `has_cached_state` and renders without waiting another tick.

## Notes on the design

- `ComputeTask` is a frozen dataclass returned from gate operations so callers run the build outside the lock. `_run_compute` (Kafka delta) and `activate_layer` (tab activation) both funnel through `_dispatch_compute_task` for uniform error reporting and state transitions.
- Plotter replacement via `job_started` clears the stash so old input cannot be flushed through a new plotter.
- `weakref.finalize` auto-releases tokens if the owning SessionLayer is garbage-collected. Explicit `set_active(False)` remains the fast path; the finalizer is belt-and-braces against missed cleanup, and it removes the `id()`-reuse footgun where a stale int could be discarded by an unrelated caller.

## Test plan

- [x] New unit tests on `LayerStateMachine`: stash-without-token, 0→1 flush, only-once flush, multi-token ref counting, intermediate-update collapse, unknown-token release, pending-cleared-on-plotter-replacement, no-flush-before-`job_started`, gc auto-release.
- [x] Existing dashboard suite: 1009 passed (pre-existing Panel/Widget deprecation failures unchanged from main).
- [ ] Manual smoke against `dummy` instrument: hidden-tab plots stop computing; tab switches still show fresh state.

<details>
<summary>Relationship to #944</summary>

#944 implements the same runtime semantics — gating compute on viewer interest with a synchronous 0→1 flush — but places the gate inside `Plotter` itself (`set_active`, `_pending`, `_dirty`, `_active_tokens`, `_build_lock`, a `compute` → `_build` rename). That couples a layer-orchestration concern to every plotter subclass: `CorrelationHistogramPlotter` has to forward `set_active`, `StaticPlotter.set_active` is a no-op stub, six subclasses inherit a `_build`/`compute` split they don't need, and ~30 Plotter test sites need fixtures for token acquisition.

This PR puts the gate on `LayerStateMachine` — which is already per-layer, shared across sessions, and owns both the plotter reference and the lifecycle. Plotter and its subclasses are untouched; no new abstractions in the plotting layer.

| | #944 | this PR |
|---|---|---|
| Files touched | 15 | 5 |
| Plotter API change | yes (`set_active`, `_build`) | none |
| Plotter subclasses refactored | 8 | 0 |
| Plotter unit tests changed | yes (~30 sites) | no |
| Static / correlation special-cases | yes | none |

Benchmarks from #944 (tab-switch latency, hidden-tab CPU) apply unchanged — the build path itself is identical.

</details>


src/ess/livedata/dashboard/plot_data_service.py
```diff
@@ -130,6 +175,11 @@ def job_started(self, plotter: Plotter) -> None:
         self._plotter = plotter
         self._error_message = None
         self._version += 1
+        # Drop any input stashed for a previous plotter; the new plotter's
+        # subscription will repopulate when its first delta arrives.
+        with self._gate_lock:
+            self._pending = None
+            self._dirty = False
```
MridulS, line 185: Why don't we move everything inside the gate_lock? 
```suggestion
        with self._gate_lock:
            self._state = LayerState.WAITING_FOR_DATA
            self._plotter = plotter
            self._error_message = None
            self._version += 1
            self._pending = None
            self._dirty = False
```

`_consume_for_flush` below is looking at `_pending` and `_plotter` both and it should probably be safer to just add all of them inside the lock so there aren't any race conditions?

src/ess/livedata/dashboard/plot_orchestrator.py
MridulS: plotter here isn't used anymore?

src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
```diff
@@ -1095,6 +1106,8 @@ def shutdown(self) -> None:
         if self._subscription_id is not None:
             self._orchestrator.unsubscribe_from_lifecycle(self._subscription_id)
             self._subscription_id = None
+        for layer_id, session_layer in self._session_layers.items():
+            self._orchestrator.activate_layer(layer_id, session_layer, False)
         self._session_layers.clear()
         self._grid_manager.shutdown()
```
MridulS, line 1112: Unless my clauding/VSCoding has failed me, it seems like nothing is actually calling the shutdown method here? So we have these layers and the weakref hanging around until the python's gc clears it.

src/ess/livedata/dashboard/plot_data_service.py
SimonHeybrock: Good catch on the race — `_consume_for_flush` reads the (`_plotter`, `_pending`) pair, so `job_started` does need to swap the plotter and clear the stash atomically w.r.t. the gate, otherwise a concurrent 0→1 `set_active` flushes old input through the new plotter.

I scoped the lock a bit tighter than your suggestion though: `_state`/`_version`/`_error_message` are lifecycle fields serialized by the outer `PlotDataService._lock` and never otherwise touched under `_gate_lock`, so I left them out to keep the two lock domains legible. Only `_plotter`/`_pending`/`_dirty` go under the gate lock. Done in cbc4ff5a.

MridulS (approved):

PR #950, opened 2026-05-28
Fix stale mode: latest in grid templates

SimonHeybrock:
`WindowMode.latest` was dropped in 58ed3524 but four grid templates were not updated, causing Pydantic validation failures when loading them. Replaces all occurrences with `mode: window` (default `window_duration_seconds: 0`, which preserves the "latest single update" behaviour via `LatestValueExtractor`).

The dummy detector template also carried `window_duration_seconds: 1.0` and `aggregation: auto` that were silently ignored under the old `latest` mode. These are removed, leaving the minimal `mode: window` config.

The three cells in `bifrost/monitors.yaml` are intentional: they show the same monitors with different window semantics (latest-only, 30 s rolling, cumulative) simultaneously, which scientists need.


SimonHeybrock:
LGTM

PR #951, opened 2026-05-28
Remove unused SpecRequirements plotter-filtering machinery

SimonHeybrock:
`SpecRequirements` / `requires_aux_sources` was fully wired up (registered on `PlotterSpec`, validated in `get_compatible_plotters_with_spec`, threaded through `plotting_controller.py`) but never actually used: no plotter registration ever passed a non-empty `requires_aux_sources` list, so the validation was always a no-op at runtime.

ROI plotters already gate themselves correctly through `DataRequirements` (`required_coords: ['roi_index', 'x', 'y']`) — if a workflow doesn't produce data with those coords, the ROI plotters simply don't appear as compatible. That's the right mechanism.

Removes `SpecRequirements`, the `spec_requirements` field on `PlotterSpec`, `get_compatible_plotters_with_spec()`, and the `aux_sources` argument threaded through `plotting_controller.py`. The single call site becomes `get_compatible_plotters()`.


SimonHeybrock (commented):
LGTM

PR #952, opened 2026-05-28
Per-axis autoscale toggles and Fit button for plots

### LINKED ISSUE #910: Figure zoom resets when axis scale changes
To reproduce:

* Start accumulating counts in a 1D figure
* Zoom in somewhere on the figure
* Wait, the y-axis scale auto-adjusts to keep the curve centered in the image
*  Next time the scale adjusts, the zoom resets to the original level

### LINKED ISSUE #912: 2D figure color bar not updating (intermittent)
It was observed that the color bar of a 2D figure did not update when the figure accumulated more counts, this lead to the entire figure eventually becoming single colored.
The data in the figure did update (this could be seen because some white areas of the figure were filled in over time).

This was not consistently reproducible, so I don't know how useful this issue is.

SimonHeybrock:
## Motivation

Two requirements pull against each other for streaming plots: stream growth must remain visible (new counts, time-series extending right, ROI sums climbing — otherwise data silently leaves the frame), but interactive zoom/pan must survive the ~1 Hz data updates (a user who zoomed in cannot have the range yanked out from under them every second).

The previous design reconciled these with a thresholded `framewise` toggle: a single `Autoscaler` debounced HoloViews' `framewise=True` to fire only when bounds changed "enough." It had two problems:

- **#910 — zoom resets every time the threshold trips.** Once counts grew past 10 % of the previous extent, `framewise=True` snapped the range back to data, destroying the user's zoom. Early in a run this happens often enough to make interactive inspection impossible.
- **#912 — colorbar can freeze.** If an image's `nanmax` stays inside the threshold band while new tiles fill in, the colorbar never updates and the image eventually washes out to a single colour.

The deeper issue is that HoloViews' `framewise` opt is global across x, y, and the colorbar. There is no way to express "let the colorbar follow data but keep my x/y zoom" — which is the dominant use case for detector images. Per-axis decoupling requires going below the opt surface.

## What changed (user-facing)

Each plot's Bokeh toolbar now carries per-axis lock/unlock buttons (X, Y, Color, as applicable to the plotter) plus a Fit button.

- **Active: (default):** the axis follows the current data extent on every update.
- **Inactive:** the axis is frozen — manual pan/zoom survives indefinitely.
- **Fit:** one-shot snap to the current data extent, regardless of toggle state. Does not change lock state.

State is per browser session and per cell, so two users viewing the same dashboard can zoom independently.

Autoscale icons active on X+Y+color:
Screenshot 2026-05-29 at 08 41 32
Autoscale inactive on X:
Screenshot 2026-05-29 at 08 41 54

1D plots gain the same icons, minus the color-autoscale one.

## What this fixes

- Fixes #910: the user can zoom and the axes stay where they put them. No more silent resets on threshold crossings.
- Fixes #912: the colorbar follows data when its lock is open and stays put when closed. As the issue was only observed sporadically we cannot prove that the intermittent freeze is gone, but recommend closing anyway for now as the mechanism has changed.

## Notes for reviewers

The hook-based range writes (rather than HoloViews `xlim` / `ylim` opts) are necessary, not preference: those opts empirically do not propagate to Bokeh on subsequent renders of a `Pipe`-streamed element. Only `clim` does, and even that has caveats around colorbar freezing. The hook layer is the working surface for live axis updates.

## Test plan

- [x] `python -m pytest tests/dashboard/` — 1445 passed.
- [x] Tester: open a streaming 1-D plot, zoom in, and confirm the zoom survives across many data updates.
- [x] Tester: lock the Y axis, zoom on Y, confirm the Y range stays frozen while X (if unlocked) continues to follow.
- [x] Tester: lock the colorbar on a streaming image, confirm `clim` stays put as new tiles fill in.
- [x] Tester: hit Fit and confirm all axes snap to current data extent without changing lock state.
- [x] Tester: open the same plot in two browser sessions and confirm the lock states are independent.


SimonHeybrock:
I'll merge this as it blocks some other PRs that are necessary to fix dashboard freezed. Happy to iterate on the UI in the future, and general review still welcome as well.

SimonHeybrock:
Test fails are not related.

PR #953, opened 2026-05-29
Fix AdaptiveMessageBatcher oscillation on no-fixed-point loads

### LINKED ISSUE #877: AdaptiveMessageBatcher: oscillation from sensitive de-escalation + mis-classified transition reports
## Summary

While adding integration tests for `AdaptiveMessageBatcher` + `RateAwareMessageBatcher` (see #XXX / PR coming), two interacting issues with the adaptive controller surfaced. Neither is rate-aware-specific — both also apply to `SimpleMessageBatcher` as the inner.

## Finding 1: De-escalation is easy to trigger

`DEESCALATION_HEADROOM_RATIO = 0.75` and `DEESCALATION_UNDERLOAD_THRESHOLD = 3` mean that any workload whose `processing_time_s < 0.75 * batch_length_s` for three batches in a row triggers a de-escalation step. A workload that comfortably fits its window (say, 50 % utilisation) will therefore continuously de-escalate until it overloads at a smaller window, escalates, and de-escalates again — steady oscillation instead of a stable level.

Writing `test_oscillation_preserves_messages` exposed this directly: the first draft used `processing_time=1.0` during the level-2 (2 s) run phase; the controller drifted back to level 0 within a few reports.

## Finding 2: Transition-batch classification is ambiguous

The moment `AdaptiveMessageBatcher` escalates or de-escalates, its own `batch_length_s` property returns the **new** value. But the inner batcher's currently-active batch was created with the **old** window (both `SimpleMessageBatcher` and `RateAwareMessageBatcher` document this: "the current active batch keeps its boundaries"). The next `report_batch(count, processing_time)` therefore compares a processing time that belongs to the old window against the new threshold.

Concrete consequence:

- **After escalation** (e.g., 1 s → 2 s): the first report is almost certainly `processing_time < 0.75 * 2 s`, so it's classified as "underloaded" and increments the de-escalation counter — through no fault of the workload.
- **After de-escalation** (e.g., 2 s → 1.43 s): the first report can be spuriously classified as "overloaded" by the new, smaller threshold.

Combined with finding 1, this means every escalation is *immediately followed* by a free step toward de-escalation, amplifying the oscillation tendency.

## Why both find-ings show up together

Finding 2 contributes one spurious underload report per escalation. Finding 1 converts only three such reports into a level change. So in a workload that has some variability, the controller can be nudged down just by the transition itself, not by the workload genuinely changing.

## Fix sketches (not decided)

Options to discuss:

1. **Skip classification for one cycle after a level change** — the adaptive wrapper tracks "just changed" and treats the next `report_batch` as neutral.
2. **Classify reports against the window the inner batch was actually created with** — the adaptive wrapper captures "the threshold at the time the batch started" and uses that for classification.
3. **Raise `DEESCALATION_HEADROOM_RATIO`** and/or `DEESCALATION_UNDERLOAD_THRESHOLD` so settling behavior becomes the norm. This only mitigates finding 1; finding 2 remains.
4. **Asymmetric counter reset**: after a level change, reset both consecutive counters to 0 (already done) *and* discard the first report. Covers both findings cheaply.

Option 2 is the most physically accurate but couples the adaptive wrapper more tightly to the inner's batch lifecycle. Option 4 is the smallest change and handles both findings.

## Scope / applies to

- `src/ess/livedata/core/message_batcher.py` → `AdaptiveMessageBatcher`
- Affects any inner batcher (`SimpleMessageBatcher`, `RateAwareMessageBatcher`).
- Existing `tests/core/adaptive_batching_scenarios_test.py` does not exhibit these problems because its simulation model treats `batch_length_s` as always matching the active window — the divergence between adaptive's view and the inner's view is the bug these tests don't model.

SimonHeybrock:
## Why

`AdaptiveMessageBatcher` could oscillate indefinitely (0→2→1→0) instead of settling on a stable batch length. An overhead-dominated load that overloads at the base window but fits with headroom one or more steps up has *no* level where utilisation lands in the stable dead zone: it escalates, finds every escalated level "underloaded", de-escalates back until it overloads again, and repeats. This is Finding 1 of #877, and it reproduces deterministically with a flat ~1.05 s processing time.

## The fix

The dead zone is `[headroom, 1.0]` in utilisation, and de-escalating one half-step raises utilisation by the consecutive-level ratio. For every workload to have a stable level, the dead zone must span one such step (`headroom ≤ 1 / ratio`). The nominal ratio is √2 ≈ 1.414, but pulse-quantization rounds windows so the widest consecutive ratio is ~1.43 (`round(14·√2)/14 = 20/14`), requiring `headroom ≤ ~0.70`. The previous `0.75` left exactly the gap that caused the oscillation. I verified `1/√2 ≈ 0.707` is still insufficient at the 1.0→1.43 grid step; `0.70` is the binding value.

This deliberately does not address Finding 2 (a report after a level change is classified against the new window though its batch ran under the old one). In isolation it only ever wastes a single report, since the next steady cycle resets the counter — it is an amplifier of Finding 1, not an independent problem. The discard-the-first-report mitigation would add state to suppress a real signal for near-zero benefit, and the wider dead zone shrinks its relevance further. It is pinned as a documented-limitation test instead; the faithful fix, if ever warranted, is to classify each report against the window its batch actually ran under.

## Notes

The remaining genuine limitation (a load landing in the 70–100% dead zone at an escalated level cannot de-escalate even if a lower level would fit) is unchanged and still documented by `dead_zone_stuck`. Removing that stickiness needs active probing of lower levels, which is out of scope here.

## Test plan

- [x] `pytest tests/core/` — 406 passed.
- [x] New `TestNoFixedPointLoad` scenario fails at `0.75` and passes at `0.70`, confirming it is a regression test for the fix.

Closes #877

### DIFF
--- a/src/ess/livedata/core/message_batcher.py
+++ b/src/ess/livedata/core/message_batcher.py
@@ -189,7 +189,18 @@ class SimpleMessageBatcher(MessageBatcher):
 
 ESCALATION_OVERLOAD_THRESHOLD = 2
 ESCALATION_HALF_STEPS = 2
-DEESCALATION_HEADROOM_RATIO = 0.75
+# A batch is "underloaded" (de-escalation candidate) when its processing time is
+# below this fraction of the window; the stable "dead zone" is therefore
+# [ratio, 1.0] in utilisation.  De-escalating one half-step shrinks the window by
+# 1/sqrt(2), i.e. raises utilisation by the consecutive-level ratio.  For every
+# workload to have a stable level (utilisation landing in the dead zone at *some*
+# level) the dead zone must span one such step: ratio <= 1 / (level ratio).  The
+# nominal ratio is sqrt(2) ~ 1.414, but pulse-quantization rounds windows so the
+# widest consecutive ratio is ~1.43 (e.g. round(14*sqrt(2))/14 = 20/14).  Hence
+# ratio <= 1/1.43 ~ 0.70.  A larger value (e.g. 0.75) leaves a gap where a
+# workload is underloaded at every escalated level yet overloaded one step down,
+# causing steady oscillation.
+DEESCALATION_HEADROOM_RATIO = 0.70
 DEESCALATION_UNDERLOAD_THRESHOLD = 3
 DEESCALATION_IDLE_WINDOWS = 3
 
--- a/tests/core/adaptive_batching_scenarios_test.py
+++ b/tests/core/adaptive_batching_scenarios_test.py
@@ -117,6 +117,15 @@ LIMITS: dict[str, dict[str, float]] = {
         "min_level": 1,
         "min_final_level": 1,
     },
+    # -- No-fixed-point load (issue #877, Finding 1) ----------------------
+    # A flat ~1.05s processing time overloads at level 0 (1.0s) but fits with
+    # headroom at every escalated level.  The dead zone must be wide enough to
+    # give it a stable level; otherwise it oscillates 0->2->1->0 forever.
+    "overhead_dominated_no_fixed_point": {
+        "min_level_reached": 1,
+        "max_oscillations": 1,
+        "min_final_level": 1,
+    },
     # -- Creeping overload ------------------------------------------------
     "creeping_overload": {
         "min_level_reached": 4,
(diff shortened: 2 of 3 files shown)


SimonHeybrock:
LGTM

PR #957, opened 2026-06-02
Fix Download button icon centering broken by Panel 1.9+

SimonHeybrock:
Panel 1.9 changed how `FileDownload` processes `color`/`button_type`: the value is stripped before it reaches the Bokeh model, which always retains `button_type='default'`. The rendered button therefore has CSS class `bk-btn-default`, never `bk-btn-primary`.

The old CSS selector `:host(.solid) button.bk-btn.bk-btn-primary` never matched, silently disabling all flex-centering styles. Without them, Panel's default `.bk-btn a { display: inline; padding: 6px }` placed the icon at bottom-left.

Fix: change selector to `bk-btn-default` and drop the misleading `color='primary'` argument (it has no effect on the Bokeh model or DOM).

### DIFF
--- a/src/ess/livedata/dashboard/widgets/buttons.py
+++ b/src/ess/livedata/dashboard/widgets/buttons.py
@@ -51,8 +51,9 @@ def create_tool_button_stylesheet(
         RGBA color for the hover background.
     selector:
         CSS selector for the button element. Default is 'button' for regular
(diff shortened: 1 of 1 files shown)


SimonHeybrock (commented):
Works in UI

PR #960, opened 2026-06-03
Cell titlebar: rename, add-layer, toolbar toggle, and freshness indicator

### LINKED ISSUE #646: Plot header toolbars: Support collapsing or grouping?
### Overview

Each plot layer adds its own toolbar. This is usually ok with two layers, but with three or more it begins taking a lot of space. An example where this occurs are ROI plots, which might need three or even five layers. There are a couple of ways to address this:

- Add button to expand/collapse the header rows.
- Allow defining layer-groups, potentially with custom names. Groups could also support expand/collapse, to provide access to individual settings.

### Related problems and considerations

For plots with many layers, the setup in itself is tedious and error prone for non-experts. Do we need a way to pre-define layer groups so they can be created in one step? For example, for a ROI plot we would usually need to create individual layers for the image, the readback, and the request. Could we allow for selecting multiple outputs, and have a compatible layer group be displayed for creation? Or do we need a separate wizard for this? One key thing to note is that the current mechanism is also error-prone and bad UX: We need to make sure to select the same source name (detector name) for each layer. This should be "automatic" or enforced somehow &mdash; indicating again that the current approach is bad UX flow. The default is `workflow->output->plotter->source_name`, do we need a different order/subset for layer-groups? Maybe `plotter->workflow->source_name` (`output` being implicit, if the layer group depends on a specific set of outputs)?

### Considered alternatives

Combine things into layers, so we have fewer. I don't think this is a solution since it (1) complicates the mechanism and (2) does not provide easy access to plot params via the UI (unless we make more complicated dialogs).

SimonHeybrock:
## Motivation

Each plot layer carried its own toolbar. With two layers this is fine, but ROI-style plots routinely need three to five layers (image, readback, request, ...), and the stacked toolbars consume a large share of the cell's vertical space. At the same time, cell-level chrome (a title, add-layer) had no natural home and was either missing or duplicated per layer, and the lag/time indicator lived inside the HoloViews plot title — which both ate plot space and was expensive, minting a fresh `OptionTree` on every tick via `result.opts(title=...)`. Last but not least, the lag/time indicated was actually only visible in single-layer plots as composing layers using `hv.Overlay` hid the plot titles.

### UI Examples

Collapsed details (default):
screenshot-2026-06-04T13:45:08
Cell config modal:
screenshot-2026-06-04T13:57:33
Expanded:
screenshot-2026-06-04T13:58:13

### Description

A new cell-level titlebar sits above the per-layer toolbars and absorbs the cell-level chrome:

- A user-editable cell title (modal rename, persisted and propagated cross-session) with a derived default when unset.
- Add-layer consolidated into a single per-cell button whose dropdown is the union of every layer's overlay suggestions, each remembering its source layer so overlays still inherit the right workflow/sources.
- A toggle that hides/shows the per-layer toolbars. Toolbars are hidden by default for a compact monitoring view; the choice is per-session local state.

The lag/time indicator moves out of the plot title and into the titlebar. Plotters now expose raw `TimeBounds` from `compute()` rather than formatting a string into the title. The titlebar shows a colour-banded freshness pill (worst-case lag across the cell's layers, recomputed against the wall clock each poll, so a stalled stream visibly goes stale), and per-layer time-range rows carry the full start–end/lag detail, revealed by the toolbar-visibility toggle. Updates happen in place via `pane.object`, a single cheap property write instead of rebuilding the option store.

Per-layer toolbars are slimmed to gear + close.

## Default cell title

The derived default is deliberately minimal: the first non-static layer's workflow title, plus the source name only when the cell's layers share a single source (a lone source is otherwise not visible in the plot labels). Output names and window info are dropped.

## Tradeoffs

- **Minimal default title.** Base-workflow output names (`Image`, `Histogram`) are implied by the workflow and add noise, and window/freshness now lives in the titlebar indicator, so neither is repeated in the title. The cost is that distinct reduction outputs sharing one workflow (e.g. `I(Q)` and `I(t)`) collapse to the same default and must be told apart with an explicit title. We considered a per-output "primary output" flag to selectively surface the meaningful names, but deferred it as over-engineering: the explicit-title escape hatch already exists, and we would rather see it prove insufficient before adding spec-wide annotation machinery.
- **Toolbars hidden by default.** Optimises for the common compact monitoring view at the cost of an extra click to reach per-layer settings.

## Relation to #646

Fixes #646 — the toolbar show/hide toggle provides the expand/collapse the issue asks for. The issue's separate idea of user-defined *layer groups* (and group-aware creation flows) is not addressed here, but likely obsolete given that we have the derived-layer mechanism now.

## Test plan

- [ ] Cell titlebar renders; rename modal persists and survives a page reload / second session.
- [ ] Per-cell add-layer dropdown lists the union of layer suggestions and creates layers with the right workflow/source.
- [ ] Toolbar toggle hides/shows per-layer toolbars; state is per-session.
- [ ] Freshness pill reflects the worst-case lag and grows stale when a stream stalls.
- [ ] Default title shows `Workflow (Source)` for single-source cells and the workflow title alone otherwise.

### DIFF
--- a/src/ess/livedata/dashboard/correlation_plotter.py
+++ b/src/ess/livedata/dashboard/correlation_plotter.py
@@ -29,7 +29,7 @@ from .plot_params import (
     PlotDisplayParams1d,
     PlotDisplayParams2d,
 )
-from .plots import ImagePlotter, LinePlotter, PresenterBase, TitleResolver
+from .plots import ImagePlotter, LinePlotter, PresenterBase, TimeBounds, TitleResolver
 from .range_hook import Axis, RangeTargets
 
 
@@ -256,6 +256,11 @@ class CorrelationHistogramPlotter:
     def get_cached_state(self) -> Any | None:
         """Get the last computed state from the renderer."""
         return self._renderer.get_cached_state()
+
+    @property
+    def time_bounds(self) -> TimeBounds | None:
+        """Time bounds of the most recently computed data (from the renderer)."""
+        return self._renderer.time_bounds
 
     def has_cached_state(self) -> bool:
         """Check if the renderer has computed state."""
--- a/src/ess/livedata/dashboard/plot_orchestrator.py
+++ b/src/ess/livedata/dashboard/plot_orchestrator.py
@@ -203,10 +203,14 @@ class PlotCell:
     The plots are placed in the given row and col of a :py:class:`PlotGrid`, spanning
     the given number of rows and columns. A cell can contain multiple layers that
     are composed via hv.Overlay.
+
+    ``user_title`` is an optional user-defined cell title shown in the cell
+    titlebar. When ``None`` the titlebar shows a title derived from the layers.
     """
 
     geometry: CellGeometry
     layers: list[Layer]
+    user_title: str | None = None
 
 
 @dataclass
@@ -616,7 +620,12 @@ class PlotOrchestrator:
         self._notify_grid_created(new_grid_id)
         return new_grid_id
 
-    def add_cell(self, grid_id: GridId, geometry: CellGeometry) -> CellId:
+    def add_cell(
+        self,
+        grid_id: GridId,
+        geometry: CellGeometry,
+        user_title: str | None = None,
+    ) -> CellId:
         """
         Add an empty cell to a grid.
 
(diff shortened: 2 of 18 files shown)


SimonHeybrock:
LGTM. UI may need more tweaking, time will tell.

PR #961, opened 2026-06-08
Reset-on-move accumulator for monitor + detector views

### LINKED ISSUE #828: Automatic accumulator reset on position context change

SimonHeybrock:
## What this is

Makes a view's cumulative histogram **reset when the viewed component moves** instead of crashing (#828), and demonstrates **monitor motion** (#780) on the existing context-binding / chain-patch mechanism. Proven end-to-end for a moving LOKI detector and a moving DREAM monitor.

Closes #828. Follow-up #965 tracks robustness to noisy motor readback.

Targets `jobmanager-context-gate` (not `main`); review the diff against that branch.

## #828 — reset on geometry change

A cumulative histogram summed across windows would crash (`self._value += value`, coord mismatch) when the component moves mid-run. The geometry coord is already on the data, so the accumulator already knows the configuration changed — treat the mismatch as a reset. `NoCopyAccumulator` gains a `reset_coord` knob (via `make_no_copy_accumulator_pair`): a differing scalar coord discards the buffer and restarts. An absent coord or `reset_coord=None` is a no-op, so it is safe in TOA mode and for reduction workflows.

- **Monitor views** default to `reset_coord='position'` (changes on *any* move, unlike `Ltotal` which misses a lateral move).
- **Detector views** reset on the resolved `NeXusTransformation` (the live, carriage-driven placement), stamped as the scalar `detector_transform` coord at the single histogram funnel. Projection-independent — works for geometric and logical, TOA and wavelength. A derived per-projection scalar would miss a lateral move and crash. Sources without live geometry stamp nothing (no-op): file-less logical views and identity-sentinel chains. The spatial transform is dropped by the da00 serializer, so the coord stays internal.

The reset lives on both accumulators; the window's never fires today (one push per finalize) but guards against that ever changing.

## Deterministic projector noise (scipp/ess#629)

The geometric projector's screen edges are recomputed on every context update, and essreduce's `position_noise_for_cylindrical_pixel` used an unseeded RNG — so each `detector_carriage` f144 sample (moved or not) jittered the edges and crashed the cumulative `+=`. Not motion-specific: a *parked* detector streaming its position would crash. essreduce 26.6.1 seeds the noise upstream (scipp/ess#629), so recomputes are reproducible while a real move still shifts the edges (handled by the reset). This PR bumps to `essreduce>=26.6.1`;

## #780 — monitor motion

Works with **zero monitor-workflow changes**: `wire_dynamic_transforms` is already component-agnostic. A moving monitor needs only an instrument chain-patch binding plus the monitor's f144 NXlog in its `depends_on` chain — same recipe as the LOKI carriage. `TestMonitorMotion` drives the real wavelength workflow (DREAM geometry + lookup table) and asserts the spectrum tracks the streamed position.

## Test plan

- [x] Moving LOKI detector carriage resets rear detector view without warnings/errors

## What remains (not in this PR)

- **#780:** no real LOKI monitor binding — the latest file's monitor transform `trans_20` is a placeholder (`source='source'`) and the real motor PV sits outside the monitor's `depends_on` chain; geometry-script and orphan-NXlog coverage also remain.
- **Readback noise (#965):** the exact coord comparison would reset every cycle against a noisy live readback — to be solved orthogonally by patching the **setpoint** (`target_value`) not the readback (distinct from the projector-noise seed above).
- A window containing the move histograms all its events — including pre-move ones — under the post-move transform: a one-window labelling inaccuracy, not corruption. Reduction workflows opt out of reset entirely (`reset_coord=None` default).


configs/log_producer_bifrost.json
```diff
@@ -6,9 +6,9 @@
             "max": 90.0,
             "step": 0.1,
             "initial": 0.0,
-            "value_stream": "detector_tank_angle_r0_value",
-            "target_stream": "detector_tank_angle_r0_target_value",
-            "idle_stream": "detector_tank_angle_r0_idle_flag",
+            "value_stream": "detector_tank_angle_r0/value",
+            "target_stream": "detector_tank_angle_r0/target_value",
+            "idle_stream": "detector_tank_angle_r0/idle_flag",
```
nvaytet, line 11: Are these something that we changed ourselves or were they changes on the ecdc side?

src/ess/livedata/handlers/monitor_workflow.py
```diff
@@ -234,7 +243,7 @@ def create_monitor_workflow(
     # Only accumulate CumulativeMonitorHistogram and WindowMonitorHistogram.
     # MonitorCountsTotal and MonitorCountsInRange are computed from
     # WindowMonitorHistogram during finalize, not accumulated separately.
-    cumulative, window = make_no_copy_accumulator_pair()
+    cumulative, window = make_no_copy_accumulator_pair(reset_coord=reset_coord)
```
nvaytet, line 302: I didn't completely understand why for the detector workflow we are monitoring the `NeXusTransformation`, while for the monitor we check the `position` coord.

I guess ultimately they probably boil down to the same thing, but it was confusing to have different approaches.

nvaytet, line 302: I didn't completely understand why for the detector workflow we are monitoring the `NeXusTransformation`, while for the monitor we check the `position` coord.

I guess ultimately they probably boil down to the same thing, but it was confusing to have different approaches.

tests/handlers/monitor_workflow_test.py
```diff
@@ -771,3 +780,198 @@ def test_wavelength_mode_with_histogram_input(
...
+    def _position_log(z: float) -> sc.DataArray:
+        """An f144 z-position log holding a single sample, in metres."""
+        return sc.DataArray(
+            sc.array(dims=['time'], values=[z], unit='m'),
+            coords={
+                'time': sc.array(dims=['time'], values=[0], unit='ns', dtype='int64')
+            },
+        )
+
+    def _build(
```
nvaytet, line 862: Difficult to follow what goes on in these tests with such a large setup, but it is probably unavoidable on such a complex system...

configs/log_producer_bifrost.json
SimonHeybrock: Changed in ESSlivedata.

src/ess/livedata/handlers/monitor_workflow.py
SimonHeybrock: `position` on a detector is a per-pixel array. But the accumulators work on projected detectors in many cases, so the position would need to be projected somehow as well. This requires bespoke code, and could also miss certain movements. So the real answer was to use a *scalar* for detecting the detector move. It cannot be something like the position of the detector origin as that cannot detector detector rotations. So we ended up with the full transformation.

tests/handlers/monitor_workflow_test.py
SimonHeybrock: I think it is unfortunately unavoidable for the test at this level. There are unit tests for lower-levels (like the accumulator itself), but this mechanism spans several moving (😬) parts so I wanted a test that covers the entire thing.

src/ess/livedata/handlers/monitor_workflow.py
nvaytet: And can we use the transformation for the monitor as well? Some monitors appear to have (a small number of) pixels?

SimonHeybrock: Done.

nvaytet (approved):

Issue #962, opened 2026-06-08
BIFROST: unpin detector geometry once a corrected NeXus file is available


SimonHeybrock:
## Summary

BIFROST's detector geometry is temporarily **pinned to the old geometry artifact** (`geometry-bifrost-2025-01-01.nxs`) while the chopper workflow consumes a newer one (`geometry-bifrost-2026-06-08.nxs`). Both are in the geometry registry; each consumer selects the file in which its data is valid. This issue tracks collapsing back to a single artifact once a corrected source file is available.

## Why the split exists

The new source file `coda_bifrost_999999_00006061.hdf` has correct, up-to-date **chopper** geometry (new group names `bandwidth_chopper_*`, `frame_overlap_chopper_*`, `pulse_shaping_chopper_*`; pulse-shaping choppers moved ~7 cm), but a **broken detector transformation chain**:

- The `detector_tank_angle` group lost its numeric prefix (was `117_detector_tank_angle`), but `detector_tank_angle_r0/value` still has `depends_on` → `/entry/instrument/117_detector_tank_angle/transformations/detector_tank_angle_t0` (stale prefix).
- The static `detector_tank_angle_t0` zero-offset node is **missing entirely** from the file.

Loading the detector geometry therefore raises a dangling-`depends_on` warning and fails the reduction. The chopper workflow is unaffected — it loads only `NXdisk_chopper` + `NXsource`, never the detector chain. (Reported upstream to the file-writer side.)

The missing `_t0` value cannot be safely reconstructed: in the old file it was 0.1192°, identical to the choppers' `*_t0_r` offset; in the new file that offset was re-surveyed to 0.0974°. Grafting the old value would inject a known-stale number; inferring 0.0974° is unverified. So we pin rather than fabricate.

## Cleanup when a corrected file lands

A single NeXus file valid for **both** detectors and choppers (complete `detector_tank_angle` chain incl. `_t0`, new chopper names):

- Regenerate the geometry artifact (`ess-livedata-make-geometry-nexus`) and `streams_parsed.py` (`nexus_helpers --generate`) from it; add one new registry entry in `detector_data_handler.py`.
- In `bifrost/factories.py`:
  - Drop the `date=` pin on the reduction's `get_nexus_geometry_filename('bifrost')` call so both consumers use the latest entry.
[...]

PR #963, opened 2026-06-08
Chopper-equipped wavelength-LUT workflow on context bindings

SimonHeybrock:
## Summary

Replaces the chopperless v0 wavelength-LUT scaffolding with the real chopper-equipped workflow, built on the context-binding mechanism (ADR 0002/0003). LOKI is wired as the reference instrument.

The workflow now loads static `NXdisk_chopper` geometry from the NeXus artifact and delegates the `RawChoppers → DiskChoppers` conversion to essreduce's `to_disk_choppers` (essreduce bumped to 26.6.0), dropping the neutron simulator and its `NumberOfSimulatedNeutrons`/`Simulation` parameters. Per-chopper rotation-speed and delay setpoints flow as Sciline context keys, gated at the JobManager.

`ChopperSynthesizer` grows from a single vacuous tick into a per-chopper plateau detector: it locks each chopper's noisy delay readback into a synthetic `delay_setpoint`, caches the clean rotation-speed setpoint, and fires the `chopper_cascade` primary trigger once all choppers are locked. The dummy instrument's chopperless spec is removed.

## Example

Loki chopper cascade driven by fake setpoint and delay streams:
image

## Why

In-process synthesis of the cascade trigger is expected to be permanent — an upstream "all choppers reached setpoints" signal may never exist. What remains provisional is the chopper quantity set and how each is treated (delay vs. phase; which quantities arrive as setpoints vs. plateau-detected readbacks). The docs and code now state that distinction and enumerate the sites a future quantity change must touch.

## Notes for reviewers

- `configs/log_producer_loki.json` and the `log_producer_widget` noise/periodic-publish additions are dev-only simulation tooling — they are how the synthesizer is exercised manually (a noisy delay readback for the plateau detector to lock onto).
- `chopper_delay_atol_ns` is a ns threshold on `Instrument`; the readback stream's `ns` unit is enforced at config time. Loosen per-instrument once real readback noise is measured.
- BIFROST wiring is kept as a follow-up: it carries a geometry workaround (two artifacts, blocked on #962) that should not gate this mechanism PR.

## Follow-up work

I am considering if and how to split the workflow output per-component. Should we run one workflow per monitor and detector? Or simply post-process the single output table? Or add a slicer plotter that allows for selection of a fixed slice?

## Test plan

- [x] `pytest -m "not integration and not slow"` green
- [x] Manual: run the timeseries service for LOKI with the dev log-producer, confirm `chopper_cascade` fires after delay readbacks plateau and the LUT recomputes.


pyproject.toml
```diff
@@ -191,6 +191,8 @@ ignore-words-list = [
     # But we use the word in our code of conduct which is the contributor covenant.
     # Let's not modify it if we don't have to.
     "socio-economic",
+    # FO is a technical designation for LOKI's choppers (Fast Opening or similar)
```
SimonHeybrock:
```suggestion
    # FO is a technical designation for LOKI's frame-overlap choppers
```

src/ess/livedata/config/chopper.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def declare_chopper_setpoint_streams(
+    streams: dict[str, Stream], choppers: Sequence[str]
+) -> None:
+    """Declare the synthetic ``delay_setpoint`` streams the synthesizer emits.
+
+    The ``ChopperSynthesizer`` (timeseries service) plateau-detects each
+    chopper's noisy ``delay`` readback and injects a synthetic
+    ``<chopper>/delay_setpoint`` f144. It is not a Kafka topic (no topic/source,
```
nvaytet, line 59: Is this needed because there will not be a `delay_setpoint` in kafka?
It sounds like it's going to be problematic if there is no setpoint (also in the files) 😞

src/ess/livedata/config/instrument.py
```diff
@@ -123,6 +123,18 @@ class Instrument:
...
+    #: ``DiskChoppers`` and declares per-chopper setpoint context bindings) and
+    #: the ``ChopperSynthesizer`` wired into the timeseries service.
+    choppers: list[str] = field(default_factory=list)
+    #: Plateau-detection tolerance for chopper delay readbacks, in the delay
+    #: stream's own unit (ns for LOKI). Used by ``ChopperSynthesizer`` for both
+    #: noise rejection (rolling-window std must be below this) and change
+    #: detection (drift since the last lock). 1 us is a tight default suitable
+    #: for sub-degree phase tracking on typical ESS choppers; loosen
+    #: per-instrument once real readback noise is measured.
+    chopper_delay_atol: float = 1000.0
```
nvaytet: I did not understand (also from the PR description) why this does not have a unit.

src/ess/livedata/handlers/wavelength_lut_workflow.py
```diff
@@ -86,37 +186,84 @@ def _attach_provenance(table: LookupTable, params: ParamsKey) -> WavelengthLut:
 
 
 def _build_pipeline(params: WavelengthLutParams) -> sciline.Pipeline:
-    wf = LookupTableWorkflow()
+    wf = GenericUnwrapWorkflow(
+        run_types=[AnyRun], monitor_types=[], wavelength_from='analytical'
+    )
     wf[PulsePeriod] = params.pulse.get_period()
-    wf[PulseStride] = int(params.pulse.stride)
+    wf[PulseStride[AnyRun]] = int(params.pulse.stride)
```
nvaytet: Note that we now have a mechanism that can guess the pulse stride from the chopper frequencies (leaving choppers with 0 frequency out).
If you want to simplify the UI...

src/ess/livedata/config/chopper.py
SimonHeybrock: It is still unclear whether we will get a setpoint. See details in the synthesizer module docstring.

src/ess/livedata/handlers/wavelength_lut_workflow.py
SimonHeybrock: ... by not setting this parameter at all? Does the workflow do it automatically?

nvaytet: Yes, the workflow will do it automatically if the param is not set.

src/ess/livedata/config/instrument.py
SimonHeybrock: Technically because f144 does not carry units. But actually we can lookup the unit from parsed NeXus files. Doing that now.

nvaytet (approved):

src/ess/livedata/handlers/wavelength_lut_workflow.py
SimonHeybrock: Have you found it generally reliable? As there is no way to work around bad detection in livedata we have to be sure it works before removing the option -- or add an `GuessStrideFromRotationSpeed` checkbox. I think I'll do the latter.

SimonHeybrock: @nvaytet See latest commit

nvaytet (approved):

PR #967, opened 2026-06-09
Show data age in freshness pill; demote pipeline lag to detail

SimonHeybrock:
## Motivation

Moving the lag indicator into the cell titlebar (#960) changed it to report the wall-clock age of the displayed data (`now − data_end`), recomputed each poll, so it ticks up while a stream is stalled. That looked like a regression versus the old plot-title indicator, which froze the lag at render time. On reflection it is the right behavior: the stalled data really is getting older, and — crucially — an age measured against a shared `now` is comparable across plots and against the experiment control software's clock, whereas a per-plot frozen lag is anchored to each plot's own render moment and so cannot be compared. The original purpose of the indicator was exactly that cross-plot / cross-system correlation, not just pipeline-latency diagnosis.

## Changes

Keeps data age as the headline pill (color-banded by staleness, formatted `3s` → `1m` → `5m` so a dead stream reads as obviously stale) and demotes the frozen pipeline lag to the per-layer toolbar detail row, where it sits next to the absolute time range — the skew-free primitive for correlating with external timestamps. `TimeBounds` records `created_at` so the toolbar can still report that frozen lag, while the pill uses `age_seconds` against the poll clock.

The pill carries no native `title` tooltip. It re-renders as the age ticks (up to once per data frame), which tears down a hover tooltip on every update — flicker at the typical 1 Hz rate, and unavoidable for a hover tooltip on a live-updating element. Detail therefore lives in the toolbar row, which updates without disturbing a hover.

## Test plan

- [x] Watch a live stream: pill age stays low and green, ticking with each frame.
- [x] Stop the stream: age climbs and the band goes stale → old.
- [x] Toggle the toolbar: per-layer row shows the absolute range and frozen lag.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -80,3 +80,19 @@ Python constants rather than CSS variables for centralized color management.
 ## Avoiding flicker
 
 Make sure all widget-updates that touch more than a single widget (or a single widget multiple times) use `pn.io.hold()`.
+
+### Native tooltips on re-rendering panes
+
+A native HTML `title=` tooltip baked into a `pn.pane.HTML` string is torn down
+every time `pane.object` is reassigned — Panel replaces the pane's inner DOM, so
+the browser drops any open hover tooltip. This makes hover tooltips unworkable on
+any element whose content updates frequently (e.g. a live freshness/lag readout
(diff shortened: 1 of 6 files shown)


SimonHeybrock:
LGTM

Issue #968, opened 2026-06-10
MAGIC: Configure instrument with detector and monitor views


SimonHeybrock:
## Overview

- There are two MAGIC detector banks with about 500k and 200k pixels (the latter behind an analyzer).
- Each looks like the DREAM *mantle* detector, but as a vertical cylinder, i.e., along Y instead of Z. Exact dimensions (shape) differs.

## Acceptance criteria

- Configured monitor workflow.
- Wire-view (see DREAM)
- Strip-view (see DREAM)
- Cylinder-Y projection (equivalent to DREAM cylinder-Z projection for mantle) for each banks.

## Extra details

- Active surface in voxels is only on one side (unlike DREAM where it is on two).
- There might not be an `[xyz]_pixel_offset` defined in the NeXus files, only `NXoff_geometry`. Can the generic workflow from `essreduce` load this already?

SimonHeybrock:
## Findings from the reference file and first implementation pass

Investigated `coda_magic_999999_00006299.hdf` and implemented the geometry-free views plus the cylinder-Y projection. Summary of what we learned and the current state.

### Reference file structure
- Two detectors, **no monitors** at all: `magic_detector_a` (491520 voxels, `detector_number` 1–491520) and `magic_detector_b` (131072 voxels, 491521–622592).
- Both carry only a per-voxel `NXoff_geometry` `pixel_shape` (8 vertices/voxel) and **no `[xyz]_pixel_offset`**; `depends_on = '.'` (positioned directly by the OFF vertices).
- Each voxel is a cuboid. Local vertices 0–3 are the inner radial (mantle) face, 4–7 the outer; the other four are side walls. Per-voxel radial depth ≈ 16.5 mm.

### Geometry: can the generic essreduce workflow load this?
**No, not as-is.** essreduce computes pixel positions from `[xyz]_pixel_offset` (via `zip_pixel_offsets`); with only `NXoff_geometry` it raises `KeyError: 'x_pixel_offset'`. It does **not** derive positions from OFF geometry.

[...]

PR #970, opened 2026-06-10
Add MAGIC instrument: detector views, cylinder-Y projection, monitor

SimonHeybrock:
Adds dashboard support for the MAGIC instrument (#968): its two detector banks, the geometry-free wire and strip logical views, the cylinder-Y mantle projection for each bank, the f144 log streams, and the beam monitor.

Rebased onto current `main` and rebuilt against the newer reference file `coda_magic_999999_00015510.hdf`, which settled several things the original version had to guess.

## Example output

image

## Geometry

The MAGIC NeXus files carry only a per-voxel `NXoff_geometry` `pixel_shape` and no `[xyz]_pixel_offset`, which essreduce's position calculation requires. Rather than add a runtime path that reads OFF geometry, `make-geometry-nexus` gains an `--off-active-face` option that derives the pixel offsets from the voxel vertex centroids (only one radial face is the active surface, so a single face is used) and drops the bulky mesh.

The previously registered artifact carried a dangling `depends_on`: the banks point at `detector_a_rotation/value`, but that group was absent, because the file predates the script's `depends_on` chain resolution. This surfaced only at runtime — the workflow built fine and then failed on first data. The re-derived artifact has bit-identical pixel offsets and adds the chain groups.

## Logical layout

Both banks are segments of a vertical (Y-axis) cylinder with the same axis semantics, derived from the voxel centroids by checking which coordinate varies along which axis: radius varies only along `wire`, Y only along `strip`, phi only along `segment`. Bank A is 32x128x120, bank B 32x16x256 — the earlier bank B layout (`module`/`cassette`/`strip`/`wire`/`counter`) was wrong on every axis, and bank A's segment count had drifted to half the real value. The `detector_number` ranges are now computed from these sizes rather than restated separately, so a fold can no longer disagree with the pixel count.

The banks hang off live rotation stages, so their placement is only known once the motor readback arrives. The wire and strip views consume no position and opt out.

Projection resolution is one screen bin per strip and per segment. The banks are radially deep, so the mantle projection scales each pixel's Y by `r_min/r` and the wires land at slightly different arc positions; bin counts that are not commensurate with the strip/segment structure alias into a moire of empty bins (bank B was 17% empty at the previous settings).

## Streams

The reference file gives the real Kafka routing: one topic per bank, both with source name `magic`. The original guess of a single `magic_detector` topic with per-bank source names was wrong on both counts.

The beam monitor's Kafka source name upstream is still the placeholder `TODO1`, so the PROD mapping assumes the conventional `cbm1` and will need correcting once ESS assigns the real one.

## Live rotation context

Each bank binds its rotation stage into the `depends_on` chain, and its projection workflows are gated until the readback arrives. Three separate problems left them gated forever.

The binding named the device substream `detector_a_rotation/value`. `DeviceSynthesizer` merges a device's RBV/VAL/DMOV substreams and suppresses them from forwarding, emitting only the merged `detector_a_rotation` stream, so the gate could never open — in PROD either, not just dev. BIFROST and LOKI bind the device name; MAGIC was the odd one out. A substream binding is now rejected at config time, pointing at the owning device, since it cannot work for any instrument.

The DEV stream mapping was built without `log_names`, leaving no f144 route installed at all.

The dev-mode log producer widget had no config for MAGIC, so nothing drove the stages. The new one exposes a slider per bank. Its 0-90 degree range is provisional: the real travel lives in the EPICS motor record rather than in the repo or the geometry artifact, and the banks' 62 and 123 degree arcs describe detector extent, not stage travel.

Nothing guarded these slider configs, and a slider naming a stream the instrument does not declare is dropped as unmapped with no error anywhere — the same silent failure as a missing config. The new test requires every slider stream to be declared and every gated device to be drivable. It also found all three sliders in the dummy config naming nonexistent streams, i.e. the whole file was dead; dummy has no gated devices, so it is removed rather than rewired.

## Event routing

`event_id` is assumed to lie within the `detector_number` range, as the fake detector source and corrected real data do. The reference file still numbers `event_id` against an older 245760-pixel assumption for bank A, so bank B's ids fall entirely outside its range; that is a producer-side bug to fix upstream and is intentionally not worked around here, so replaying that file leaves the polarization bank empty.

## Known gaps

`ProjectionType` in `workflows/detector_view_specs.py` still lists only `xy_plane` and `cylinder_mantle_z`. MAGIC registers its spec directly rather than through `register_detector_view_spec`, so nothing is broken, but the convenience helper cannot yet express a cylinder-Y projection. Left for whenever a second cylinder-Y instrument needs it.

The `fake_logdata` service publishes a single hardcoded stream name that matches no instrument's dev mapping, so it is dead for every instrument, not just MAGIC. The dashboard's log producer widget covers the dev need here; deciding whether to fix or drop the service is out of scope.

## Test plan

- [x] Unit tests for the OFF-to-offset geometry derivation, the logical views, and the projection roundtrip.
- [x] Both banks driven end-to-end through the backend job path with one event per pixel: every event lands on screen, 0.1% and 0.0% empty screen bins.
- [x] Chain-patch bindings validated against the published geometry artifact.
- [x] Projection gating driven end-to-end: the job stays pending until the rotation substreams arrive, then activates.
- [x] Run the dashboard in dev mode, drive both rotation sliders, and confirm the projection follows the bank while the wire/strip views are unaffected.


SimonHeybrock:
LGTM, even though detector-numbering schemas are unclear. Expect future fixes to be necessary.

PR #979, opened 2026-06-18
Add stable CSS hooks to tool buttons for automation

SimonHeybrock:
## Motivation

Dashboard tool buttons (gear, play, stop, reset, remove, expand) render as label-less icons inside per-widget shadow DOM, exposing no text, `title`, or `aria-label`. Browser automation and screenshot tooling were therefore left clicking by pixel coordinates, which silently breaks on any layout change. This makes it hard for me (and other automation) to reliably drive and screenshot the UI.

## Change

`create_tool_button()` now tags every button with `lt-tool` and `lt-tool-{icon_name}`, and accepts a `css_classes` argument for caller context. Workflow rows add `lt-wf-{workflow_id.name}`, so each control is uniquely addressable, e.g. `.lt-wf-monitor_histogram.lt-tool-settings`. Playwright pierces shadow DOM, so these class locators resolve directly.

The classes have no associated style rules and are visually inert. They are intended as a committed convention; a helper test guards against silently dropping them, and the convention plus the no-Kafka launch recipe (`--transport none`) are documented for contributors and automation.

Per-widget context classes are currently threaded only through the workflow rows. Other repeated-icon views (e.g. plot grid controls) keep the universal type class and can gain context classes when a concrete automation flow needs them.

## Test plan

- [x] `buttons_test.py` asserts the type and caller classes are present
- [x] Existing workflow/button widget tests pass
- [x] Verified with Playwright against a running dashboard: targeting `.lt-wf-monitor_histogram.lt-tool-settings` by class alone (no coordinates) yields a unique match and opens the correct config modal; UI renders identically (classes visually inert)

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -43,6 +43,25 @@ def refresh(self):
 
 Do not use Unicode characters for button icons. Use embedded SVG icons from `dashboard/widgets/icons.py` via `get_icon()`. Use the `create_tool_button()` helper from `dashboard/widgets/buttons.py` for consistent styling.
 
+## Stable CSS hooks for automation
+
+Tool buttons render as label-less icons inside per-widget shadow DOM, so they carry
+no text, `title`, or `aria-label` — leaving nothing semantic for browser automation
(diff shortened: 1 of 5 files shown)


SimonHeybrock:
LGTM and no issues in UI observed

PR #980, opened 2026-06-18
Fix fake_monitors and fake_detectors int timestamp in da00 mode

SimonHeybrock:
`FakeMonitorSource` and `FakeDetectorSource` both passed a plain `int` (from `time.time_ns()`) as `Message.timestamp`. `Da00Serializer` calls `message.timestamp.to_ns()`, which fails with `AttributeError: 'int' object has no attribute 'to_ns'`. The ev44/ad00 serializers in these files also need `.to_ns()` since they now receive a `Timestamp` instead of an `int`.

Fix: wrap timestamps with `Timestamp.from_ns()` at the call sites and call `.to_ns()` when passing to streaming-data-types serializers. `fake_logdata.py` already did this correctly.

Note that this only affected local development, this change is irrelevant for production/users.

## Test plan
- [x] Run `fake_monitors` in da00 mode and confirm no serialization errors in the log

### DIFF
--- a/src/ess/livedata/services/fake_detectors.py
+++ b/src/ess/livedata/services/fake_detectors.py
@@ -18,6 +18,7 @@ from ess.livedata.config.config_loader import load_config
 from ess.livedata.config.instruments import get_config
 from ess.livedata.config.streams import stream_kind_to_topic
 from ess.livedata.core import IdentityProcessor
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
LGTM

SimonHeybrock:
Test fails due to DST webserver issues are unrelated

PR #981, opened 2026-06-18
Fix announcement fetch blocking session creation and IOLoop

SimonHeybrock:
The initial announcement fetch was called synchronously during Panel session creation (`_create_announcements_pane`), with a 10-second `urlopen` timeout. This blocked every new browser session for up to 10 s when the URL was unreachable. The periodic `add_periodic_callback` had the same problem — `refresh` ran on the Tornado IOLoop thread, stalling all sessions for up to 10 s every 5 minutes.

Both the initial fetch and the periodic refresh now spawn daemon threads, so session creation and the IOLoop are never blocked by a slow or unreachable announcements URL.

## Test plan
- [x] Open a fresh dashboard session with the announcements URL unreachable (or firewall blocked) — session should load immediately, then show the error text in the sidebar once the background fetch times out
- [x] Confirm the periodic refresh (5 min) also no longer causes visible lag

### DIFF
--- a/src/ess/livedata/dashboard/reduction.py
+++ b/src/ess/livedata/dashboard/reduction.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 import argparse
 import os
+import threading
 import urllib.request
 from urllib.error import URLError
 from urllib.request import urlopen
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
LGTM

PR #984, opened 2026-06-18
Fix per-session Kafka producer leak in log-producer widget

SimonHeybrock:
## Problem

Streaming fake chopper signals via the dev-mode log-producer widget makes dashboard actions progressively slower after a few minutes / many reloads.

The dev-only `LogProducerWidget` is created **per browser session**, but entered its `KafkaSink` into the **application-level** exit stack. The producer — with its background librdkafka thread and broker connections — was therefore never released until process shutdown. Every session and every dev autoreload leaked a producer, accumulating background threads and connections that drag the whole process down over time. The widget's streaming-readback periodic callbacks were likewise untracked, with no way to stop them, and each published value logged at INFO (~12 lines/s/session).

This leak is specific to this widget: the dashboard's command/ROI sinks and consumer are app-level singletons created once, so their lifetimes are already correct.

## Fix

The widget now owns a session-scoped exit stack for its sink and exposes `close()`. Readback callbacks are tracked and stopped before the sink closes, so none fire against a closed producer. Per-publish logging is dropped to debug.

Cleanup is routed through the **heartbeat-based** session teardown rather than `pn.state.on_session_destroyed` alone, which is unreliable (browser crashes, dropped connections) — the very reason the heartbeat stale-session reaper exists. `SessionUpdater` gains a cleanup-handler hook invoked from `cleanup()`, the single teardown chokepoint reached both by `on_session_destroyed` (via `SessionRegistry.unregister`) and by the stale reaper (`cleanup_stale_sessions`). The widget registers its `close()` there, so the producer is released on clean disconnect and, failing that, when the session goes stale. `create_sidebar_content` now receives the session updater to allow registering the handler.

The per-call `flush()` in `KafkaSink.publish_messages` is deliberately left untouched: it is the delivery contract shared by the command/ROI sinks and the backend data pipeline, and with the leak fixed its cost in a single session is negligible.

## Test plan

- [x] Unit tests cover the cleanup-handler hook, including invocation via the stale-session reaper path.
- [x] Full dashboard test suite passes (1531 tests).
- [x] Run the dev dashboard, stream chopper signals for several minutes, and confirm dashboard actions stay responsive.
- [x] Reload the dashboard many times and confirm Kafka producers do not accumulate (clean disconnect and, by leaving a tab to go stale, the heartbeat path).

### DIFF
--- a/src/ess/livedata/dashboard/dashboard.py
+++ b/src/ess/livedata/dashboard/dashboard.py
@@ -92,8 +92,18 @@ class DashboardBase(ServiceBase, ABC):
             raise ValueError(f"Unknown transport type: {transport}")
 
     @abstractmethod
-    def create_sidebar_content(self) -> pn.viewable.Viewable:
-        """Override this method to create the sidebar content."""
+    def create_sidebar_content(
+        self, session_updater: SessionUpdater
+    ) -> pn.viewable.Viewable:
+        """Override this method to create the sidebar content.
+
+        Parameters
+        ----------
+        session_updater:
+            The session updater for this browser session. Use
+            ``register_cleanup_handler`` to release per-session resources on
+            session teardown.
+        """
         pass
 
     @abstractmethod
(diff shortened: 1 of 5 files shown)


SimonHeybrock:
LGTM

PR #985, opened 2026-06-18
Add fake backend transport and seeded UI config for Kafka-free dashboard testing

SimonHeybrock:
## Motivation

We could launch the dashboard UI without Kafka (`--transport none`, see #979), but had no way to push data into it, so plots could never be verified or screenshotted. Setting up a representative view also meant clicking through source selection and plot-grid configuration every time.

## Change

Adds an in-process **fake backend** transport and a way to seed the UI from committed config, so the whole flow runs without Kafka and without repetitive clicking.

**`--transport fake`** runs an in-process backend that mimics a real worker: it captures the `WorkflowConfig` command the dashboard sends (which carries the runtime `job_number`), acknowledges it, reports the job ACTIVE, and emits periodic result data. Because data is synthesized from each workflow's output templates (the `default_factory` DataArrays on `WorkflowSpec.outputs`) — the same templates the dashboard uses for plotter selection — generated data matches what the plotter expects, so plots render correctly. Driven entirely by the normal "start workflow" UI flow. Timeseries/scalar outputs emit a fresh `time`-stamped scalar each update, which the dashboard accumulates into a series.

This supersedes the old HTTP-injection idea (#602): routing through the command path means the fake backend learns the runtime `job_number` itself, rather than an external injector having to guess it.

**`--config-dir`** points the persistent UI config store at an arbitrary base directory (a CLI equivalent of the existing `LIVEDATA_CONFIG_DIR`). Combined with the committed dummy fixture (all workflows staged with default sources + a detector plot grid), launching gives a pre-configured dashboard where starting a workflow immediately renders live plots. The dashboard writes to this directory, so callers copy the fixture to a scratch dir first; the recipe is documented in `CLAUDE.md`. A test guards the fixtures against workflow/output schema drift.

The automation rule gains a note on when workflow rows rebuild (state-version changes only) versus in-place status refresh, and the cold-start / multi-step windows where a browser click can race a rebuild.

## Test plan

- [x] Unit tests for data synthesis and the fake backend's command/response/data flow
- [x] Fixture drift-guard test (workflow ids, sources, output views resolve against the registry)
- [x] Verified live: launched `--transport fake --config-dir <fixture copy>`, started workflows via Playwright, confirmed detector (2-D image) and other plots render with synthesized data

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -62,6 +62,21 @@ Treat them as a stable contract: do not drop them in refactors (a test in
 the same icon (e.g. per-grid/per-cell controls in `plot_grid_manager.py`), pass a
 context `css_classes` entry so each instance is uniquely addressable.
 
+### Driving workflow config flows
+
+A `WorkflowStatusWidget` rebuilds its row (`_build_widget`) only when that workflow's
+state *version* changes — staging, commit, or stop (`job_orchestrator.py`). Steady-state
+status refresh just reassigns badge/dots/timing HTML in place, so it does *not* detach
+elements. Two windows still detach the element under the cursor:
+- **Cold start**: the first refresh tick after page load can fire one rebuild while
+  Bokeh models are still settling. Wait a few seconds after load before the first click.
+- **Multi-step flows**: each stage/commit rebuilds *that* row, so a click landing on a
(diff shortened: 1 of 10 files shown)


SimonHeybrock:
LGTM and works also for dev use without Kafka

PR #986, opened 2026-06-18
Wire DREAM choppers into wavelength-LUT workflow with per-chopper band diagnostic

SimonHeybrock:
## Summary

Wires DREAM's chopper cascade into the wavelength lookup-table workflow and adds a per-component wavelength-band diagnostic to make it easy to see, at a glance, which chopper in a tight cascade is blocking the beam.

## What changed

- DREAM's five choppers (pulse-shaping 1/2, band, overlap, T0) are declared on the instrument and flow into the wavelength-LUT workflow. Pulse-shaping chopper 2 defaults to counter-rotation, and initial chopper settings are tuned so neutrons make it through the full cascade.
- New `chopper_cascade_bands` workflow output: the surviving wavelength band evaluated at the *exact* distance of the source and each chopper — no rasterization onto the table's distance grid, so closely-spaced choppers stay individually resolved. An all-NaN row marks where the beam is blocked.
- A `cut_distances` workflow parameter adds extra rows to that diagnostic at scientist-chosen beamline points (typically monitor and detector positions), propagating the cascade forward to each. Entered as a comma-separated list, since the parameter widget has no native list input; the parsing is shared with the existing static line-coordinate input.
- `Overlay1DPlotter` colors curves by position rather than by coordinate value when coordinates are non-integer (e.g. distances in metres), so nearby distances no longer collapse onto the same color.
- A "Choppers → Wavelength" grid template, a dev log-producer config for the DREAM choppers, and a new no-shape DREAM geometry artifact in the registry.

## Examples

New workflow config options:
image

New workflow output, showing disabled curve as well as two custom distances (70m and 200m, the latter with large wavelength spread, indicating frame overlap at this unrealistic distance):
screenshot-2026-06-18T14:40:13

The band diagnostic reuses essreduce's private `_estimate_wavelength_by_polygon_centers`; the intent is to promote this upstream into `ess.reduce.unwrap` once the design proves out.

## Test plan

- [x] Drive the "Choppers → Wavelength" grid in the dashboard, configure a few cut distances, and confirm the band overlay distinguishes individual choppers and shows blocked rows vanishing.

### DIFF
--- a/configs/log_producer_dream.json
+++ b/configs/log_producer_dream.json
@@ -4,14 +4,14 @@
             "stream_name": "pulse_shaping_chopper1/rotation_speed_setpoint",
             "label": "Pulse-shaping chopper 1 speed setpoint (Hz)",
             "min": 7.0,
-            "max": 14.0,
+            "max": 210.0,
             "step": 7.0,
             "initial": 14.0
         },
         {
             "stream_name": "pulse_shaping_chopper2/rotation_speed_setpoint",
             "label": "Pulse-shaping chopper 2 speed setpoint (Hz, counter-rotating)",
-            "min": -14.0,
+            "min": -196.0,
             "max": -7.0,
             "step": 7.0,
             "initial": -14.0
--- a/configs/log_producer_dream.json
+++ b/configs/log_producer_dream.json
@@ -1,0 +1,94 @@
+{
+    "sliders": [
+        {
+            "stream_name": "pulse_shaping_chopper1/rotation_speed_setpoint",
+            "label": "Pulse-shaping chopper 1 speed setpoint (Hz)",
+            "min": 7.0,
+            "max": 14.0,
+            "step": 7.0,
+            "initial": 14.0
+        },
+        {
+            "stream_name": "pulse_shaping_chopper2/rotation_speed_setpoint",
+            "label": "Pulse-shaping chopper 2 speed setpoint (Hz)",
+            "min": 7.0,
+            "max": 14.0,
+            "step": 7.0,
+            "initial": 14.0
+        },
+        {
+            "stream_name": "overlap_chopper/rotation_speed_setpoint",
+            "label": "Overlap chopper speed setpoint (Hz)",
+            "min": 7.0,
+            "max": 14.0,
+            "step": 7.0,
+            "initial": 14.0
+        },
+        {
+            "stream_name": "band_chopper/rotation_speed_setpoint",
+            "label": "Band-control chopper speed setpoint (Hz)",
+            "min": 14.0,
+            "max": 112.0,
+            "step": 14.0,
+            "initial": 112.0
+        },
+        {
+            "stream_name": "T0_chopper/rotation_speed_setpoint",
+            "label": "T0 chopper speed setpoint (Hz)",
+            "min": 14.0,
+            "max": 28.0,
+            "step": 14.0,
+            "initial": 28.0
+        },
+        {
+            "stream_name": "pulse_shaping_chopper1/delay",
+            "label": "Pulse-shaping chopper 1 delay readback (ns)",
+            "min": 0.0,
+            "max": 71428571.0,
+            "step": 100000.0,
+            "initial": 21000000.0,
+            "noise_stddev": 100.0,
+            "publish_rate_hz": 5.0
+        },
+        {
+            "stream_name": "pulse_shaping_chopper2/delay",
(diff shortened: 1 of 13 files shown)


configs/log_producer_dream.json
```diff
@@ -0,0 +1,94 @@
+{
+    "sliders": [
+        {
+            "stream_name": "pulse_shaping_chopper1/rotation_speed_setpoint",
+            "label": "Pulse-shaping chopper 1 speed setpoint (Hz)",
+            "min": 7.0,
+            "max": 14.0,
```
nvaytet: Note that the numbers for Dream's 'high-resolution' mode go quite a bit higher than 14 Hz.
I have these numbers (from the `tof` package):

 ```
dream_high_resolution["PSC1"]["frequency"]["value"] = 15 * 14.0
dream_high_resolution["PSC1"]["phase"]["value"] = 25 - 180

dream_high_resolution["PSC2"]["frequency"]["value"] = 14 * 14.0
dream_high_resolution["PSC2"]["phase"]["value"] = 100.5

dream_high_resolution["OC"]["frequency"]["value"] = 14.0
dream_high_resolution["OC"]["phase"]["value"] = 297.0 - 180.0 - 90.0

dream_high_resolution["BC"]["frequency"]["value"] = 112.0
dream_high_resolution["BC"]["phase"]["value"] = 200.0 - 180.0

dream_high_resolution["T0"]["frequency"]["value"] = 28.0
dream_high_resolution["T0"]["phase"]["value"] = 270.0 - 180.0
```

File: src/ess/livedata/handlers/wavelength_lut_workflow.py [hunk 2]
```diff
@@ -204,14 +281,20 @@ def _build_pipeline(params: WavelengthLutParams) -> sciline.Pipeline:
     )
     wf[ParamsKey] = params
     wf.insert(_attach_provenance)
+    # Per-component diagnostic evaluated at exact chopper distances, sidestepping
```
nvaytet, line 284: Can I ask what was the original motivation for adding this in livedata? Was it because you were visualizing the LUT but nothing was coming through the chopper system and you couldn't figure out where it was going wrong (because the LUT distance resolution was poorer than the separation between the PSC choppers)?

An alternative could have been to have checkboxes or toggle buttons for each chopper to control whether we want to include them or not, and just visualize the LUT. Adding them in one by one would quickly reveal which is the offending one.

I'm asking because we discussed how maybe this figure is not always so useful as it does not show when choppers would be open (if their phase is off) if no part of the pulse goes through the opening.
So we are still sort of 'blind'. As we also discussed, we could add the open times to the graph, etc.

But I'm wondering if just checkboxes could be a cheap and useful alternative?

tests/config/chopper_wavelength_lut_test.py
```diff
@@ -0,0 +1,119 @@
...
+    params = WavelengthLutParams(
+        cascade_bands=CascadeBands(distances='7.0, 12.0')
+    ).model_dump()
+    job = _create_lut_job(instrument, params=params)
+    aux = {}
+    for chopper in instrument.choppers:
+        speed_unit = instrument.streams[speed_setpoint_stream(chopper)].units
+        delay_unit = instrument.streams[delay_readback_stream(chopper)].units
+        aux[speed_setpoint_stream(chopper)] = _nxlog(14.0, speed_unit)
+        aux[delay_setpoint_stream(chopper)] = _nxlog(0.0, delay_unit)
```
nvaytet: Both `delay_readback_stream` and `delay_setpoint_stream` are used here. Is that intentional?

src/ess/livedata/handlers/wavelength_lut_workflow.py
SimonHeybrock: > Was it because you were visualizing the LUT but nothing was coming through the chopper system and you couldn't figure out where it was going wrong (because the LUT distance resolution was poorer than the separation between the PSC choppers)?

Yes, that. The 2D plots have limited utility and would require very high resolution (or switching back and forth been limiting params to a local range and a global plot).

> I'm asking because we discussed how maybe this figure is not always so useful as it does not show when choppers would be open (if their phase is off) if no part of the pulse goes through the opening.
So we are still sort of 'blind'. As we also discussed, we could add the open times to the graph, etc.

I have another branch that adds a workflow producing acceptance diagrams. I consider the two independent, but I do not want to guess whether one of them is more useful to scientists, or whether they want both.

nvaytet (approved):

PR #988, opened 2026-06-19
Playwright driving kit for the dashboard, plus automation-hook and modal fixes

SimonHeybrock:
## Motivation

Driving the live dashboard with Playwright (for verification and screenshots) meant re-discovering the layout and re-writing navigation boilerplate every session, with no clean way to recover when a prior instance held the hardcoded port. This makes that flow turnkey and removes the friction the previous fake-backend work (#985) left behind.

## What changed

A reusable Playwright driving kit, `scripts/drive_dashboard.py`, exposed both as a `Dashboard` library class (readiness wait, tab navigation, retry-on-detach click, modal open, screenshot, runtime UI inventory) and a CLI. `--map` dumps the live tabs and stable `lt-*` hooks so a session starts informed rather than blind; `--launch` runs the whole Kafka-free fake-backend lifecycle (fixture copy, readiness, teardown) in one command, defaulting off port 5009 so automation never collides with interactive dev.

The dashboard gains a `--port` flag for running several instances or sidestepping a held port.

Two rounds of agent-driven dogfooding then surfaced real gaps, which are fixed here:

- **Stable hooks completed.** Per-panel tool buttons that were hand-rolled (the layer-details toggle, the multi-layer settings dropdown) and the per-grid row controls carried no `lt-*` automation hook, unlike the buttons built via `create_tool_button`. They are now tagged and guarded by tests.
- **Escape closes modals.** Panel's modal (a11y-dialog) binds Escape on `document.body`, but the dialog renders in Bokeh's shadow DOM and Panel focuses the dialog container on open, so Escape did nothing until focus moved into the content — affecting real users, not just automation. An invisible per-session `ModalEscapeCloser` installs a document-level Escape handler covering every modal. The cleaner long-term fix is upstream in Panel; this is a local shim.

Finally, the docs were brought in line: the automation rule documents the shadow-DOM selector gotcha, the tab/modal inventory, and a "keeping automation working" checklist mapping each kind of UI change to what must be kept in sync; its path scope now also covers `reduction.py` and the driving kit. The verbose driving and fixture recipes were moved out of the always-loaded `CLAUDE.md` into that path-scoped rule, trimming `CLAUDE.md` by a third with no loss of information.

## Test plan

- [x] Unit tests for the new automation hooks (per-panel buttons, per-grid rows)
- [x] Existing dashboard widget suite passes
- [x] Verified live with the kit: `--launch --map` inventories tabs/hooks, `--screenshot` renders detector plots, and Escape closes the settings, edit, and config modals from initial focus

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -1,5 +1,5 @@
 ---
-paths: src/ess/livedata/dashboard/widgets/**/*.py
+paths: src/ess/livedata/dashboard/widgets/**/*.py, src/ess/livedata/dashboard/reduction.py, scripts/drive_dashboard.py
 ---
 
 # Dashboard Widget Patterns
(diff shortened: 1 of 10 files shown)


SimonHeybrock:
Confirmed that ESC did not work before this but works now.

Issue #989, opened 2026-06-19
Source position keyed off NXsource ignores NXmoderator (origin-pinned accelerator trap)


SimonHeybrock:
## Problem

BIFROST's geometry models the real neutron source as an `NXmoderator` (~162 m upstream) and reserves `NXsource` for accelerator/production metadata that is **explicitly pinned at the origin** (`depends_on='.'`, empty transformations) — coinciding with the sample. This matches essreduce's own `ProductionInfo` docstring ("its position is not relevant for data reduction").

essreduce, however, keys the source position off the *unique* `NXsource` component. For BIFROST that resolves to the origin-pinned accelerator group, not the moderator. Any reduction path that consumes `Position[NXsource, RunType]` therefore silently measures distances from the origin (i.e. the sample) for BIFROST.

## Where it currently bites / is worked around

The chopper-cascade wavelength-LUT workflow hit this as a hard failure (chopper distance came out negative, since the choppers sit "behind" an origin source). It is worked around in `wavelength_lut_workflow._moderator_source_position`: resolve the source from `NXmoderator` when present, else fall back to `NXsource`.

That fix is **local to the LUT workflow only**. The same trap is latent anywhere else source position matters — most notably a detector `Ltotal` computation. BIFROST's detector reduction is currently TOA-based, so it does not exercise the source position today, but the moment it (or any other instrument adopting the `NXmoderator` convention) uses an essreduce TOF/`Ltotal` path, it will silently compute from the sample with no error.

## Possible directions

- Centralise the moderator-preferring source resolution so every workflow shares it, rather than the one-off in the LUT workflow.
- Or push it upstream into essreduce: prefer `NXmoderator` for the source position when present, treating `NXsource` as production metadata.

Whichever way, the goal is convention-agnostic behaviour: works whether the source is labelled `NXsource` (e.g. LOKI) or `NXmoderator` (BIFROST), since the convention is not expected to be uniform across instruments.

Context: surfaced while reviving the BIFROST chopper-cascade work (`bifrost-choppers`). Related to the geometry-file issues tracked in #962.

SimonHeybrock:
Swept all nine instruments in the 2026-09-07 coda drop. The trap is broader than the moderator-vs-source split this issue describes: it now also fires `NXsource` against `NXsource`, in two distinct shapes.

[...]

PR #990, opened 2026-06-19
Wire BIFROST chopper-cascade wavelength-LUT workflow

SimonHeybrock:
The substantive part here is the neutron-source handling. BIFROST models the real source as an `NXmoderator` ~162 m upstream and reserves `NXsource` for accelerator metadata pinned at the origin. essreduce keys the chopper-cascade reference off the unique `NXsource`, so it would measure flight distance from the accelerator and reject the upstream choppers as lying behind the source. Two fixes address this:
1. `make_geometry_nexus` now carries the `NXmoderator` placement chain into the artifact (it was silently dropping it)
2. the wavelength-LUT workflow resolves the source position from the `NXmoderator` when present, falling back to `NXsource` otherwise. The behaviour is convention-agnostic, so LOKI (source labelled `NXsource`) is unaffected.

This workflow-local fix leaves a latent trap for any other essreduce path that keys source position off `NXsource` for BIFROST (e.g. a detector `Ltotal` if the reduction ever leaves TOA mode); tracked in #989.

The geometry file's `detector_tank_angle` chain is still broken (missing `_t0` offset, stale group prefix), so detector geometry is pinned to the prior artifact while the choppers use the corrected one. The two collapse to a single file once a fixed source file is available (#962). The corrected `geometry-bifrost-2026-06-08.nxs` artifact has been re-uploaded to the `geometry-v0` release and the registry hash updated.

### DIFF
--- a/configs/log_producer_bifrost.json
+++ b/configs/log_producer_bifrost.json
@@ -21,6 +21,114 @@
             "target_stream": "rotation_stage/target_value",
             "idle_stream": "rotation_stage/idle_flag",
             "ramp_seconds": 2.0
+        },
+        {
+            "stream_name": "pulse_shaping_chopper_1/rotation_speed_setpoint",
+            "label": "PSC 1 speed setpoint (Hz)",
+            "min": 14.0,
+            "max": 210.0,
+            "step": 14.0,
+            "initial": 14.0
+        },
+        {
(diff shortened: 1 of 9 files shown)


SimonHeybrock:
LGTM

Issue #884, opened 2026-04-20
Unify spectrum view into generic detector view workflow


SimonHeybrock:
# Spectrum View Unification

## Problem

Bifrost and ESTIA each maintain a bespoke `spectrum_view` workflow alongside the generic detector view. Both compute essentially the same quantity the generic detector view already computes internally — a histogram with spatial and time-of-arrival dims — but then emit it as a separate output instead of summing it away. The current state:

- **Bifrost** (`src/ess/livedata/config/instruments/bifrost/`): `specs.SpectrumViewParams`, `specs.SpectrumViewOutputs`, `factories._make_spectrum_view` (+ three `NewType`s and factory wiring). Runs a separate `TofWorkflow` pipeline.
- **ESTIA** (`src/ess/livedata/config/instruments/estia/`): `specs.EstiaSpectrumViewParams`, `specs.SpectrumViewOutputs`, `factories._make_spectrum_view`. Runs a separate `EstiaWorkflow` pipeline.

Two pipelines per instrument, two specs, two output schemas, redundant param definitions (`time_bins` vs `toa_edges` both expressing TOA binning).

The generic detector view already computes `AccumulatedHistogram[Cumulative]` (see `detector_view/providers.py:183`) with shape `(spatial_dims × spectral)`. `detector_image` then sums the spectral dim away. Exposing the pre-sum histogram as an additional output is a natural extension — the data is already there.

### Hook point

Add a new optional output derived from `AccumulatedHistogram[Cumulative]` via a per-instrument transform. The transform can sum over spatial dims not wanted in the spectrum and/or rebin along kept spatial dims.

```
project → ScreenBinnedEvents        (spatial_dims × spectral events)
  ↓
compute_detector_histogram → DetectorHistogram         (spatial_dims × spectral)
  ↓
accumulated_histogram → AccumulatedHistogram[Cumulative|Current]
  ↓                                           ↓
detector_image (sums spectral)       spectrum_view (sums unwanted spatial, rebins)
  ↓                                           ↓
DetectorImage                         SpectrumView  ← new output
```

### Two config levels

1. **Spec-time (per-instrument)** — the transform itself: how to reshape `AccumulatedHistogram` into the desired spectrum output shape (fold/sum unwanted spatial dims, ordering). Declared at registration.
[...]

PR #996, opened 2026-06-22
Fix Bifrost unified detector view layout to match static grid

SimonHeybrock:
## Problem

The Bifrost unified detector view rendered as 5 rows by 2700 columns, so the data overflowed the grid horizontally and showed only the 5 arcs as rows. The static triplet-bounds grid, overlaid as a second plot layer, is drawn for a 15×900 image (5 arcs × 3 tubes stacked as rows, 9 channels × 100 pixels as columns), so it no longer lined up with the data.

## Cause

The shared 2D logical view flattened `(tube, channel, pixel)` into a single 2700-wide axis while preserving only `arc`. This regressed in the spectrum-view unification (#884), which adopted that layout to make the per-arc spectrum fold trivial but left the image incompatible with the grid. The detector geometry and stream indexing were unchanged.

## Fix

Restore the `(arc/tube, channel/pixel)` = 15×900 layout for the image, and unfold the shared cumulative histogram from that layout in the spectrum transform. The spectrum view still produces `(arc, detector_number, toa)` with the same per-tube pixel grouping, and counts are conserved.

## Test plan

- [x] Image is 15×900 and the spectrum is `(arc, detector_number, toa)` across all `pixels_per_tube` values; counts conserved (unit checks)
- [x] Existing Bifrost, logical-view, detector-view-spec, and detector-data tests pass
- [x] Grid overlay aligns with the data in the live Bifrost unified detector view

### DIFF
--- a/src/ess/livedata/config/instruments/bifrost/specs.py
+++ b/src/ess/livedata/config/instruments/bifrost/specs.py
@@ -221,15 +221,18 @@ monitor_handle = register_monitor_workflow_specs(
 
 
 def _logical_view(obj: sc.Variable | sc.DataArray, source_name: str) -> sc.DataArray:
-    """Reshape raw detector data into ``(arc, detector_number_full)``.
+    """Reshape raw detector data into the 2D ``(arc/tube, channel/pixel)`` image.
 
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
LGTM

PR #997, opened 2026-06-22
Stamp ROI requests with epoch-0 timestamp to avoid event-time batcher delay

SimonHeybrock:
## Summary

ROI requests were stamped with the dashboard's wall-clock time and routed through the event-time message batcher, which orders messages by data event-time. When the live detector stream's event-time trails wall-clock — as on the deployed LoKI `detector_data` service, where ~10 s of DAQ pipeline lag was observed — the request was parked in the batcher's future buffer until the data watermark caught up. This delayed the ROI readback (and made live plots look stale) by roughly that lag, independent of load and unaffected by restarts.

ROI requests carry no meaningful event-time: the selection applies to data accumulated since run start, and `LatestValueHandler` discards the message timestamp. Stamping the request at epoch 0 makes the batcher treat it as already-current and apply it at the next batch close. Result timestamps are unaffected — they derive from the data batch's event-time window, not the request. Confirmed batcher-agnostic: an epoch-0 request lands in the current / non-gated bucket and is delivered at the next close under both `SimpleMessageBatcher` and the optional `RateAwareMessageBatcher`, never the future-hold path.

This is the dashboard-side fix; the `ns=0` value applied as a live diagnostic on the deployed dashboard already exhibits the corrected behaviour.

## Test plan

- [x] `roi_publisher` unit tests, including a new regression test asserting the epoch-0 stamp
- [x] Manual: draw/move an ROI on a live (lagging) detector view and confirm the readback updates within ~1–2 s rather than ~10 s

### DIFF
--- a/src/ess/livedata/dashboard/roi_publisher.py
+++ b/src/ess/livedata/dashboard/roi_publisher.py
@@ -8,6 +8,7 @@ from ..config.models import PolygonROI, RectangleROI
 from ..config.roi_names import ROIGeometry
 from ..config.workflow_spec import JobId
 from ..core.message import Message, MessageSink, StreamId, StreamKind
+from ..core.timestamp import Timestamp
 
 logger = structlog.get_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
Tried and verified at LOKI in production deployment.

PR #998, opened 2026-06-22
Add since_start output stream for monitor totals

### LINKED ISSUE #991: `since_start` window mode not working for monitors
Observed for Bifrost monitors.

SimonHeybrock:
## Why

After unifying workflow output streams behind the dashboard's window-mode selection, the `since_start` mode worked for the Lines plotter (histogram view) but the Bars plotter of the total counts always showed the latest per-update value. The monitor scalar Total / Total-in-range views were missed during the unification: they declared only a `per_update` stream, so selecting "since run start" silently fell back to that same stream via `OutputView.field_for` and changed nothing. Detectors already exposed cumulative scalar streams; monitors did not.

Fixes #991.

## What

Produce cumulative scalar totals in the monitor workflow, mirroring the detector. The two scalar views now declare both stream roles, so `since_start` mode subscribes to the run-cumulative stream while `window`/latest stays on the per-update stream. To avoid duplicating the sum logic, the monitor adopts the detector's generic `AccumulationMode` pattern (one provider serves both the cumulative and window views, instantiated per mode by sciline); the shared `Current` / `Cumulative` / `AccumulationMode` markers move to a small `accumulation_mode` module both workflows depend on.

Separately, this fixes a latent footgun: selecting "since run start" on a view that has no cumulative stream was a silent no-op (`OutputView.field_for` falls back to the per-update stream). Rather than hide controls — the window duration aggregates the per-update stream and stays useful regardless of whether a cumulative stream exists — the modal now rejects the unsupported selection at config time with a clear error.

### DIFF
--- a/src/ess/livedata/dashboard/plotting_controller.py
+++ b/src/ess/livedata/dashboard/plotting_controller.py
@@ -254,16 +254,35 @@ class PlottingController:
 
 
 def output_view_supports_windowing(workflow_spec: WorkflowSpec, view_name: str) -> bool:
-    """Return whether a window-mode choice (latest/window) is meaningful.
+    """Return whether the window controls (mode, duration, aggregation) apply.
 
-    Windowing is meaningful when the view exposes a ``per_update`` stream.
-    Views that are cumulative-only (e.g. ``I(Q)``) lock the mode to
-    ``since_start``.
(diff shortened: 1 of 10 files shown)


SimonHeybrock:
LGTM

PR #999, opened 2026-06-23
Drop al00/ep01 status schemas on forwarder log topics

SimonHeybrock:
## Motivation

EPICS forwarder log topics carry `f144` numeric data interleaved with `al00` (alarm) and `ep01` (EPICS connection-status) messages for the same PVs — this is the forwarder's intended convention, not a misconfiguration. We consume only `f144`, so every `al00`/`ep01` message reached the `f144` deserializer and raised `WrongSchemaException`. On LOKI this was a continuous flood (~4–5/s, roughly balanced between the two schemas) and it affects every service that subscribes to log topics: detector, timeseries, and data_reduction.

These status schemas are *validity metadata* orthogonal to the data we reduce; there is nothing useful to do with them in live processing (alarms surface in NICOS, and dropping data while any device alarms is overkill). So they are dropped, not consumed.

## Change

Route log topics by schema: `f144` to the existing adapter chain, `al00`/`ep01` to a `NullAdapter` that drops them silently via a dedicated `IgnoredMessageError`. The single change in the route builder covers all three consuming services.

While here, the unknown-schema warning was a blind spot: it named neither the topic nor the offending schema. Genuinely-unexpected schemas now warn with both.

## Test plan

- [x] Unit tests: status schemas dropped silently (no log, no error count); unknown schema warns with topic + schema; builder-level log route drops `al00`/`ep01`.
- [x] Full `tests/kafka/` suite green.

### DIFF
--- a/src/ess/livedata/kafka/routes.py
+++ b/src/ess/livedata/kafka/routes.py
@@ -18,6 +18,7 @@ from .message_adapter import (
     KafkaToF144Adapter,
     KafkaToMonitorEventsAdapter,
     MessageAdapter,
+    NullAdapter,
     ResponsesAdapter,
     RouteBySchemaAdapter,
     RouteByTopicAdapter,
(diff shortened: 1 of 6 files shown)


src/ess/livedata/kafka/routes.py
```diff
@@ -100,13 +101,24 @@ def with_area_detector_route(self) -> Self:
...
+            routes={
+                'f144': ChainedAdapter(
+                    first=KafkaToF144Adapter(
+                        stream_lut=self._stream_mapping.logs,
+                        stream_counter=self._stream_counter,
+                    ),
+                    second=F144ToLogDataAdapter(),
+                ),
+                'al00': NullAdapter(),
+                'ep01': NullAdapter(),
```
nvaytet, line 120: And I guess it is better here to explicitly list all the schemas we want to ignore, rather than only listing `f144` and ignoring everything else via a DefaultDict or something like that. If we receive something unexpected, we want to know about it.

nvaytet (approved):

PR #1000, opened 2026-06-23
Log per-stream producer lag for upstream staleness detection

SimonHeybrock:
## Motivation

Debugging the Loki `detector_data` service showed an 8-10 s displayed lag while Kafka offset lag stayed near zero. The cause was upstream: ev44 payloads arrive promptly (Kafka CreateTime ~current) but carry a `reference_time` ~8.8 s in the past, i.e. the producer buffered the data before publishing. The existing `consumer_metrics` offset lag cannot reveal this, and diagnosing it required hacking a `print` into production code.

## What this adds

Per-stream **producer lag** = `kafka_create_time - payload_timestamp`, which isolates upstream producer staleness from this service's own consumption delay. It is tracked in `StreamCounter` keyed by `(topic, source, wire schema)` — the identifiers needed to report issues upstream, not the internal stream name — and the processor logs one `stream_lag` line per stream each metrics window:

```
stream_lag  topic=... source=... schema=ev44 min_lag_s=8.51 max_lag_s=8.85 count=412   [WARNING]
```

Severity reflects the lag, so issues surface by log level without flooding:

- **ERROR** — payload timestamped in the future (clock/logic fault), beyond a 100 ms inter-host clock-skew tolerance
- **WARNING** — more than 2 s stale at publish time
- **INFO** — baseline

Flat per-stream lines (stable `(topic, source, schema)` ordering) keep the output greppable through journald, including by severity.

## Design notes

Recording lives in the adapter layer, where the wire identifiers and the schema-specific payload timestamp coexist (each adapter already normalizes its own timestamp). The lag DTOs live in `core` so the `StreamStatsProvider` protocol return type respects the `kafka -> core` import direction. `StreamStats` and the status heartbeat are unchanged, so this is log-only today; surfacing lag in the worker-status UI later is a small follow-up (add a field to `StreamStat` and a column).

## Test plan

- [x] Unit tests: `StreamCounter` lag aggregation, ordering, reset, EPICS-suffix filtering, and per-stream severity bands incl. the skew tolerance
- [x] Adapter-level: real ev44 through `KafkaToEv44Adapter` records lag keyed by topic/source/schema; unavailable broker timestamp is skipped
- [x] Full fast suite passes

### DIFF
--- a/src/ess/livedata/core/job.py
+++ b/src/ess/livedata/core/job.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 from __future__ import annotations
 
+import logging
 import traceback
 from dataclasses import dataclass
 from enum import StrEnum, auto
@@ -128,12 +129,64 @@ class StreamStats:
     streams: tuple[StreamStat, ...]
 
 
+# Defaults chosen against realistic ESS conditions: inter-host NTP/chrony skew
+# stays in the single-digit-millisecond range and Kafka CreateTime has only
+# millisecond resolution, so 100 ms comfortably clears clock noise while a real
+# "payload from the future" anomaly (seconds) trips the ERROR band. The WARNING
+# band flags data already seconds stale at publish time.
+LAG_WARN_THRESHOLD_S = 2.0
+LAG_FUTURE_TOLERANCE_S = 0.1
+
+
+@dataclass(frozen=True, slots=True)
+class StreamLag:
+    """Lag statistics for one (topic, source, schema) over a time window.
+
+    Lag is ``kafka_create_time - payload_timestamp`` in seconds: how stale a
+    payload already was when the broker recorded it, isolating upstream producer
+    staleness from this service's own consumption delay.
+    """
+
+    topic: str
+    source: str
+    schema: str
+    min_s: float
+    max_s: float
+    count: int
+
+
+@dataclass(frozen=True, slots=True)
+class StreamLagReport:
+    """Per-stream lag for one window, ordered stably by (topic, source, schema)."""
+
+    streams: tuple[StreamLag, ...]
+    warn_threshold_s: float
+    future_tolerance_s: float
+
+    def level(self, lag: StreamLag) -> int:
+        """Python logging level reflecting a single stream's lag.
+
+        ERROR if the payload is meaningfully in the future (clock/logic fault),
+        WARNING if it is more than ``warn_threshold_s`` stale at publish time,
+        otherwise INFO.
+        """
+        if lag.min_s < -self.future_tolerance_s:
+            return logging.ERROR
+        if lag.max_s > self.warn_threshold_s:
+            return logging.WARNING
+        return logging.INFO
+
+
 @runtime_checkable
 class StreamStatsProvider(Protocol):
-    """Provides accumulated per-stream message counts."""
+    """Provides accumulated per-stream message counts and lag."""
 
     def drain(self, window_seconds: float) -> StreamStats:
-        """Return accumulated stats and reset counters."""
+        """Return accumulated counts and reset."""
+        ...
+
+    def drain_lag(self) -> StreamLagReport | None:
+        """Return accumulated per-stream lag and reset, or None if empty."""
         ...
 
 
--- a/src/ess/livedata/core/orchestrating_processor.py
+++ b/src/ess/livedata/core/orchestrating_processor.py
@@ -20,6 +20,7 @@ from .job import (
     JobStatus,
     ServiceState,
     ServiceStatus,
+    StreamLagReport,
     StreamStats,
     StreamStatsProvider,
 )
@@ -371,6 +372,7 @@ class OrchestratingProcessor(Generic[Tin, Tout]):
                 self._pending_stream_stats = self._stream_stats_provider.drain(
                     window_seconds
                 )
+                self._log_stream_lag(self._stream_stats_provider.drain_lag())
 
             logger.info(
                 'processor_metrics',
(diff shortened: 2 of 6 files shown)


jl-wynen:
Is this for us to debug? This sounds like you are diagnosing issues with ECDC software. Don't they have debugging tools for this?

SimonHeybrock:
> Is this for us to debug? This sounds like you are diagnosing issues with ECDC software. Don't they have debugging tools for this?

Maybe they do, but adding this avoids us blindly saying "we see some lag and we guess it is not on our side". The lag indicators we have in UI are a blanket "time-of-plot-creation minus event-timestamp", so several Kafka and processing latencies are included, in addition to the whole latency of the readout->RMM->EFU chain latency.

jl-wynen:
> > Is this for us to debug? This sounds like you are diagnosing issues with ECDC software. Don't they have debugging tools for this?
> 
> Maybe they do, but adding this avoids us blindly saying "we see some lag and we guess it is not on our side". The lag indicators we have in UI are a blanket "time-of-plot-creation minus event-timestamp", so several Kafka and processing latencies are included, in addition to the whole latency of the readout->RMM->EFU chain latency.

But what is the purpose of this? Is it a commissioning tool to help debug kafka? In that case, it is outside of our scope. Is it a timing tool for our workflows? I.e., do you want to use it to detect when a lag is due to our software being slow? Then we should measure that time directly.
I think users will ultimately not care about where the latency comes from. This seems to be a tool for expert, inhouse users.

SimonHeybrock:
> > > Is this for us to debug? This sounds like you are diagnosing issues with ECDC software. Don't they have debugging tools for this?
> > 
> > 
> > Maybe they do, but adding this avoids us blindly saying "we see some lag and we guess it is not on our side". The lag indicators we have in UI are a blanket "time-of-plot-creation minus event-timestamp", so several Kafka and processing latencies are included, in addition to the whole latency of the readout->RMM->EFU chain latency.
> 
> But what is the purpose of this? Is it a commissioning tool to help debug kafka? In that case, it is outside of our scope.

Is it? I can't simply make a bug report "our end-to-end chain from event timestamp to plot creation sees 10 seconds lags, please investigate". It is not about Kafka latency by the way, so far that has never been the culprit.

> Is it a timing tool for our workflows? I.e., do you want to use it to detect when a lag is due to our software being slow? Then we should measure that time directly. I think users will ultimately not care about where the latency comes from. This seems to be a tool for expert, inhouse users.

The users will not see this in the current implementation, as the lag reports are only shown in the logs, not the dashboard. Yes, we want to discriminate where the time lag comes from. You are right that we could do that by adding additional timestamps at raw-data-ingestion time based on the service's `now`, then feed that through the system. That seems slightly more complicated though and would require extending schemas. At raw data ingestion we already have two timestamps that tell us the latency of the chain that lies outside our domain, so it is simpler to use that instead.

jl-wynen:
If this only shows up in the logs, did you check whether Grafana already has that info? And has it been requested and checked with ECDC whether we need to have this feature?

SimonHeybrock:
> If this only shows up in the logs, did you check whether Grafana already has that info? And has it been requested and checked with ECDC whether we need to have this feature?

This is useful when debugging production issues in ESSlivedata, by simply looking into log files. Grafana is not easily accessible, e.g., by coding agents.

jokasimr:
I also don't think this fits in livedata, but I can see how it can be extremely useful for debugging during commissioning and quickly resolving issues like the ones discovered for Loki yesterday.

So I'm fine with it being added here unless there is a clear better place to do it / unless it already exists elsewhere. Having it in Grafana seems optimal.

**Are the definitions of "lag" here correct and well defined for all message types?**

For ev44 the definition is clear. But for other message types I'm not sure.

For example, f144 messages are re-sent regularly with the latest log value, even if no new movement or measurement was added to the log. In that case I'd expect the log timestamp to show the time when the log was last updated, not the current timestamp when the message was replayed. But if I understand correctly this will be registered as a "stale" log value here, even though it's not stale or incorrect.

SimonHeybrock:
> For example, f144 messages are re-sent regularly with the latest log value, even if no new movement or measurement was added to the log. In that case I'd expect the log timestamp to show the time when the log was last updated, not the current timestamp when the message was replayed. But if I understand correctly this will be registered as a "stale" log value here, even though it's not stale or incorrect.

Good point actually, this would likely be mis-reported as lag and we have no way of telling (unless we'd check if it was the *first* message with that timestamp, which we cannot a service startup).

SimonHeybrock:
I disabled this for f144 in the latest commit.

YooSunYoung:
It is a bit tricky to monitor latency of the publishing part using grafana since you have to see the payload timestamp...

But is this `lag` expected to happen in the production? Or is it considered error?
If it's error, I think it shouldn't go into the esslivedata. It should be monitored/tested by upstream data source.

SimonHeybrock:
So is the consensus here that I should replace this with a solution the measures the lag in our system? This would mean something as explained above:

> You are right that we could do that by adding additional timestamps at raw-data-ingestion time based on the service's now, then feed that through the system. That seems slightly more complicated though and would require extending schemas. At raw data ingestion we already have two timestamps that tell us the latency of the chain that lies outside our domain, so it is simpler to use that instead.

Basically, I would either take the raw data production timestamp (you can argue against that because it will measure latency of the Kafka part), or generate a timestamp at message ingestion (not great since the measurement would not include the time the message "waited" on our side). Then feed this timestamp through the backend message adapter chains, preprocessors, into the running jobs which produce a result, where we add a new field for the timestamp. The message sink can then log this before publishing. This would give us the intra-backend lag, to some approximation.
To get the full lag in our system, we would further store the timestamp as a new coord on the result `DataArray`, which will make it into `da00`. The dashboard can then either log or display this in the UI.

jokasimr (approved):
I'm fine with the current shape here.

The implementation is quite small, and we know it will be useful to notice these lags, at least for ev44, maybe not for the other message types.
If we notice that it causes issues or spams the log we can always revert it.

If this is eventually added to some other observability system later we can remove it here.

SimonHeybrock:
Thanks! Anyway, need this now to help debugging Loki, so I'll merge even if there is no consensus.

Issue #971, opened 2026-06-17
Correctness survey


SimonHeybrock:
Details below are raw results that need triaging.

---

I'll hunt for bugs in the core data-flow paths. Let me start by getting an overview of the codebase structure, then read the critical components carefully.

While those run, I'll audit `core/` myself, starting with the message batching and orchestration logic.

`Job.process` forces finalize when active data is added successfully, so no lag there. Continuing with the remaining core files.

The dashboard is ~15.5k lines, so the agent there will prioritize the data-path files. Meanwhile I'll audit the smaller cross-cutting modules myself.

Now the remaining config modules and `service_factory.py`.

While the environment installs, I'll read the kafka sink and consumer.

Confirmed — `DimensionError` propagates instead of triggering the reset path. Let me see who feeds `DataService` to gauge the blast radius.

All three confirmed. Let me also verify the `BufferError` behavior of confluent-kafka's producer (the sink only catches `KafkaException`).

All verification is done — I have a complete picture now. Here's the full report.

---

# Bug hunt results: esslivedata

I audited the core message flow, Kafka layer, handlers/accumulators, and dashboard data path (partly via parallel review agents, with all key findings re-verified by hand; four bugs reproduced with actual code execution). Findings ranked by severity.

### 1. Dashboard drops data after a workflow shape change (TemporalBuffer + Orchestrator)
`dashboard/temporal_buffers.py:342,481-491` — `TemporalBuffer.add()` is designed to re-initialize the buffer when incoming metadata doesn't match, but `_metadata_matches()` calls `new.assign(self._reference.data)`, which **raises `DimensionError`** when the data shape changed (e.g. the user rebins a workflow from 100 to 200 bins) instead of returning `False`. Reproduced:

```
buf.add(d_4bins); buf.add(d_4bins)   # ok
buf.add(d_8bins)                     # DimensionError instead of buffer reset
```

The blast radius comes from `dashboard/orchestrator.py:74-77`: `update()` forwards all polled messages in one loop with no per-message exception handling, so once one stream is poisoned, the exception aborts the loop and **every message sorted after it in the batch is silently lost — on every polling cycle** until the plot is closed. This only affects plots whose extractors force a `TemporalBuffer` (window/full-history); `SingleValueBuffer` plots are fine.

### 2. `Cumulative` accumulator crashes when coordinate structure changes
[...]

PR #1001, opened 2026-06-23
Strip unused compact_messages and PlotToPngSink

SimonHeybrock:
Addresses item 4 of #971.

`compact_messages` had a single user, `PlotToPngSink`, which was itself effectively dead: its standalone service (`services/plot_da00_to_png.py`) loaded `config_names.visualization`, an attribute that no longer exists, so the service could not even start. Rather than fix the function's timestamp-tie behaviour (it kept the oldest message on equal-timestamp ties instead of the latest), both the function and the sink are removed.

Services now always use `KafkaSink`; the `--sink-type` CLI argument is gone since `png` was its only non-default value.

## Test plan

- [x] Full fast test suite passes
- [x] `--sink-type` no longer appears in service `--help`


SimonHeybrock:
LGTM

PR #1003, opened 2026-06-23
Fix dead log-scale start validator in edges models

SimonHeybrock:
Item 3 of #971.

The `start_must_be_positive_if_log` validator on `EdgesModel` never ran its check: it read a non-existent `log` field (the field is `scale`), and as a `field_validator('start')` it could not have seen `scale` regardless, since `scale` is declared after `start`. Log-scale edges with `start <= 0` therefore validated cleanly on the frontend and only failed later inside workflow creation with an opaque backend job error (`ValueError: Geometric sequence cannot include zero` from `sc.geomspace`).

Replaced it with a `model_validator(mode='after')` that inspects the actual `scale` field, so the constraint is enforced at input validation with a clear message.

### DIFF
--- a/src/ess/livedata/parameter_models.py
+++ b/src/ess/livedata/parameter_models.py
@@ -14,7 +14,7 @@ from enum import StrEnum
 from pathlib import Path
 
 import scipp as sc
-from pydantic import BaseModel, Field, field_validator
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1004, opened 2026-06-23
Isolate per-message failures in dashboard ingestion loop

SimonHeybrock:
## Motivation

`Orchestrator.update` forwarded every polled message in a single loop with no per-message error handling. One poisoned stream (corrupt payload, or data whose shape no longer fits its buffer) raised out of `forward()` and aborted the loop, silently dropping every message sorted after it. Because the same offending message reappears on each polling cycle, the loss repeats indefinitely until the offending plot or job is removed.

This wraps each `forward()` in a try/except that logs and continues, so a single bad message no longer takes down ingestion for all other streams. The buffer write happens synchronously inside the loop (only subscriber notification is deferred to transaction commit), so this guard also covers buffer-level failures, not just malformed payloads.

Catching broad `Exception` is deliberate here: this is the live ingestion loop, where one backend's bad message must not halt all dashboards. That is fault isolation in a streaming loop, not masking a misconfiguration.

Companion to #1002; both address #971 (item 1). That PR stops `TemporalBuffer` from raising on a shape change; this one prevents any per-message failure in the loop from cascading.

## Test plan

- Added an `update()`-level regression test with a poisoned message followed by a valid one; verified it fails without the fix (whole batch aborted) and passes with it.

### DIFF
--- a/src/ess/livedata/dashboard/orchestrator.py
+++ b/src/ess/livedata/dashboard/orchestrator.py
@@ -74,7 +74,17 @@ class Orchestrator:
         with self._active_job_registry.ingestion_guard():
             with self._data_service.transaction():
                 for message in messages:
-                    self.forward(stream_id=message.stream, value=message.value)
+                    # Isolate per-message failures: a single poisoned stream (e.g.
+                    # corrupt payload, or data whose shape no longer fits its buffer)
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1005, opened 2026-06-23
Restart Cumulative and AreaDetectorView accumulation on incompatible data

SimonHeybrock:
## Problem

Two accumulators that sum data across batches got permanently stuck when the incoming data's structure changed at unchanged shape (e.g. shifted coordinates from an upstream reconfiguration), instead of restarting the accumulation. This was item 2 ("Cumulative accumulator crashes when coordinate structure changes") in the correctness survey #971, plus a sibling found while checking for the same pattern elsewhere.

**`Cumulative` preprocessor.** Only a coordinate change for 1D data with a coord on its dimension was handled. Any other structural change at unchanged shape — a 2D image whose coordinate values shift, a coordinate appearing mid-stream, a dtype change — poisoned the accumulator. It lives in the service-level message preprocessor, shared across jobs, so a **workflow restart does not clear it** (only a data shape change or a service restart did). The preprocessor catches and logs the exception as `preprocessing_error`, so the affected stream goes silent with no crash and no error surfaced to the frontend.

**`AreaDetectorView` workflow.** Accumulates the area-detector image across batches with `self._cumulative += transformed`. The same structural change makes `+=` raise; the workflow is not reset on error, so the job reports the failure to the frontend on **every subsequent batch until the workflow is restarted** — on exactly the high-rate, reconfiguration-prone area-detector path.

## Fix

Let `+=` be the sole compatibility check. It already validates shape, coordinates, units, variances, and dtype in fused C++, so a separate Python-level pre-check duplicated that work — measurably: ~110 µs/add of pybind overhead, several times the cost of the addition itself, which matters at the up-to-100 Hz area-detector rates these see (`add` runs per message). On any structural mismatch `+=` raises and the accumulation restarts from a copy of the incoming data. This is both faster (steady-state add drops ~9× for histograms, ~4× for images, leaving only the irreducible `+=`) and more robust against the stuck state: any incompatibility recovers, rather than only the cases an explicit check happened to enumerate.

For `AreaDetectorView`, the delta baseline (`_previous`) and the window start time are reset together with the cumulative, so the next `current` output is the full new accumulation rather than a mismatched subtraction.

One assumption for `Cumulative`: a coordinate present *only* in the accumulation (a coord vanishing mid-stream) does not raise — scipp keeps it — so that direction merges rather than resets. Coordinate presence is assumed stable per stream, which holds for the histogram and image sources that use this accumulator; the reverse direction (a coord appearing) still recovers via the exception path.

## Scope

Two related accumulators were checked and left for a separate issue (lower realism, and partly upstream code): `ToNXlog` (first sample locks dtype and shape; a later differing dtype is silently truncated rather than crashing — a distinct design question) and the monitor cumulative/window histogram `StreamProcessor` accumulators (same `+=` pattern via upstream `EternalAccumulator`, but histogram edges are fixed per job so a reconfiguration spawns a fresh accumulator).

## Test plan

- [x] Regression tests for the previously-broken paths in both accumulators (coord-value change at same shape, coordinate appearing, dtype change; area-detector reconfiguration restart and delta-baseline reset); verified to fail on the unfixed code and pass with the fix.
- [x] Existing handler suites pass.

### DIFF
--- a/src/ess/livedata/handlers/accumulators.py
+++ b/src/ess/livedata/handlers/accumulators.py
@@ -188,20 +188,29 @@ class _CumulativeAccumulationMixin:
         self._cumulative: sc.DataArray | None = None
 
     def _add_cumulative(self, data: sc.DataArray) -> None:
-        """Add data to the cumulative accumulation."""
-        if self._cumulative is None or data.sizes != self._cumulative.sizes:
+        """Add data to the cumulative accumulation.
+
+        Restarts the accumulation from a copy whenever incoming data is
+        structurally incompatible with it: a changed shape, a changed set of
+        coordinates, or changed coordinate values (e.g. rebinning or an upstream
+        reconfiguration). Otherwise the data is summed onto the accumulation.
+
+        ``+=`` already validates compatibility in fused C++, so it is the sole
+        check: any structural mismatch raises (``RuntimeError`` covers
+        dimension/coordinate/unit/variances errors, ``TypeError`` covers dtype),
+        and we restart rather than letting the accumulator get stuck rejecting
+        every subsequent batch. A coordinate *present only in the accumulation*
+        does not raise (``+=`` keeps it); coordinate presence is assumed stable
+        per stream, which holds for the histogram and image sources that use
+        this accumulator.
+        """
+        if self._cumulative is None:
             self._cumulative = data.copy()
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1006, opened 2026-06-23
Fix three backend robustness bugs: consumer assignment, sink backpressure, service exit code

SimonHeybrock:
Three independent backend robustness fixes from the correctness survey (#971), items 5, 6, and 7. Each turns a transient or startup-time fault into a silent, often permanent failure.

## Consumer partition assignment (item 7)

`make_bare_consumer` assigned partitions one topic at a time in a loop. `Consumer.assign` replaces the entire assignment, so only the last topic's partitions ended up manually assigned; the others depended on a later group rebalance and, with `auto.offset.reset=latest`, could silently skip early messages on startup. It also mixed `subscribe` and `assign`, which are mutually exclusive in librdkafka. All partitions of all topics are now collected and assigned in a single call, and `subscribe` is dropped.

## Sink backpressure death (item 5)

`KafkaSink` only caught `KafkaException`, but a full local producer queue raises `BufferError`. During a broker outage the queue fills until `BufferError` propagates out of the worker loop and kills the service, turning a transient hiccup into a permanent outage. The sink now drops the message and serves delivery callbacks to free queue space, consistent with its existing live-data drop semantics for non-fatal failures.

## Nonzero exit on worker-loop death (item 6)

When the worker loop raised, the service signalled the main thread and `_handle_shutdown` unconditionally called `sys.exit(0)`, so a crash was indistinguishable from a clean operator shutdown. Supervisors with `restart: on-failure` left the crashed service down. The exit code now comes from an overridable `_exit_code` property: 1 when a worker error was recorded, 0 for operator-initiated shutdown. This compounds with item 5 — together a transient broker problem could silently and permanently take a backend down.

## Test plan

- [x] `tests/kafka/` (308 passed)
- [x] New consumer tests: single assign covering all topics; raises on empty-partition topic
- [x] New sink test: `BufferError` from `produce` does not propagate
- [x] New service tests (subprocess): worker-loop error exits 1; SIGTERM shutdown exits 0

### DIFF
--- a/src/ess/livedata/core/service.py
+++ b/src/ess/livedata/core/service.py
@@ -49,7 +49,12 @@ class ServiceBase(ABC):
         self._logger.info("Received signal %d, initiating shutdown...", signum)
         self.stop()
         self._finalize_processor()
-        sys.exit(0)
+        sys.exit(self._exit_code)
+
+    @property
+    def _exit_code(self) -> int:
+        """Process exit code on shutdown; nonzero signals a fault to the supervisor."""
+        return 0
 
     def _finalize_processor(self) -> None:  # noqa: B027
         """Finalize processor after the worker thread has stopped.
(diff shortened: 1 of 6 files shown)


SimonHeybrock (commented):
LGTM

PR #1007, opened 2026-06-23
Fix subscriber skipped when subscriber list mutated during notification

SimonHeybrock:
Item 10 of #971. `DataService._notify_subscribers` iterated `self._subscribers` directly. When a subscriber is removed mid-iteration, the list-index iterator silently skips the subscriber positioned after the removed one, so it misses that notification cycle.

Two ways this is reached: a subscriber's `trigger` removing another subscriber, and — the production case — a UI-thread plot teardown or reconfiguration (`PlotOrchestrator.remove_cell` / `remove_grid` / `update_layer_config` → `unregister_subscriber`) racing the background ingestion thread that drives notifications. These two threads share the global subscriber list with no lock covering it. Impact is mild and self-healing (one skipped refresh, recovered on the next data batch), but it is a real correctness gap.

Fix iterates over a snapshot of the list. This resolves the skip but does not close the broader gap that `register_subscriber`/`unregister_subscriber` also mutate `TemporalBufferManager` extractors off-thread without synchronization; that is left as a follow-up since it warrants a deliberate locking decision.

## Test plan
- [x] New regression test fails without the snapshot, passes with it
- [x] `data_service_test.py` suite green

### DIFF
--- a/src/ess/livedata/dashboard/data_service.py
+++ b/src/ess/livedata/dashboard/data_service.py
@@ -184,7 +184,11 @@ class DataService(MutableMapping[K, V]):
         updated_keys
             The set of data keys that were updated.
         """
-        for subscriber in self._subscribers:
+        # Iterate over a snapshot: a subscriber's trigger, or a concurrent
+        # unregister from the UI thread, may mutate self._subscribers mid-loop.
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1008, opened 2026-06-23
Add Table plotter for 0D scalar data

SimonHeybrock:
## Summary

Adds a **Table plotter** for 0D scalar data as an alternative to the Bars plotter, and the supporting machinery for cells that contain layers which cannot be fused into a single Bokeh figure.

This was requested by at least one instrument, since visualizing, e.g. monitor counts in a plot like a bar chart is not as useful as seeing a value in a table.

Config options of the Table plotter:
screenshot-2026-06-23T13:39:15
Table view of monitor count readings:
screenshot-2026-06-23T13:38:22

## Non-overlayable layers

Tables (and layout-mode plots) render as a `DataTable` / `hv.Layout`, neither of which can be fused via `hv.Overlay`. Rather than detecting this at runtime, non-overlayability is now a config-time property: `Plotter.is_overlayable` (False for `TablePlotter`), with layout mode always producing an `hv.Layout` so the property depends only on params, not the runtime source count. `PlotOrchestrator` forbids placing a non-overlayable layer in a cell with other layers; the cell-properties modal hides the overlay-add button for such bases, and add/reconfigure flows surface the rejection as an error toast. This removes a latent `Overlay([Layout, ...]).collate()` crash by construction.

## Number formatting

Value columns are right-aligned (Bokeh `DataTable` has no decimal-separator alignment) and formatting is configurable via `PlotParamsTable.format`: notation (`auto` / `scientific` / `decimal` / `compact`) and precision. Compact uses numbro's k/m/b/t abbreviations; negative precision rounds the magnitude to powers of ten in decimal notation (e.g. `-3` → thousands). The formatter is carried as a plot option on a custom `TablePlot` subclass so it persists across streaming updates and reaches columns whose names sanitize to a different Bokeh field — both defects of the initial finalize-hook approach.

## Test plan

Unit tests cover the plotter pivoting, formatter selection per notation/precision, persistence across streaming updates, and the orchestrator's non-overlayable rejection.

Manual UI checks:
- [x] Table plotter renders 0D scalar outputs; numbers readable and right-aligned across notations
- [x] Adding a second layer onto a table (or a table onto an existing cell) is rejected with a clear toast
- [x] Overlay-add button is hidden in the cell-properties modal for table/layout bases


src/ess/livedata/dashboard/widgets/cell.py
```diff
@@ -616,20 +616,17 @@ def _compose_plot(self) -> hv.DynamicMap | hv.Element | None:
                 # and carry no autoscale axes; exclude them from the controller.
                 if state.plotter is not None and not layer.config.is_static():
                     cell_plotters.append(state.plotter)
+                # Tables and layout-mode plotters produce a DataTable/Layout with
```
nvaytet, line 619: I did not understand what is a 'layout-mode plotter'. Can you explain a little?

src/ess/livedata/dashboard/widgets/cell_properties_modal.py
```diff
@@ -211,11 +229,17 @@ def _render_layers(self) -> None:
                 get_source_title=self._orchestrator.get_source_title,
             )
             controls: list = []
-            overlays = overlay_suggestions_for_layer(
-                layer.config,
-                existing,
-                self._workflow_registry,
-                self._plotting_controller,
+            overlays = (
```
nvaytet: Minor:

To improve readability, can we set `overlays` in 2 steps? Maybe like
```Py
overlays = []
if self._plotting_controller.is_overlayable(
    layer.config.plot_name, layer.config.params
):
    overlay_suggestions_for_layer(
                    layer.config,
                    existing,
                    self._workflow_registry,
                    self._plotting_controller,
                )
```

src/ess/livedata/dashboard/plotter_registry.py
```diff
@@ -37,7 +37,6 @@ class DataRequirements:
     required_coords: list[str] = field(default_factory=list)
     deny_coords: list[str] = field(default_factory=list)
     required_dim_names: list[str] = field(default_factory=list)
-    multiple_datasets: bool = True
```
nvaytet, line 40: Can you explain why this is no longer needed?

src/ess/livedata/dashboard/plotting_controller.py
```diff
@@ -252,6 +252,19 @@ def create_plotter(
             plotter.set_roi_publisher(self._roi_publisher)
         return plotter
 
+    def is_overlayable(self, plot_name: str, params: dict | pydantic.BaseModel) -> bool:
+        """Whether a layer with this config can share a cell with other layers.
+
+        Tables and layout-mode plotters produce elements that cannot be fused
+        via ``hv.Overlay``. A config that fails to build is treated as
+        overlayable; the failure surfaces later during plotter creation.
```
nvaytet: > the failure surfaces later during plotter creation.

A bit of an unusual pattern to me. I didn't follow exactly where the failure surfaces in the end, but maybe it's fine?
In what case does `create_plotter` fail?

src/ess/livedata/dashboard/plotting_controller.py
```diff
@@ -252,6 +252,19 @@ def create_plotter(
...
+        """
+        try:
+            plotter = self.create_plotter(plot_name, params=params)
+        except Exception:
+            return True
+        return getattr(plotter, 'is_overlayable', True)
```
nvaytet, line 272: So a plotter that doesn't have the `is_overlayable` attr also returns `True`? I guess the argument would again be that 'the failure surfaces later during plotter creation.' ?

src/ess/livedata/dashboard/table_plotter.py
```diff
@@ -0,0 +1,181 @@
...
+        if self._notation is TableNotation.scientific:
+            # Empty fixed-point window (limits equal) forces scientific notation
+            # for every magnitude.
+            return ScientificFormatter(
+                precision=decimals,
+                power_limit_low=0,
+                power_limit_high=0,
+                text_align='right',
+            )
+        return ScientificFormatter(precision=decimals, text_align='right')
```
nvaytet, line 173: So is this the `auto` case? (the one that switches to scientific for large and small numbers, and decimal otherwise?)

src/ess/livedata/dashboard/table_plotter.py
```diff
@@ -0,0 +1,181 @@
...
+
+        unit = next(iter(units)) or None
+        value_dim = hv.Dimension(output_name, label=label, unit=unit)
+        return hv.Table(
+            {'source': rows, value_dim.name: values},
+            kdims=kdims,
+            vdims=[value_dim],
+        )
+
+    def _value_formatter(self):
```
nvaytet, line 146: Does the code handle datetime values properly? There could be some use case where this is required?

nvaytet (commented):
Overall changes look fine, I mostly have questions rather than change requests.

src/ess/livedata/dashboard/widgets/cell.py
SimonHeybrock: There is `hv.Overlay` and `hv.Layout`, the latter puts, e.g., image plots in a grid-layout. We cannot overlay sth. else on that.

src/ess/livedata/dashboard/plotter_registry.py
SimonHeybrock: It was dead before this change. A leftover from before we had the current system of creating plots by adding layers. The old UI allowed for selecting multiple dataset for plotting in a single plot directly.

src/ess/livedata/dashboard/table_plotter.py
SimonHeybrock: Can't think of any case where a datetime is the *value* -- it is always a coord as far as I am aware.

src/ess/livedata/dashboard/widgets/cell_properties_modal.py
SimonHeybrock: Agreed, clearer that way. Done.

src/ess/livedata/dashboard/plotting_controller.py
SimonHeybrock: Good question -- the pattern was under-explained. `is_overlayable` isn't a static flag: for the general plot it derives from the params (`combine_mode`), so we have to build the plotter to ask it. `create_plotter` raises when params are incomplete/invalid for that plotter type. At these call sites the config comes from an existing layer, so it won't actually fail; the `try` is defensive. We return `True` on failure because this method only gates the overlay UI (disabling "Add layer", hiding overlay buttons) -- a genuinely broken config should not be blocked here but raise again at actual plotter creation, the authoritative path. Clarified in the docstring.

src/ess/livedata/dashboard/plotting_controller.py
SimonHeybrock: Yes, same reasoning. The `getattr(..., True)` default covers plotters that don't define the attr at all; like the build-failure case, defaulting to overlayable just avoids blocking the UI, and any real problem surfaces at plotter creation. Noted in the clarified docstring.

src/ess/livedata/dashboard/table_plotter.py
SimonHeybrock: Yes, exactly -- this is the `auto` fall-through. `ScientificFormatter`'s default power limits (`-3`/`5`) render decimal within 1e-3..1e5 and switch to scientific outside. Added a comment to spell that out.

src/ess/livedata/dashboard/table_plotter.py
nvaytet: I never cease to be surprised at what users come up with, but we're probably ok for now :wink:

nvaytet (approved):

PR #1009, opened 2026-06-23
Serialize DataService access with a lock

SimonHeybrock:
`DataService` and its `TemporalBufferManager` were mutated from two threads with no synchronization between them: the background ingestion thread (`transaction` / `__setitem__`) and the UI thread (`register_subscriber` / `unregister_subscriber`, which calls `add_extractor` and can swap the buffer object). The `ActiveJobRegistry` ingestion guard only serialized the active set and buffer eviction, not the register/`add_extractor` paths. Adding or reconfiguring a plot on an already-running workflow could therefore race a concurrent buffer write on the same key.

The hazard is worse than a dropped batch. Buffer `get()` returns views into a backing store that `add()`/`drop()` overwrite in place, and the latest- and full-history extractors pass those views straight through. A concurrent write can mutate data a plot is already holding.

The fix guards all internal `DataService` state with a single `RLock`. The important design point: the lock is never held while calling `subscriber.trigger`. Extracted data is copied out of the buffers under the lock, then triggered with the lock released. This detaches compute from the live buffer (so it cannot observe a concurrent mutation) and keeps arbitrary compute off the lock, so neither the Bokeh document lock nor the ingestion guard can deadlock against it. Lock ordering is `ingestion_guard` → `DataService` lock, documented on the class.

## Test plan

- [x] New `TestThreadSafety` cases: pass reliably on this branch, and the two concurrency cases fail consistently on `main` (confirming they exercise the race).
- [x] Full `tests/dashboard` suite green.
- [ ] Manual: add/reconfigure a plot on a live workflow under data flow; confirm no blank frames or stale data on the new plot.

### DIFF
--- a/docs/developer/adr/0005-frame-gated-session-flush.md
+++ b/docs/developer/adr/0005-frame-gated-session-flush.md
@@ -1,6 +1,6 @@
 # ADR 0005: Frame-gated per-session plot flush
 
-- Status: accepted
+- Status: accepted (amended 2026-07-02, see bottom)
 - Deciders: Simon
 - Date: 2026-06-25
 
@@ -183,3 +183,28 @@ above, and this section should be revisited if intra-tab stagger proves visible.
   compute per batch (observed in practice) holding the loop. The result is a
   one-tick (~100 ms) stagger: acceptable and self-correcting. How often this
   happens in real traffic is unmeasured.
+
+## Amendment (2026-07-02): pull-on-frame input side
+
+The input side of the mechanism changed from push to pull (issue #1044). As a
+burst drains, `DataService` no longer extracts data and delivers it to each
+layer's subscriber; it notifies keys-only, and the orchestrator merely marks
+the affected layers dirty (per grid). `flush_frames` then pulls each dirty
+*viewed* layer's input through its extractors from the current buffer state
+(`DataService.snapshot`), rebuilds, and commits the grid as before. Dirty
+flags of unviewed layers are dropped; the 0→1 viewer transition rebuilds from
+current buffer state unconditionally.
+
+Consequences for the guarantees above:
+
+- The frame a grid's layers repaint from is now *the buffer state at flush
+  time* rather than *the drained burst's payload*. Layers of one grid are
+  pulled back-to-back on the ingestion thread with no writes in between, so
+  intra-grid coherence is unchanged.
+- A logical frame that splits across bursts now heals at the next flush by
+  construction (the pull sees whatever has arrived), rather than by a
+  one-tick stagger of stashed payloads.
+- Deferring or coalescing flushes is lossless: a pull can never observe older
+  data than a delivery would have carried. This is what makes a future
+  min-frame-interval throttle in `flush_frames` a policy choice rather than a
+  correctness question.
--- a/docs/developer/adr/0005-frame-gated-session-flush.md
+++ b/docs/developer/adr/0005-frame-gated-session-flush.md
@@ -200,7 +200,10 @@ Consequences for the guarantees above:
 - The frame a grid's layers repaint from is now *the buffer state at flush
   time* rather than *the drained burst's payload*. Layers of one grid are
   pulled back-to-back on the ingestion thread with no writes in between, so
-  intra-grid coherence is unchanged.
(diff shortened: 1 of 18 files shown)


SimonHeybrock:
Merge plan for this stack (#1009 ← #1043 ← #1045), as part of the #1046 program:

**Treat the stack as one unit of intent and land it bottom-up in one sequence** — do not land #1043 and pause. #1045 deletes machinery #1043 adds (the `active` gate, the stash/`ComputeTask` plumbing and their documented races), so the intermediate state has no soak value and its manual-verification burden would be paid twice. One manual verification session against a real backend can cover all three PRs' open checklist items at once (add/reconfigure plot on a live workflow; tab switching across 2+ sessions with hidden-tab fill and timeseries continuity; update cadence + freshness pill; multi-plot repaint without stagger).

What each PR contributes to the end state: this PR's lock and `TestThreadSafety` (scope of the lock shrinks to buffer writes + `snapshot` once #1045 is in — worth a docstring touch-up then, not a rework now); #1043's buffering-while-hidden split and the 0→1 snapshot refresh; #1045's pull inversion, which makes "no viewers → no work" the default and is the prerequisite for #1042 Option B slice (i) (design being posted on #1042).

---

SimonHeybrock:
Full-stack review of the combined branch (this PR + #1043 + #1045, reviewed as one changeset per the D1 decision on #1046) found four cross-PR issues that the individual diffs hid. Fixed in bece0dea2 and a4b11119f:

**bece0dea2 — transaction-scoped DataService lock.** The per-operation lock made transactions atomic for *notifications* but not for *pulls*: a snapshot from the polling thread (0→1 activation) could observe a mid-burst state, and a flush pull racing `ActiveJobRegistry.cleanup` could observe a partial eviction — exactly the "subset of sources" bug cleanup's docstring claims the transaction prevents. Worse, the partial-eviction frame would persist, since the post-eviction pull assembles `None` and skips the rebuild. The lock is now held for the whole transaction (notifications still run outside it; depth stays thread-local). This is deliberately the primitive the #1042 slice (i) generation-flip design needs: "pulls observe only committed state" now comes from `DataService` instead of having to be re-derived there.

Status/ack messages are hoisted out of the ingestion transaction: they never touch DataService, and JobService/ServiceRegistry callbacks must not run under the now longer-held lock. Handling control before data also fixes a real (if narrow) loss: an ack that activates a job now admits the same batch's data instead of dropping it at the active filter.

**a4b11119f — pull-and-build hardening.**
- Extraction moved from a guarded context (`_notify_subscribers` caught per-subscriber) into `_pull_and_build`, which ran unguarded — and `flush_frames` sat outside `_update_loop`'s try. A raising extractor would have killed the update thread and silently frozen ingestion for every session. Extraction failures now transition the layer to ERROR like compute failures, and `flush_frames` is inside the guard as a backstop.
- The old stash design swapped plotter + pending atomically under the gate lock; the pull inversion dropped that, so a job restart racing an in-flight build could push data assembled through the old subscription into the new plotter and flip it READY with no cached state. Builds now capture the layer version before the pull and skip READY/ERROR transitions when the version moved on. Residual: the window shrinks from the whole pull+compute to nanoseconds; closing it fully needs transitions gated inside `LayerStateMachine`, which the #1042 slices (ii)–(iv) rework owns.
- `_refresh_layer` now tolerates concurrent layer removal (same race deb2c8324 fixed for the dirty-mark path).

Six new regression tests, each verified to fail on the pre-fix code (the H6 standard): transaction mutual exclusion, snapshot atomicity, ack-before-data admission, raising extractor → ERROR, stale build vs. swapped plotter, removal during activation build.

Notes for the record: the `.copy()` discipline makes `DataService` de-facto scipp-only (`__getitem__`/`snapshot` require copyable values); the generic `V` and the `DataService[ResultKey, object]` annotation in `workflow_controller.py` overstate what it accepts. And `snapshot` double-copies for aggregating extractors (they already return a fresh array) — extractor-declared ownership is future work if the lock-held copy ever shows up in profiles.

SimonHeybrock:
LGTM, even though I think we should do a fresh review of the final code state after this series of refactorings is done.

PR #1010, opened 2026-06-23
Stop/reset scheduled jobs on workflow and broadcast commands

SimonHeybrock:
Addresses item 8 of #971.

A stop/reset command targeting a `workflow_id`, or a broadcast command (no `job_id`, no `workflow_id`), only reached active and pending-context jobs. Scheduled-but-not-yet-active jobs were left untouched and later activated as zombies the user believed were stopped. The per-`job_id` path already handled every job bucket, so this aligns the broadcast/by-workflow path by iterating `all_jobs`.

Also corrects the `JobManagerAdapter` docstring, which described a legacy replace-on-restart behavior that `set_workflow_with_config` never implemented — it only schedules a new job.

## Test plan

- [x] New unit tests cover stop-by-workflow and broadcast-stop reaching a scheduled job; both fail without the fix.

### DIFF
--- a/src/ess/livedata/core/job_manager.py
+++ b/src/ess/livedata/core/job_manager.py
@@ -369,8 +369,7 @@ class JobManager:
             self._perform_action(action=command.action, sel=lambda job: True)
 
     def _perform_action(self, action: JobAction, sel: Callable[[Job], bool]) -> None:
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
LGTM

PR #1012, opened 2026-06-25
Use cached watermark offsets for consumer lag

SimonHeybrock:
## Problem

PROD detector services flood the journal with `REQTMOUT` warnings for `ListOffsetsRequest`:

```
REQTMOUT|rdkafka#consumer-2| Timed out ListOffsetsRequest in flight (after 100ms, timeout #0): possibly held back by preceeding FetchRequest with timeout in 59998ms
```

The source is the consumer-lag monitor. Every 30s its background thread called `get_watermark_offsets(tp, timeout=0.1)` (default `cached=False`) for every assigned partition, issuing a live `ListOffsetsRequest` to the broker. rdkafka serializes requests per broker connection, so each of these queues behind the consumer's long-poll `FetchRequest` (up to 60s) and times out after 100ms — exactly what the log's "held back by preceeding FetchRequest" notes. A detector consumer spanning many brokers therefore produces a burst of timeouts every metrics cycle.

## Fix

Query with `cached=True`. The high watermark is piggybacked on every `FetchResponse` for partitions we actively consume, so the cached value is fresh with no network round-trip, removing the `ListOffsetsRequest`s entirely. `position()` is already local.

Trade-off: cached high only refreshes while messages flow on a partition, so a genuinely idle partition could report stale lag — but that is the case where lag is ~0 and uninteresting.

## Follow-on simplification

The lag computation ran in its own thread purely to keep those blocking ListOffsets calls off the consume loop. With cached offsets it is a local, microsecond-cheap operation, so the thread, lock, and cached-lag indirection are removed and lag is computed inline where metrics are logged. As a side effect the consumer is no longer accessed concurrently from two threads.

### DIFF
--- a/src/ess/livedata/kafka/source.py
+++ b/src/ess/livedata/kafka/source.py
@@ -142,7 +142,6 @@ class BackgroundMessageSource(KafkaMessageSource):
         self._batches_dropped_since_last_metrics = 0
 
         # Lag monitoring (computed in a separate thread to avoid blocking consumption)
-        self._lag_query_timeout = 0.1
         self._lag_lock = threading.Lock()
         self._cached_lag: dict[str, Any] | None = None
         self._lag_thread: threading.Thread | None = None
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
LGTM

PR #1014, opened 2026-06-25
NICOS derived devices: contract, output extraction, and dashboard surface

SimonHeybrock:
## Why

NICOS drives counting and multi-day scans against derived quantities computed by ESSlivedata (monitor/detector total counts, fitted peak statistics). It needs them as a **static, stably-named device list** on a **dedicated low-volume topic**, must be able to **reset** the accumulation behind one to start a fresh count, and must not have them reset/stopped/reconfigured by accident. This wires up that mechanism end-to-end and surfaces it in the dashboard. The design rationale and the alternatives considered live in `docs/developer/adr/0006-nicos-derived-devices.md`.

## What

**Contract (`config/device_contract.py`).** The `(WorkflowId, source_name, output_name) -> device_name` mapping is **derived from the workflow registry**: a `WorkflowSpec` declares `device_outputs` (output field -> device-name template) and the contract expands that over the spec's source names. The registry is the single source of truth, so the contract cannot drift from it -- an output that does not exist, or a source a spec does not declare, simply cannot appear; the one remaining failure mode (a template rendering a duplicate device name) fails loud at construction. `WorkflowId.version` is part of device identity, so a breaking change retires the old device name rather than silently changing semantics. Each instrument's `device_contract.yaml` is a generated, committed export for NICOS's static-list preference, guarded by a drift test. Bifrost ships the first contract: each monitor's cumulative total counts as a device.

**Device extraction (`core/nicos_devices.py`).** A `DeviceExtractor` selects the contracted outputs from each job result and republishes them on a dedicated `LIVEDATA_NICOS_DATA` topic, keyed by stable device name -- free of the random `job_number` carried on the main data path. Cumulative outputs already carry a 0-D `start_time` coordinate (stamped by the `JobManager` as the first-data time of the current generation); the extractor republishes it untouched, so NICOS can detect when an accumulation was reset by a job transition or backend restart. `start_time` is constant for the lifetime of a generation and advances on reset/restart with no persisted counter -- the same coordinate the dashboard uses to label a plot's accumulation interval, so the reset signal cannot drift from the value's provenance. Wire schema is `da00`, reusing the existing data serializer.

**NICOS-initiated reset.** NICOS resets the accumulation behind a device by producing a `JobCommand` JSON on the `livedata_commands` topic, keyed by the `WorkflowId` string the contract export already lists -- no `job_number`, no new topic. Omitting `job_id` selects every running job of that workflow, so a single device or several counted in sync reset together (there is no per-source selector); omitting `message_id` means no service acks and the `start_time` generation jump on the device topic is the confirmation. `WorkflowId` now also validates from its `instrument/name/version` string form on the wire, so NICOS sends the exported string verbatim while the dashboard keeps sending the nested object.

**Dashboard (read-only + protection).** A workflow is "device-bearing" iff a running job's output appears in the contract. The Workflows surface gains a device-bearing badge, per-output device markers, and a confirmation gate intercepting stop / reset / commit on device-bearing workflows (staging stays free; the gate re-reads live job state inside the orchestrator). A read-only "Derived devices" overview lists currently exposed devices (contract intersect running jobs). There is no runtime activation toggle and no dashboard->backend control channel -- exposure is declared once in the contract; the dashboard only reflects and protects it.

## Test plan

- [x] Manual: bifrost monitor workflows show the device badge/markers; stop/reset/commit prompts confirmation; derived-devices overview lists live devices.
- [x] Manual: a reset command keyed by the contract `workflow_id` string zeros the device and bumps `start_time`.


docs/developer/adr/0006-nicos-derived-devices.md
```diff
@@ -0,0 +1,209 @@
...
+`python -m ess.livedata.config.device_contract`; a test fails if a committed
+export drifts from the registry. `WorkflowId.version` is part of device identity,
+so a breaking change retires the old device name — it goes silent on the wire and
+NICOS notices — rather than silently changing semantics under a stable name.
+
+### A single unified job per `(workflow, source)`, gated in the dashboard
+
+There are no dedicated NICOS jobs, no separate orchestrator, no fixed
+`job_number`. One job per `(workflow, source)` is shared by dashboard and NICOS,
+and is **live as a device iff its owning job is running** — the backend extracts
```
YooSunYoung, line 114: typo?
```suggestion
and is **live as a device if its owning job is running** — the backend extracts
```

docs/developer/adr/0006-nicos-derived-devices.md
```diff
@@ -0,0 +1,209 @@
...
+expands that over the spec's `source_names`. The registry is the single source of
+truth, read directly by the backend (what to extract) and the dashboard (which
+outputs are devices).
+
+Each instrument's `device_contract.yaml` is a **generated, committed export** for
+NICOS's static-list preference; nothing reads it back. Regenerate with
+`python -m ess.livedata.config.device_contract`; a test fails if a committed
+export drifts from the registry. `WorkflowId.version` is part of device identity,
+so a breaking change retires the old device name — it goes silent on the wire and
+NICOS notices — rather than silently changing semantics under a stable name.
```
YooSunYoung: What is the breaking change in this context?

docs/developer/adr/0006-nicos-derived-devices.md
```diff
@@ -0,0 +1,209 @@
...
+- **Granularity is whole-workflow.** Omitting `job_id` selects every running job
+  of the workflow; all its sources reset together. This matches NICOS usage — a
+  single device, or several counted in sync (e.g. a difference of two monitors) —
+  and there is no per-source selector. Reset acts on the job, so it zeros all of
+  that job's outputs together.
+- **Carries a `message_id`.** NICOS generates a uuid like any command producer. The
+  owning worker acks it with `device: "all"` — the command omits `job_id`, so the
+  random `job_number` never appears in the ack. NICOS may correlate that ack or rely
+  on the `start_time` generation jump on the device topic, which also reports a reset
+  from any source, not only NICOS's own command.
```
YooSunYoung: I don't quite understand this part.

docs/developer/adr/0006-nicos-derived-devices.md
```diff
@@ -0,0 +1,209 @@
...
+provenance.
+
+### Registry-derived, committed contract
+
+The `(workflow, source, output) → device` mapping is **derived from the workflow
+registry**, not hand-authored. A `WorkflowSpec` declares `device_outputs` (output
+field → device-name template); `DeviceContract` (`config/device_contract.py`)
+expands that over the spec's `source_names`. The registry is the single source of
+truth, read directly by the backend (what to extract) and the dashboard (which
+outputs are devices).
```
YooSunYoung, line 78: It was easier to understand when I read the device_contract.yaml file. Can you add an example yaml snippet maybe...?

docs/developer/adr/0006-nicos-derived-devices.md
```diff
@@ -0,0 +1,209 @@
...
+- a read-only **NICOS devices** overview (contract ∩ running jobs).
+
+The gate re-reads live job state inside the orchestrator method, not only in the
+widget, so an action that races a job restart cannot act on stale state.
+
+### NICOS-initiated reset
+
+NICOS resets the accumulation behind a device by producing a `JobCommand` JSON on
+the `{instrument}_livedata_commands` topic, keyed by the `WorkflowId` it already
+holds from the contract export:
```
YooSunYoung: What's the usecase of NICOS resetting the workflow accumulative result?
Does it mean NICOS'll have to reset the result in the middle of a run...? 

This resetting will be useful when users don't want to record a file and they want to see the accumulative result as time goes but the context section says it's for users driving multi day scans.

docs/developer/adr/0006-nicos-derived-devices.md
```diff
@@ -0,0 +1,209 @@
...
+| Separate generation token variable | Redundant with `start_time`, which already exists and already drives plot labelling. Rejected. |
+| Dedicated NICOS jobs | Physical isolation guarantees nothing a UI gate doesn't, while duplicating heavy detector-stream processing and maintaining a permanent parallel code path. Rejected. |
+| Runtime per-output activation toggle | Adds a control channel and persisted state the design avoids, and fights the static-list premise — liveness would depend on operator UI state. The low-volume extraction is harmless when unscanned. Rejected. |
+| Per-`source_name` reset selector | Not needed given whole-workflow usage, and avoids perturbing dashboard bookkeeping that has no per-source reset state. Rejected. |
+| Dedicated inbound NICOS command topic keyed by device name | Cleaner decoupling, but a new topic plus a contract-resolution translation point. Reusing `livedata_commands` needs no new wiring and NICOS already holds the `WorkflowId`. Rejected for v1. |
+
+## Key design choices
```
YooSunYoung, line 184: The `Decision` and `Key design choices` seem to have repeating information.
Or are the subsections related to each other but emphasizing different aspects...?

YooSunYoung (commented):
I like how it can integrate data stream more easily.
I only have some minor questions and comments.

docs/developer/adr/0006-nicos-derived-devices.md
SimonHeybrock: Not a typo, 'iff' is a common abbreviation for "if and only if".

docs/developer/adr/0006-nicos-derived-devices.md
SimonHeybrock: This is about updates to workflows: Say we want to change the behavior of a workflow in an incompatible manner (such as changing the meaning of an output), then we increment the `WorkflowId.version` such that any NICOS script that references the device doesn't silently continue working (but doing the wrong thing). This is assuming NICOS checks the version.

docs/developer/adr/0006-nicos-derived-devices.md
SimonHeybrock: Yes, NICOS needs to reset, e.g., when starting a "count" command, or many time during a "scan" command -- move motor, reset, wait until target count reached, move motor, reset, ...

docs/developer/adr/0006-nicos-derived-devices.md
YooSunYoung: Ah haha I didn't know that thanks...!

SimonHeybrock:
Thanks @YooSunYoung!

I think the code changes are low-risk, the ADR was the main important bit to review. As we want to try out and prove this mechanism in practice, I'll rely on my own (done) + agent reviews (done) for now.

PR #1018, opened 2026-06-26
Do not force-finalize a cleared accumulator after reset

### LINKED ISSUE #1016: Reset + forced retry-finalize emits a zero-valued result at epoch 0
## Summary

A job result can be finalized with `start_time is None`, producing a spurious zero-valued result timestamped at the Unix epoch (`Timestamp.from_ns(0)`). This affects the main reduced-data path, not just the NICOS projection.

## Mechanism

`reset_job` clears the accumulator and nulls `Job._start_time` (`core/job.py:443-447`) but does **not** clear `JobManager._jobs_with_primary_data`. A job whose previous `finalize()` errored is deliberately kept in that set to retry (`core/job_manager.py`, `_record_compute_result`). The two interact:

1. A job's finalize errors → it stays in `_jobs_with_primary_data`, job stays active.
2. A later `process_jobs` batch fires a scheduled / run-transition reset via `_advance_to_time` → processor cleared, `_start_time = None`.
3. That same batch carries no active primary data for the job → `has_data=False` but `has_pending=True` → forced `finalize=True`.
4. `Job.get()` finalizes the freshly-cleared accumulator with `start_time=None` → a zero-valued result.

The result is then turned into a message with:

```python
# core/orchestrating_processor.py
timestamp=result.start_time or Timestamp.from_ns(0),
```

so it is published with a timestamp of epoch 0 (1970-01-01). The NICOS projector has the same fallback, but the main `LIVEDATA_DATA` path is affected too — an epoch-0 point can land in normal plots.

## Why it matters

- Emits a zero-valued cumulative result from a cleared accumulator, which downstream consumers may mistake for a genuine reading.
- The epoch-0 timestamp can corrupt time-axis handling / lag computation on the dashboard.
- The `or Timestamp.from_ns(0)` fallback silently masks the anomaly rather than failing fast.

## Reproduction sketch

Narrow but real: needs a finalize error followed by a reset on a batch that carries no new active data for that job. Not yet covered by a test — worth adding one that drives `JobManager` through this sequence and asserts no zero/epoch-0 result is emitted.

## Possible directions

- Clear `_jobs_with_primary_data` (and any retry state) for a job on reset, so a cleared accumulator is never force-finalized.
- And/or do not emit a result when `start_time is None` (treat the epoch-0 fallback as an assertion failure rather than a silent default).

Discovered while removing the redundant generation-token mechanism from the NICOS projection (the projector now relies on the `start_time` coordinate, which exposed this latent fallback).

SimonHeybrock:
Fixes #1016.

A job whose `finalize()` errored is deliberately kept in `_jobs_with_primary_data` so the next batch retries finalization. A run-transition reset (`reset_job` via `_advance_to_time`) clears the accumulator and nulls `_start_time`, but did not touch that retry set. If the resetting batch carried no new active primary data for the job, the retained retry flag forced a finalize on the freshly-cleared accumulator, emitting a spurious zero-valued result with `start_time=None`. Downstream this was published with an epoch-0 fallback timestamp, which could corrupt time-axis handling and lag computation, and masked the anomaly as a genuine reading.

The fix drops the retry state in `reset_job`: a cleared accumulator has nothing to finalize until new primary data arrives. The silent `or Timestamp.from_ns(0)` fallback in `_job_result_to_message` is replaced by a fail-fast check — a successful result must always carry the start time of its accumulation window, so a missing one is now an error rather than a masked epoch-0 point.

Discovered while removing the redundant generation-token mechanism from the NICOS projection, which exposed this latent fallback on the main reduced-data path.

## Test plan

- New regression test drives `JobManager` through finalize-error → run-transition reset → empty batch and asserts no result is emitted from the cleared accumulator (fails before the fix).
- New unit tests for `_job_result_to_message` covering the normal path and the fail-fast on missing start time.


SimonHeybrock (commented):
LGTM

PR #1020, opened 2026-06-29
Wire shared monitor workflow factory into instruments

SimonHeybrock:
`create_monitor_workflow_factory` was tested but never wired into production: six instruments (dummy, odin, nmx, tbl, bifrost, estia) each hand-rolled an identical TOA-only closure. The factory could not be reused as-is because it was typed against `MonitorDataParams`, whereas those specs register `TOAOnlyMonitorDataParams` — a sibling model, not a subclass — so `attach_factory`'s `issubclass` check rejected the pairing.

This introduces `MonitorDataParamsBase` as the common interface for both parameter models and types the factory against it. Rather than attach the shared factory in each instrument (config-free boilerplate), `load_factories` now attaches it to any monitor spec that `setup_factories` did not give a factory of its own. This stays in the backend-only factory phase: doing it at spec registration would also run in the dashboard, which imports specs but never calls `load_factories` and must hold no factory references. The concrete parameter model still comes from the spec, so the restricted TOA-only UI form and runtime validation are unchanged. DREAM and LOKI keep their own lookup-table factories.

## Test plan

- [x] Full fast suite passes
- [x] Verified at runtime that all six instruments resolve `TOAOnlyMonitorDataParams` with the shared factory, DREAM/LOKI retain their custom factories, and workflow creation succeeds in every case

### DIFF
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -713,6 +713,8 @@ class Instrument:
         if hasattr(module, 'setup_factories'):
             module.setup_factories(self)
 
+        self._attach_default_monitor_factories()
+
         self.validate()
 
         for name in (*self.detector_names, *self._pixellated_monitors):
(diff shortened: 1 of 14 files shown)


SimonHeybrock (commented):
LGTM, even though one can argue that this may not be worth it, given that in the future all instruments will need wavelength mode &mdash; however not in the current shape and form that requires passing a table in a bespoke registration call, since we intend to use the output stream of the wavelength-LUT workflow for this. In that light, this refactor removes repeated boilerplate that would have been removed then, so that changes will be slightly simpler.

PR #1021, opened 2026-06-29
core: consolidate JobManager job state into a single record map

SimonHeybrock:
## Motivation

`JobManager` smeared each job's state across three phase-bucket dicts plus parallel maps for schedules, states, error/warning messages, a primary-data set, and a finishing list. Consistency was maintained by hand, and the reported state was re-derived from bucket membership and a separate state map that could disagree — the textbook case of machinery papering over an unclear design.

## What changed

All per-job state now lives in a single `dict[JobId, _JobRecord]`. The record holds the job, its schedule, a lifecycle phase, a finishing overlay flag, the error/warning messages, and the primary-data flag. The wire-facing `JobState` is a derived property of the record, so phase, health, and reported status cannot drift apart. This removes the multi-bucket lookups and the manual tracking-cleanup helper.

`finishing` stays a boolean overlay rather than a phase: a finishing active job must still receive its final batch and compute before removal, which its retained phase determines.

The public API is unchanged. Deriving state also fixes a latent inconsistency: previously a push (accumulate) failure overwrote the reported state to `warning` while a finalize `error_message` was still set, so a job that could not produce results was shown as a warning. State now reflects the finalize error in that case — the failure that actually explains the missing output — in preference to the secondary push warning.

The change is split into two commits: a preparatory test refactor that decouples the run-transition tests from `JobManager` internals (so they no longer poke private maps), followed by the production refactor itself.

### DIFF
--- a/src/ess/livedata/core/job_manager.py
+++ b/src/ess/livedata/core/job_manager.py
@@ -704,17 +704,6 @@ class JobManager:
             status for status in self.get_all_job_statuses() if status.state == state
         ]
 
-    def format_job_error(self, status: JobReply) -> str:
-        """Format a job error message with meaningful job information."""
-        record = self._jobs.get(status.job_id)
-        if record is None:
-            return f"Job {status.job_id} error: {status.error_message}"
-
-        return (
-            f"Job {record.job.workflow_id}/{status.job_id.source_name} "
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
LGTM

PR #1022, opened 2026-06-29
Acknowledge a selector-keyed command only from the owning worker

SimonHeybrock:
## Why

Backend services share the `livedata_commands` topic, so every service sees every command. A `job_id`-keyed command is naturally filtered to the worker that owns the job: non-owning workers raise `KeyError` and stay silent. A **selector-keyed** command (`workflow_id`-keyed, or a broadcast with neither `job_id` nor `workflow_id`) has no such filter — it simply matches no jobs on a non-owning worker and no-ops *without* raising. That worker then falls through to the success branch and returns an `ACK` (`device="all"`) whenever a `message_id` is supplied. The result is one spurious acknowledgement per service rather than a single ack from the owner.

This is latent on `main` today — the dashboard always keys commands per `job_id` — but it makes `workflow_id`-keyed and broadcast commands ill-defined for any future producer that supplies a `message_id`.

## What

`JobManager.job_command` now returns the number of jobs it acted on, and `JobManagerAdapter` acknowledges only when that count is non-zero. The owning worker (the one that matched ≥1 job) becomes the sole responder; non-owning workers stay silent, mirroring the existing `KeyError` behaviour of the `job_id` path. Behaviour for `job_id`-keyed commands is unchanged.

## Test plan

Covered by unit tests (`job_manager_test.py`, `job_manager_adapter_test.py`): `job_command` returns the matched-job count (including zero for a non-owned workflow), and the adapter suppresses the ack on a zero-match selector-keyed command while still acking when a job matched.

### DIFF
--- a/src/ess/livedata/core/job_manager.py
+++ b/src/ess/livedata/core/job_manager.py
@@ -351,7 +351,14 @@ class JobManager:
             self._finishing_jobs.remove(job_id)
         logger.info("job_stopped", job_id=str(job_id))
 
-    def job_command(self, command: JobCommand) -> None:
+    def job_command(self, command: JobCommand) -> int:
+        """Apply a command and return the number of jobs it acted on.
+
+        The count is zero when no job on this worker matches the command's
+        selector. Backend services share the commands topic but each owns a
+        disjoint set of workflows, so a worker that matches nothing must stay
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1023, opened 2026-06-29
Suppress out-of-scope streams in backend stream stats

SimonHeybrock:
## Problem

The dashboard backend-status display showed a large, misleading list of "unmapped" streams for the `detector_data` and `data_reduction` services.

Kafka subscription is per-topic, but each service's stream mapping is scoped per-source (via `scope_stream_mapping`). Shared topics — most notably the f144 log topic, which carries the readbacks of *every* device — therefore deliver messages for sources a service does not handle. Those sources missed the scoped lookup table and were counted as "unmapped", even though they are perfectly well-known streams that are simply routed to a different service.

## Change

`StreamCounter` now takes the set of known-but-unsubscribed `(topic, source)` keys — the full mapping minus the scoped mapping — and drops them instead of recording them as unmapped. Genuinely unknown sources (schema drift, mistyped PVs) still surface as unmapped, so the signal that matters is preserved.

## Test plan

- [x] Unit tests for `StreamCounter` suppression and for the `StreamMapping.input_keys` difference that the services feed it.
- [x] `tests/kafka` and `route_derivation` suites pass.
- [ ] On a running backend, confirm the dashboard status display no longer lists the device readbacks under "unmapped" for `data_reduction` / `detector_data`.

### DIFF
--- a/src/ess/livedata/kafka/stream_counter.py
+++ b/src/ess/livedata/kafka/stream_counter.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
+from collections.abc import Iterable
 from dataclasses import dataclass
 
 from ess.livedata.core.job import (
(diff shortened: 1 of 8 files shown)


SimonHeybrock:
LGTM, should test last item in deployment (not high risk since this is only for reporting).

PR #1024, opened 2026-06-29
Require message_id on commands, simplify acknowledgement

SimonHeybrock:
Building on #1022, which made a command's acknowledgement depend only on whether the owning worker acted. The optional `message_id` was a second, redundant axis: whether the initiator wanted a reply at all. That decision already lives in the dashboard's pending-command tracker, so the optional wire field duplicated it and forced `if message_id is not None` branching through the adapter for both `JobCommand` and `WorkflowConfig`.

This gives `message_id` a `default_factory` so it is always populated, and collapses the adapter to a single rule: the owning worker acks (or errs), every other worker stays silent — via `KeyError` on the `job_id` path, or `affected == 0` on the selector path. "Fire-and-forget" stops being a wire concept: an initiator that tracks the reply sets the id; one that does not — the orchestrator stopping superseded jobs, or an external producer such as NICOS — omits it and an id is generated for the ack it ignores.

`FakeMessageSink` now routes acknowledgements to a separate `response_messages` view, keeping `messages` data-only. This preserves service-test assertions about emitted data, which previously relied on omitting `message_id` to suppress acks entirely; the routing is exercised implicitly by every service test, whose data-message counts would shift if an ack leaked into the data sink.

This is a prerequisite for the NICOS derived-devices work (ADR 0006): with acks uniform, the reset path no longer needs a special "omit message_id to avoid an ack storm" story. NICOS keeps omitting `message_id` exactly as the ADR describes, and confirms the reset from the `start_time` generation jump rather than the ack.

### DIFF
--- a/src/ess/livedata/config/workflow_spec.py
+++ b/src/ess/livedata/config/workflow_spec.py
@@ -461,12 +461,13 @@ class WorkflowConfig(BaseModel):
     identifier: WorkflowId = Field(
         description="Hash of the workflow, used to identify the workflow."
     )
-    message_id: str | None = Field(
-        default=None,
+    message_id: str = Field(
+        default_factory=lambda: str(uuid.uuid4()),
         description=(
-            "Transient identifier for command/response correlation. Frontend generates "
-            "this UUID and backend echoes it in CommandAcknowledgement responses. "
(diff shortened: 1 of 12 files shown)


SimonHeybrock (commented):
LGTM

Issue #1019, opened 2026-06-26
Add a shared glossary for recurring ESS/instrument jargon in the dashboard


SimonHeybrock:
Users with limited ESS/instrument background hit unexplained jargon throughout the dashboard — TOA, Q / momentum transfer, "cut axis", wavelength edges, "window" vs reduction cadence, pixel weighting. Individual descriptions help, but the same terms recur across workflows, params, and outputs, and explaining each one in full at every site is repetitive and inconsistent.

We should add a small, central glossary of recurring terms, defined once, and link to it from the places where the terms appear (workflow-config modal, plot-creation wizard, parameter tooltips).

This is a follow-up to the broader description/discoverability survey. The per-site description improvements (param↔output back-references, TOA default explanation, friendly enum labels, filling missing descriptions) are handled separately; this issue is specifically about the shared glossary so jargon is defined in one place rather than duplicated.

Open questions for design:
- Where the glossary lives (a doc page, an in-app panel, or hover-cards).
- How terms are linked from modals/tooltips without cluttering the UI.
- Initial term set: TOA, Q (momentum transfer), wavelength edges/range, time window vs reduction cadence, pixel weighting, ROI.

Issue #993, opened 2026-06-22
Bifrost notes


SimonHeybrock:
- ~Explain 0-71 and meaning of TOA in descriptions.~
- ~Monitor workflow params explain what output~
- Do we need manual unwrapping controls?
- ~Window -> Time Window~
- Rename Bragg-peak-monitor Q-map -> Elastic Q-map
- Monitor templates broken?
- ~Unified detector view shows only 5 rows (arcs), data width exceeds grid width. Is the reshape/fold/flatten wrong? Did this break in a recent file update?~

SimonHeybrock:
Closing — the remaining items are either resolved or tracked elsewhere.

**Monitor templates broken?** Not broken. Verified end-to-end on main rather than by inspection: all three bifrost grid templates parse through the real `PlotOrchestrator` path with every cell retained and no silent fallback to plotter defaults, all three appear in the Manage Plots dropdown, and applying "Monitors" creates the tab and renders all five monitors. #1026 had already reached this conclusion and added `test_shipped_template_data_sources_resolve` as a guard against future drift. The `mode: window` + `window_duration_seconds: 0.0` in the first cell is the documented "latest update only" setting, not a stale value.

One artifact worth knowing if you eyeball this against the fake backend: the "latest" and "since start" cells render identically, because `fake_backend.py` differentiates `cumulative` and `current` only in their time coords and synthesizes the same magnitudes for both. It does not indicate a problem with the template, and would not show on real data.

**Rename Bragg-peak-monitor Q-map -> Elastic Q-map** belongs to #555, which is still open and carries the same TODO in its own description. There is nothing to rename on main — no Bragg code exists there, the workflow lives only on `bifrost-bragg-peak-monitor-qmap`. Worth flagging for whoever picks that up: "Elastic Q map" is already the title of the `elastic_qmap` unified-detector workflow, and workflow titles are not checked for uniqueness, so the rename needs a disambiguating title to avoid two identically-labelled entries in the workflow list.

Separately, 591151ebf renamed the cbm5 monitor `bragg_peak_monitor` -> `elastic_monitor` and deleted `grid_templates/bragg_peak_monitor.yaml` rather than migrating it. Bifrost consequently ships no template covering the `total_counts` / `total_in_range` monitor views — only `histogram`, three times. Happy to re-add an equivalent against one of the current monitors if that grid is missed.

**Do we need manual unwrapping controls?** Deliberately left unanswered here — please re-file if still wanted, so it does not get lost with this issue.

PR #1025, opened 2026-06-30
Improve workflow/param/output descriptions and discoverability

SimonHeybrock:
Make the relationship between workflow parameters and the outputs they shape explicit and visible, and fill scattered description gaps that left users guessing (see #993).

## Param ↔ output coupling

`OutputView` gains a `params` field naming the parameter groups that shape each output. `WorkflowSpec` resolves this both ways (lenient against params-model variants that omit some fields). It is surfaced as a back-reference in the plot-creation wizard ("Controlled by workflow parameters: ..."), a forward-reference in the workflow-config modal ("Affects outputs: ..."), and hover tooltips on the Workflows-card output chips.

## Examples

Added tooltips to workflow outputs:
Screenshot 2026-06-30 at 07 51 49

Description extended with note on affected outputs:
image

## Other improvements

- Explain the time-of-arrival defaults: TOA is time since the source pulse, the default range is one 14 Hz pulse period; add the `ESS_PULSE_PERIOD_MS` constant (rounded) replacing the bare `1000/14`.
- Friendly labels for `Literal`-valued params via Field `json_schema_extra` (e.g. coordinate mode shows "Time of arrival (TOA)" not "toa").
- Fill missing/weak descriptions: dummy "Panel 0", Bifrost ratemeter region and Q-map jargon, wavelength-LUT "Unit." fields.
- Rename the plot-config "Window" tab to "Time Window".

The shared glossary for recurring jargon is tracked as #1019.


SimonHeybrock:
LGTM -- probably not perfect, but better than before and easy to improve later.

PR #1026, opened 2026-06-30
Guard shipped grid templates against workflow/plotter schema drift

SimonHeybrock:
Investigating the "Monitor templates broken?" note in #993, I found the bifrost grid templates resolve cleanly on `main` — the schema changes that would have broken them were already fixed. The real problem is that nothing catches such breakage: shipped grid templates reference a `workflow_id`, `view_name`, and `plot_name`, but the load path (`PlotOrchestrator._parse_single_spec`) swallows mismatches and silently degrades to empty or wrong plots. The only existing guard, `test_shipped_template_workflow_ids_parse`, checks the `workflow_id` *string format* only.

This adds a test that joins the shipped templates to the real workflow specs and plotter registry. For every non-static data source in every shipped template it asserts: the `workflow_id` resolves, `source_names` is a subset of the workflow's sources, `view_name` is a declared output view, and that view's template data satisfies the chosen plotter's data requirements. Static-plotter overlay layers (sentinel `static/...` ids) are skipped.

It sits between the two existing tests and joins what neither does: `grid_template_test.py` parsed `workflow_id` strings without resolving them; `registered_workflow_specs_test.py::test_workflow_output_has_compatible_plotter` resolved real views to *any* plotter while ignoring templates. Each of the four past bifrost-template breakages (window-param rename, stale `mode: latest`, 3-part WorkflowId migration, detector-view layout) would now fail loudly. Coverage expands from bifrost/dummy (format only) to bifrost/dream/dummy (full resolution).

The templates themselves are not changed; they are not currently broken.

### DIFF
--- a/tests/config/grid_template_test.py
+++ b/tests/config/grid_template_test.py
@@ -6,6 +6,8 @@ import pydantic
 import pytest
 
 from ess.livedata.config.grid_template import GridSpec, load_raw_grid_templates
+from ess.livedata.config.instrument import instrument_registry
+from ess.livedata.config.instruments import available_instruments, get_config
 from ess.livedata.config.workflow_spec import WorkflowId
 from ess.livedata.dashboard.data_roles import PRIMARY
 from ess.livedata.dashboard.plot_orchestrator import (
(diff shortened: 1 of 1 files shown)


SimonHeybrock (commented):
LGTM

PR #1027, opened 2026-06-30
Add source offset option to wavelength-LUT workflow

SimonHeybrock:
Adds a `source` parameter to the wavelength-LUT workflow that shifts the chopper-cascade source reference along the beam (lab +Z) relative to the position loaded from the geometry file. A positive offset moves the source downstream (toward the choppers), a negative offset upstream. This lets operators correct or explore the source position without regenerating the geometry artifact.

The source reference is the moderator when present (BIFROST) and otherwise the `NXsource` (LOKI). With no moderator and a zero offset nothing is overridden, so essreduce's own `NXsource` lookup -- the long-standing default path -- is left untouched.

This implements only the source-offset half of the originally discussed extension; per-chopper disabling is deferred.

## Test plan

- [x] Unit tests for the `SourceOffset` model and for the offset shifting chopper distances (moderator and `NXsource` geometries).

### DIFF
--- a/src/ess/livedata/handlers/wavelength_lut_workflow.py
+++ b/src/ess/livedata/handlers/wavelength_lut_workflow.py
@@ -303,22 +303,16 @@ def _make_workflow(pipeline: sciline.Pipeline) -> StreamProcessorWorkflow:
     )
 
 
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
LGTM

PR #1028, opened 2026-06-30
Show plotter name in plot layer headers

### LINKED ISSUE #645: Plot header toolbars: Show plotter name?
Currently the plot toolbar does not show the plotter name. This is confusing in certain cases, in particular with ROIs where both the readback and request depend on the same data source and thus currently yield identical-looking header info. Another example is correlation plots, where the current info does not indicate that this is not plotting just the data itself, but uses the correlation plotter to plot it against some other timeseries.

SimonHeybrock:
Closes #645.

The per-layer plot header showed only workflow → output (source), never which plotter renders the layer. This made layers that differ *only* by plotter indistinguishable: ROI readback vs request share the same data source, and correlation plots gave no hint that they plot the data against another timeseries rather than plotting it directly.

`PlotterSpec.title` already existed but was never surfaced. It is now folded into `get_plot_cell_display_info` — the single label source shared by the expandable layer-detail rows, the gear dropdown menu, and the edit-cell modal — so the name appears in all three at once. The layer label becomes `Workflow → Output → Plotter (sources, window)`. The plotter segment is suppressed when it merely echoes the output name (e.g. an `Image` output rendered by the `Image` plotter), keeping the common case clean and showing the plotter only where it disambiguates (ROI readback/request, correlation).

## Test plan
- [x] Unit tests for plotter title shown, readback ≠ request, tooltip line, output-echo suppression, unregistered-plotter fallback.
- [x] Expand a cell's layer details on an ROI plot; confirm readback and request rows now read distinctly.
- [x] Confirm the gear menu and edit-cell modal list the same disambiguated labels.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/plot_widgets.py
+++ b/src/ess/livedata/dashboard/widgets/plot_widgets.py
@@ -579,7 +579,23 @@ def get_plot_cell_display_info(
         workflow_registry, config.workflow_id, config.view_name
     )
 
-    # Build title: "Workflow → Output (source, window)"
+    # Look up the plotter's display title so layers differing only by plotter
+    # (ROI readback vs request, or a correlation plot over a timeseries) are
+    # distinguishable instead of rendering identical workflow/output text.
+    from ..plotter_registry import plotter_registry
+
+    try:
+        plotter_spec = plotter_registry.get_spec(config.plot_name)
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
Works and LGTM

PR #1029, opened 2026-06-30
fix(dashboard): enforce frame aspect per sub-figure so layout-mode plots don't collapse

SimonHeybrock:
## Problem

Layout-mode plots render with their vertical axis collapsed to zero height — only the toolbar, x-axis, and colorbar remain. The dummy detector images (`square` aspect + `combine_mode: layout`) hit this. First noticed in `scripts/drive_dashboard.py` screenshots, but it is a real rendering bug in the live DOM, not a screenshot artifact.

## Cause

The frame-aspect hook (`frame_aspect.py`) switches a Bokeh figure to a one-axis responsive mode (`stretch_width`/`stretch_height`) and pins the other dimension via JS — this is how non-free aspects are enforced. It was applied once, at the cell level, to the cell's single composed figure, and skipped entirely for non-overlayable layers. But a `Layout` has one figure *per sub-plot*, so those figures never received the hook: they stayed `stretch_both` while the cell's pane asked for `stretch_width`, leaving the free (vertical) axis unconstrained and collapsing the plot.

This surfaced when layout mode began always producing an `hv.Layout` (PR #1008's config-time `is_overlayable` invariant): single-source layout cells became `Layout`s and lost the hook. PR #1008 merged onto `main` after the most recent release (26.6.4, 2026-06-23) and is not contained in any tag, so this regression never reached a public release — it exists only on `main`.

## Fix

Declare the frame-aspect hook **per leaf element type in the plotter**, alongside `responsive=True`, so it lands on every figure the plotter produces — one shared figure when overlaid, one per sub-plot when laid out. Aspect becomes a per-figure property that survives composition. This fixes the collapse, keeps square detector images square, and now also honors aspect for multi-source layouts (previously it was silently ignored there). The plotter already received `aspect_params` but never used them.

PR #1008's design is untouched: layout mode still always produces a `Layout`, and the cell-sharing forbiddance still uses the config-time `is_overlayable`. The cell widget no longer applies a composed-level aspect hook; it still skips the genuinely cell-level SaveTool/autoscale hooks for `Layout`s and `Table`s. `No data`/error placeholders keep plain `responsive` (no forced aspect).

## Test plan

- [x] New real-component test renders a multi-source layout-mode plotter and asserts every sub-figure switches to the one-axis responsive mode (fails before the fix).
- [x] Updated the figure-level aspect test to reflect that the plotter now applies the hook.
- [x] Full dashboard test suite passes (1610).
- [x] Verified via `drive_dashboard.py`: detector images render square with full vertical axes (iw == ih).

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -20,6 +20,7 @@ from ess.livedata.config.workflow_spec import ResultKey
 from ess.livedata.core.timestamp import Timestamp
 
 from .data_roles import PRIMARY
+from .frame_aspect import make_frame_aspect_hook_from_config
 from .plot_params import (
     CombineMode,
     LayoutParams,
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
Confirmed also with real dashboard and DREAM detector image in layout mode. LGTM.

PR #1030, opened 2026-06-30
tests: guard against collapsed/inconsistently-sized plots

SimonHeybrock:
The Option 3 fix (#1029) is now on main. The collapsed-detector-image bug was a close call: it passed every functional test because the data and figure models were correct — only the *rendered geometry* was wrong. These add deterministic guards for that whole class.

## Two complementary guards

**Headless invariant** (`plot_sizing_invariant_test.py`, runs in CI). Renders each plotter across aspect × combine-mode × source-count and asserts every figure — the overlay's shared one *and each layout sub-figure* — adopts the responsive mode the cell's pane will wrap it in. It asserts against the real `cell._get_sizing_mode`, so figure sizing and pane sizing are tied together: a future divergence (exactly what caused the collapse) fails here. Catches the bug at the cause, fast, no new deps.

**Browser smoke test** (`render_geometry_test.py`, local-only). Drives the Kafka-free fake dashboard with Playwright, visits every plot-grid tab, and asserts no figure's data area is degenerate (`inner_width`/`inner_height`). This is the fix-agnostic ground truth — it fails whenever a plot renders collapsed, whatever the cause (verified: it fails on buggy `main` with `ih: 0`, passes on the fix). Coverage scales with fixture coverage, so it is the natural home for a future kitchen-sink fixture.

## Marker

Added a `browser` marker, excluded from the default run and the three CI tox envs. Playwright-in-CI is a separate decision; until then the browser test runs locally via `pytest -m browser` (it `importorskip`s Playwright, so it skips cleanly where the dep is absent). The headless invariant is the CI-gating guard in the meantime.

## Also

`cell._get_sizing_mode` now takes `params` directly (it only ever used `config.params`) so the invariant test can call it without fabricating a `PlotConfig`.

## Test plan

- [x] Headless invariant: 40 cases pass on the fix; fails on `main` (hook was applied at the cell level, so plotter figures weren't sized by aspect).
- [x] Browser test: passes on the fix; fails on buggy `main` (`figures collapsed below 50px ... ih: 0`).
- [x] Full dashboard suite green (1650); browser test deselected by default and by the CI filter.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ addopts = """
 --strict-markers
 --import-mode=importlib
 --benchmark-skip
--m "not integration and not slow"
+-m "not integration and not slow and not browser"
 -ra
 """
 testpaths = "tests"
@@ -112,6 +112,7 @@ markers = [
   "instrument: Specify instrument name for parameterized tests",
   "services: Specify which services to use in integration_env fixture (monitor, detector, or reduction)",
   "slow: Slow tests (>2s) excluded by default, run in CI via tox",
+  "browser: Browser-driven (Playwright) UI tests; run locally, excluded from CI",
 ]
 filterwarnings = [
   "error",
--- a/src/ess/livedata/dashboard/widgets/cell.py
+++ b/src/ess/livedata/dashboard/widgets/cell.py
@@ -21,6 +21,7 @@ from dataclasses import dataclass
 
 import holoviews as hv
 import panel as pn
+import pydantic
 import structlog
 
 from ess.livedata.config.workflow_spec import WorkflowId, WorkflowSpec
@@ -32,7 +33,6 @@ from ..plot_orchestrator import (
     CellId,
     LayerId,
     PlotCell,
-    PlotConfig,
     PlotOrchestrator,
 )
 from ..plot_params import PlotAspectType, StretchMode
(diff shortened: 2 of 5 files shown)


SimonHeybrock:
Heads-up: #1053 adds a second `browser`-marked test (multi-session smoke via a new `Dashboard.connect_many`) and needed the same marker, so it mirrors this PR's `pyproject.toml`/`tox.ini` changes verbatim — identical lines, so the branches merge cleanly in either order and the marker definition stays single-sourced in effect.

Two coordination notes:

1. `render_geometry_test.py` here and `multisession_smoke_test.py` there both follow the pattern *importorskip playwright → `_fake_dashboard` → drive → assert on rendered state*, and both currently launch their own server (ports 5031/5032). If the `browser` suite grows further, a shared module/session-scoped fake-dashboard fixture in a `tests/dashboard/conftest.py` (gated on the marker) would amortize the ~30 s launch.
2. On the deferred Playwright-in-CI decision: with two behavioral guards now marked `browser`, the ingredients for a separate opt-in CI job would be (a) pinning `playwright` in a requirements file (it is currently devcontainer-only — `README_screenshots.md` claims it's in dev dependencies but no `requirements/*.txt` lists it), (b) `playwright install chromium --with-deps` in the workflow, (c) `pytest -m browser`. Runtime today would be ~2–3 min. Still a separate decision, but it no longer needs design work.

---

SimonHeybrock:
Superseded by #1150. This branch predates the browser-test framework: the `browser` marker and `tox.ini` filters here landed on main via #1053, main now runs `tox -e browser` in CI, and `tests/helpers/browser.py` owns the launch/port plumbing these tests rolled by hand. The headless invariant had also gone silently vacuous under the `ResultKey` → `DataKey` migration.

Opened as a new branch rather than force-pushing this one, so the original stays available for comparison — happy to close this once #1150 looks right.

PR #1031, opened 2026-06-30
dashboard: manual color limits (vmin/vmax) for 2D plots

SimonHeybrock:
## Motivation

2D plots (Image, Slicer, Flatten, 2D correlation) offered only a linear/log color scale plus an interactive color-autoscale toggle. There was no way to fix the colorbar to explicit values — useful for comparing frames on a common scale or suppressing a dominating hotspot.

## What this adds

A **Manual Color Limits** option in the 2D color-scale section of the plot config modal, with **Color Min** / **Color Max** inputs. When enabled, the fixed limits are applied to the plot and the per-cell color-autoscale toolbar button is removed, so runtime autoscaling cannot fight the user's chosen values. X/Y autoscale is unaffected.

Manual limits are validated when the config is submitted: Color Max must exceed Color Min, and Color Min must be positive on a log color scale.

## Example

image

## Design note

Effective autoscalable axes are now exposed as an instance-level `autoscale_axes` property layered over the static `AUTOSCALE_AXES` capability, letting a plotter drop the color axis at construction. This also lets the correlation plotter genuinely delegate its autoscale axes to its inner renderer, which its docstring already claimed but did not do.

## Test plan

- [x] Open a 2D plot's config, enable Manual Color Limits, set min/max, and confirm the colorbar is pinned and the color-autoscale toolbar button is gone.
- [x] Confirm X/Y autoscale toggles still work with manual color limits on.
- [x] Confirm submitting Color Max ≤ Color Min, or a non-positive Color Min on a log scale, is rejected.

### DIFF
--- a/src/ess/livedata/dashboard/cell_autoscale.py
+++ b/src/ess/livedata/dashboard/cell_autoscale.py
@@ -101,7 +101,7 @@ class CellAutoscaleController:
     def __init__(self, layer_plotters: list[Plotter]) -> None:
         self._plotters: list[Plotter] = list(layer_plotters)
         self._axes: frozenset[Axis] = frozenset().union(
-            *(plotter.AUTOSCALE_AXES for plotter in self._plotters)
+            *(plotter.autoscale_axes for plotter in self._plotters)
         )
         # Lazy-created on first render so each session's tools live in the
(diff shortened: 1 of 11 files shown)


SimonHeybrock:
Tested with DREAM, works and LGTM.

PR #1032, opened 2026-06-30
dashboard: enable hover tool for data plotters

SimonHeybrock:
## Motivation

Plots had no hover tool, so users could not read off coordinate/value at a point. This enables hover broadly across the data plotters.

## What

Hover is now active on line, image, slicer, overlay and bar plots, and on correlation plots via their line/image renderers. Tooltips carry axis labels with units and, for images, the cell value.

Error elements (`ErrorBars`, `Spread`) are deliberately excluded: Bokeh raises when a `HoverTool` is attached to the Whisker glyph, and the whisker/band values merely duplicate the line they annotate.

Bar charts previously kept pan/zoom/reset active behind a hidden toolbar, leaving zoom usable but reset unreachable. They now expose only an always-active hover.

`FlattenPlotter` keeps its bespoke axis-decomposing hover unchanged.

## Examples

Screenshot 2026-06-30 at 15 21 37
Screenshot 2026-06-30 at 15 21 51

## Test plan

- [x] Hover over a line plot (with error bars and band) shows the coordinate and value, with no hover on the error whiskers/band.
- [x] Hover over an image / slicer shows x, y and the cell value.
- [x] Hover over a bar chart shows the value; pan/zoom/reset are gone.
- [ ] Flatten plot hover still decomposes the flattened axes.

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -878,6 +878,12 @@ _LINE1D_LEAF_ELEMENTS: tuple[type, ...] = (
     hv.ErrorBars,
     hv.Spread,
 )
+# Element types that get a hover tool. Error elements (ErrorBars, Spread) are
+# deliberately excluded: Bokeh cannot attach a HoverTool to the Whisker glyph
+# (it raises during rendering), and the whisker/band values merely duplicate the
+# line they annotate.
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
Works and LGTM

PR #1033, opened 2026-07-01
Drop streams on unauthorized Kafka topics in PROD

SimonHeybrock:
## Summary
- PROD is currently missing topic ACL grants for some Kafka topics referenced by `<instrument>/streams_parsed.py`, so instruments would fail to subscribe to them.
- Adds a global (non per-instrument) allow-list — `<instr>_choppers`, `<instr>_motion`, `<instr>_sample_env`, and `tn_data_general` — and drops any `streams_parsed.py` entries on topics outside it before stream naming.
- Offending topics currently found: `dream_nicos_devices`, `loki_misc_devices`, `loki_nicos_devices`, `odin_misc_devices`, `odin_nicos_devices`, `tbl_misc_devices`, `tbl_nicos_devices`.
- This is a temporary workaround, to be reverted once PROD ACLs cover the full topic set.

## Test plan
Full unit test suite and ruff pass locally with this change.

### DIFF
--- a/src/ess/livedata/config/__init__.py
+++ b/src/ess/livedata/config/__init__.py
@@ -2,7 +2,14 @@
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 
 from .instrument import Instrument, SourceMetadata, instrument_registry
-from .stream import ContextBinding, Device, F144Stream, Stream, name_streams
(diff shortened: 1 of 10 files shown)


SimonHeybrock (commented):
LGTM

PR #1034, opened 2026-07-01
Cap inactive backend worker entries per kind

SimonHeybrock:
## Summary
- A crash-looping backend service mints a fresh `worker_id` (UUID) on every process restart, so `ServiceRegistry` accumulated a permanent new entry per restart, only ever cleared via the manual "Clear stopped workers" button — flooding the Backend Workers widget.
- `ServiceRegistry.status_updated` now auto-prunes the oldest `stopped`/stale entries once a given `(instrument, service_name)` kind exceeds a small cap (default 3), keeping the most recently seen ones. Active (`running`/`starting`) workers are never pruned automatically.

## Test plan
- [x] Added unit tests in `tests/dashboard/service_registry_test.py` covering oldest-first pruning under a restart loop, active workers being exempt, and per-kind/per-instrument scoping.


SimonHeybrock (commented):
LGTM

PR #1035, opened 2026-07-01
Bifrost: Fix Detector Ratemeter rate display via current/cumulative outputs

SimonHeybrock:
## Problem

Enabling a plot's Rate option ("counts per second") on the Bifrost **Detector Ratemeter** left the value in raw counts and the column header showing `counts` rather than `counts/s`. Regular detector/monitor "total" outputs were unaffected.

The ratemeter emitted a single `counts` output and hand-stamped only a scalar `time` coord. That coord caused `_add_time_coords` to skip `start_time`/`end_time`, so `_normalize_to_rate` found no duration to divide by and silently returned the data unchanged — no division, unit stayed `counts`.

## Fix

Restructure the ratemeter to the same current+cumulative output pattern already used by the monitor and detector-view workflows. A bare per-update provider feeds an accumulation-mode-scoped output accumulated via `make_no_copy_accumulator_pair`. The current output is a `window_output`, so `StreamProcessorWorkflow` stamps `time`/`start_time`/`end_time` from the update window; the cumulative output receives the job span via `_add_time_coords`. Both outputs stay `counts`, so the plot Rate option divides each by the appropriate duration (window vs. whole run) to give the correct `counts/s`.

An `OutputView` pairs the two streams under the existing view name, so the shipped grid template cells resolve unchanged and gain the standard "since run start" / "latest update" toggle. The cumulative count resets only on a run transition.

## Test plan

- [x] Unit + slow tests pass; new integration test drives the real `StreamProcessorWorkflow` and confirms the current output normalizes to `counts/s`, cumulative defers its time bounds to the job layer
- [x] Grid-template and workflow-spec validation pass
- [x] Bifrost dashboard: open the Detector Ratemeter, enable the plot Rate option, confirm the Table header reads `counts/s` and the value is divided by the window duration
- [x] Confirm the bars and timeseries cells still render as before

### DIFF
--- a/src/ess/livedata/config/instruments/bifrost/factories.py
+++ b/src/ess/livedata/config/instruments/bifrost/factories.py
@@ -60,7 +60,8 @@ def setup_factories(instrument: Instrument) -> None:
     )
     from scippnexus import NXdetector
 
-    from ess.livedata.handlers.accumulators import LatestValue
+    from ess.livedata.handlers.accumulation_mode import Cumulative, Current
+    from ess.livedata.handlers.accumulators import make_no_copy_accumulator_pair
     from ess.livedata.handlers.stream_processor_workflow import StreamProcessorWorkflow
 
     from .streams import detector_number
@@ -90,9 +91,8 @@ def setup_factories(instrument: Instrument) -> None:
     (
         reduction_workflow,
         DetectorRegionCounts,
-        detector_ratemeter,
+        _DetectorRegionCountsRaw,
     ) = _create_base_reduction_workflow()
-    reduction_workflow.insert(detector_ratemeter)
 
     @specs.detector_ratemeter_handle.attach_factory()
     def _detector_ratemeter_workflow(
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
LGTM and works locally

src/ess/livedata/config/instruments/bifrost/factories.py
```diff
@@ -311,21 +322,44 @@ def _create_base_reduction_workflow():
...
-        counts.coords['time'] = time
         counts.variances = counts.values
-        return DetectorRegionCounts(counts)
+        return DetectorRegionCountsRaw(counts)
+
+    def _accumulated_region_counts(
+        counts: DetectorRegionCountsRaw,
+    ) -> DetectorRegionCounts[AccumulationMode]:
+        """Route per-update counts to the accumulation-mode-specific accumulator."""
+        return DetectorRegionCounts[AccumulationMode](counts)
```
jl-wynen, line 362: This looks like you always return the same `DetectorRegionCounts` regardless of `AccumulationMode`. is this intentional?

SimonHeybrock: Yes, essentially this just sets up two distinct nodes in the graph. Then, a distinct accumulator is attached to each node -- see args passed to `StreamProcessorWorkflow` above.

jl-wynen (approved):

Issue #1040, opened 2026-07-02
PlotOrchestrator shared state has no thread-safety discipline (background thread vs UI; cross-session widget push)


SimonHeybrock:
## Summary

`PlotOrchestrator` holds the shared grid/cell/layer topology (`_grids`, `_cell_to_grid`, `_layer_to_cell`) for **all** browser sessions, but has no locking or thread-confinement discipline over that state. Two latent concurrency hazards exist. Neither is known to cause failures in practice today, but both are real races that can surface as intermittent `KeyError`/`RuntimeError` or cross-session UI corruption under load.

This was found while fixing the grid-edit crash on overlapping cells (see the linked PR). The overlap fix itself is race-free — all UI callbacks are serialized on a single IOLoop — so these are tracked separately.

## Threading model (for context)

The app is served by a plain `pn.serve(...)` with no `threaded=`/`num_procs=` (`dashboard.py`), so every UI event callback and every per-session periodic callback runs on a **single Tornado/Bokeh IOLoop thread**, under each session's document lock. UI mutations of the orchestrator are therefore serialized. But `DashboardServices` also spawns a separate `orchestrator-update` daemon thread (`dashboard_services.py`) running `orchestrator.update()` + `plot_orchestrator.flush_frames()` every ~50ms — a genuine second thread.

## Hazard 1 — background thread vs UI thread on unguarded state

`PlotOrchestrator` has zero locking over `_grids` / `_cell_to_grid` / `_layer_to_cell` (`plot_orchestrator.py`). During ingestion the update thread dereferences those mappings (`_enqueue_compute` → `_grid_of_layer`) while UI callbacks `remove_grid` / `replace_grid` delete or reassign them on the IOLoop thread. The GIL keeps individual dict operations from corrupting memory, and there are ad-hoc membership guards, but multi-step invariants spanning the three dicts can be observed mid-mutation (e.g. a layer mapping is gone while its grid mapping still exists). `DataService` even carries a comment acknowledging the UI thread mutates `_subscribers` mid-iteration.

## Hazard 2 — synchronous, lock-crossing cross-session widget invocation

[...]

PR #1041, opened 2026-07-02
Reject overlapping cells when building plot grids

SimonHeybrock:
## Motivation

A downloaded/hand-edited grid config (`esslivedata_loki_*.yaml`) could not be edited: clicking the pencil showed the title and row/column counts, but the preview stayed blank and only "Add Grid" appeared — no Save/Copy. The grid had two cells claiming the same slot. Building the edit preview feeds each cell into a Panel `GridSpec`; an overlapping region that still contains a not-yet-placed slot makes Panel dereference the empty slot and raise, and the exception aborts edit mode before the Save/Copy buttons are shown.

## Root cause

Grid cells must tile without overlap, but only the click-to-place editor enforced that (via its own occupancy tracking). Every other path into the orchestrator — file upload, persisted-config load, templates, edit-save — went through `add_cell` with no validation, so an overlapping layout could enter freely. It renders in the live view (cells overwrite last-wins) but is uneditable.

## Change

Enforce the no-overlap invariant in the model rather than papering over it in the preview: `add_cell` raises on overlap, and the collection-level entry points validate up front. Validating a whole cell set before any of it is applied is what makes the outcome clean — uploads are refused with a user-facing error before anything is staged, and a faulty persisted grid is skipped on load (logged) without leaving a half-built grid behind or blocking the remaining tabs.

## Follow-up

While investigating I mapped the orchestrator's threading model. The overlap fix is race-free (all UI callbacks are serialized on a single IOLoop), but I found two pre-existing, unrelated concurrency hazards in `PlotOrchestrator` shared state — tracked in #1040, out of scope here.

## Test plan

The manual items are now automated — the browser harness landed after this PR was opened.

- [x] Unit: `add_cell` rejects overlap and allows adjacent cells; `reject_overlapping_cells` accepts disjoint / rejects overlap; a corrupt persisted grid is dropped on load while the intact ones still load; an overlapping upload is rejected.
- [x] Browser (`pytest -m browser`): uploading an overlapping grid config in Manage Plots is refused with a user-facing error, stages nothing, creates no tab and logs no server exception; re-uploading the same grid tiled imports it and the new tab goes live.

The symptom was re-verified against current `main` before rebasing: an overlapping config still imports, and the pencil still aborts edit mode with a `TypeError` from Panel's `GridSpec` and no Save/Copy buttons. Note it needs a *partial* overlap ordered ahead of the cell owning the slots it grows into — an exact duplicate is equally invalid but only repaints occupied slots and renders without raising. The browser test uses the shape that actually breaks.


SimonHeybrock (commented):
LGTM

PR #1043, opened 2026-07-02
Pause DataService delivery for plot layers without viewers

SimonHeybrock:
Stacked on #1009.

## Motivation

Data subscriptions in the dashboard are process-global: one `DataSubscriber` per plot layer, shared by all sessions. On every `DataService` update, extract + copy + assemble + stash ran for every subscribed layer on every tab — only `plotter.compute` was gated on viewer interest. A dashboard with many tabs paid delivery costs for all of them even with nobody looking, on the ingestion thread and ahead of visible-tab frame flushes, and each hidden layer pinned a copied, assembled input dict in its stash indefinitely.

## What changed

The subscription's two roles are now split. Buffering (extractors drive buffer type and history retention) keeps running for hidden layers, so a timeseries is complete when its tab is opened. Delivery is gated per subscriber: `DataService` consults a new `active` property before extracting, and `DataSubscriber` derives it from the layer gate's viewer-token count — so tokens released by the weakref finalizer pause delivery too, with no mirrored flag that could go stale.

On the 0→1 viewer transition the layer is refreshed synchronously from a `DataService` snapshot through the subscriber's own extractors, reusing the synchronous `_run_compute` path. This preserves the same-poll-pass `has_cached_state` contract for tab switches and keeps `_frame_buckets` ingestion-thread-only. `LayerStateMachine.set_active` accordingly reports the transition as a bool instead of flushing the (now stale) stash; the stash remains as the handoff for the active delivery path.

## Performance impact

Measured per-delivery cost (extract + copy per layer per update, devcontainer CPU, realistic shapes):

| Layer type | per update |
| --- | --- |
| 0D latest value | 0.005 ms |
| Timeseries full history (10k–100k points) | ~0.25 ms |
| Image latest value (128²–1024²) | 0.04–0.18 ms |
| Image with time-window aggregation (256², ~10-frame window) | ~4 ms |

Plain copies are memcpy-bound and nearly free; the dominant cost is `WindowAggregatingExtractor`, which reduces over the buffered window on every delivery and scales with window length × image size. For a typical deployment (5–10 tabs × 5–10 layers of mixed types, ~1 Hz updates, 1–3 sessions each viewing one tab, so 70–90% of layers hidden) this saves roughly 20–80 ms/s of ingestion-thread CPU; windowed-image-heavy configurations or higher update rates scale that to hundreds of ms/s. Session count itself is a non-factor since subscriptions are process-global.

Beyond CPU: delivery ran serially on the ingestion thread ahead of `flush_frames`, so every visible-tab frame previously waited on all hidden tabs' delivery — that per-burst latency is removed from the plots someone is actually watching. Each hidden layer also pinned one assembled, copied input (0.1–8 MB for images) in its stash indefinitely; the stash is now only populated transiently.

## Test plan

- Unit tests cover paused delivery, snapshot refresh, buffering-while-paused, gate transitions, and an end-to-end orchestrator flow (hidden layer skips compute, activation computes the latest value, releasing the last token pauses again).
- Verified in the running dashboard (fake transport, Playwright): a tab hidden since startup receives zero deliveries and fills immediately on first activation; after a hidden dwell it shows fresh data (moved peak, grown counts); visible tabs update live; rapid tab churn produces no notification or compute errors in the server log.
- [ ] Manual check against a real backend: switch between tabs with active workflows in 2+ browser sessions; hidden tabs must fill immediately on switch and timeseries must cover the hidden period.

### DIFF
--- a/docs/developer/adr/0005-frame-gated-session-flush.md
+++ b/docs/developer/adr/0005-frame-gated-session-flush.md
@@ -1,6 +1,6 @@
 # ADR 0005: Frame-gated per-session plot flush
 
-- Status: accepted
+- Status: accepted (amended 2026-07-02, see bottom)
 - Deciders: Simon
 - Date: 2026-06-25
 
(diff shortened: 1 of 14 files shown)


SimonHeybrock:
Merging into base branch since it is part of a stack.

Issue #1044, opened 2026-07-02
Audit results: dashboard data-delivery waste and bottlenecks in long-running operation


SimonHeybrock:
Results from audit, to be triaged:

Follow-up from the `pause-inactive-delivery` branch (viewer gate for DataService delivery): an audit of the wider dashboard data mechanism for compute/memory waste and architectural bottlenecks in long-running (multi-day) operation.

Overall: memory is mostly bounded by design (20 MB/key buffers with ring-buffer trim, job cleanup retains only current+previous job per workflow). But there is one real leak that defeats the viewer gate, one retention ratchet, and a category of per-update compute waste the gate only partially addresses.

## Leak: stale-session reaper skips `PlotGridTabs.shutdown()`

`shutdown()` (releases viewer tokens, unsubscribes lifecycle) is wired only to `pn.state.on_session_destroyed` (`reduction.py:187`). The stale-session reaper (`session_registry.py:149`) only calls `SessionUpdater.cleanup()`. On a missed destroy event (browser crash, network drop -- routine over multi-day ops), `PlotOrchestrator._lifecycle_subscribers` keeps the dead session's `PlotGridTabs` and its layers strongly reachable, so the token `weakref.finalize` fallback (`plot_data_service.py:301`) can never fire. Result: those layers hold viewer tokens forever, `has_viewers` stays True, and the delivery gate is permanently open for a session nobody is watching. Compute + memory until process restart. Fix is cheap: register `shutdown` as a cleanup handler so the reaper path runs it too.

## Ratchet: extractors are added, never removed

`TemporalBufferManager` has no `remove_extractor`; `set_extractors` has zero production callers; `DataService.unregister_subscriber` (`data_service.py:255`) only drops the subscriber. Per key: every plot ever opened appends an extractor (`temporal_buffer_manager.py:135`, no dedup), retention timespan only grows (one full-history plot ever pins 20 MB retention for the key's remaining life), and the Temporal-to-SingleValue downgrade is dead code in practice. For a multi-day job with users opening/closing plots, the extractor list on hot keys grows without bound (objects are small; the retained history is the real cost). `unregister_subscriber` calling `set_extractors` with the surviving subscribers' extractors is the missing symmetry.

## Compute waste on the visible path (per update, forever)

The viewer gate pauses hidden layers; visible layers still pay all of this:

[...]

PR #1045, opened 2026-07-02
Invert DataService delivery to pull-on-frame

SimonHeybrock:
Closes the main architectural item from #1044 (see the evaluation comments there for the full analysis). Stacked on #1043; targets `pause-inactive-delivery`.

## What

Inverts the DataService hot path from push-per-delta to pull-on-frame. Ingestion now only marks keys updated and notifies subscribers keys-only; the plot orchestrator marks affected layers dirty per grid, and `flush_frames` pulls each dirty *viewed* layer's input through its extractors from current buffer state, rebuilds, and commits the grid frame. The 0→1 viewer transition still rebuilds synchronously on the polling thread, as introduced in #1043.

## Why

Extraction (dominated by window aggregation over buffered history for image-sized data) previously ran per subscriber per ingestion batch, on the notification path, and its results traveled through per-layer stashes — which is what the viewer gate in #1043 had to work around. Pulling at flush time makes "no viewers → no work" the default instead of a gated special case, and removes the machinery the gate required: the `DataServiceSubscriber.active` gate, the `LayerStateMachine` stash/`ComputeTask` plumbing, and the documented delivery races (out-of-order data delivery, in-flight delivery racing a pause) are all gone. Net −100 lines.

Because a pull always reads current buffer state, coalescing or deferring flushes is lossless — a future min-frame-interval throttle in `flush_frames` (to cap compute at display cadence under ingestion load) becomes a one-line policy choice rather than a correctness question. Deliberately not included here; cadence today is identical to before.

The frame-gated session flush from ADR 0005 is unchanged: per-grid commits, generation-gated session flushes, and the threading constraint all hold. The ADR gains an amendment describing the input-side change (frame = buffer state at flush, rather than the drained burst's payload).

Also fixes static overlays never rendering when created without a viewer token, a gap the gate in #1043 introduced: they are now built unconditionally at creation, since they have no data subscription to pull from later and the build is a one-off.

## Test plan

Verified end-to-end against the fake backend via `scripts/drive_dashboard.py`: detector images fill on the visible tab and keep updating (peak moves between frames), tab-away/tab-back shows fresh data promptly (activation refresh), and rapid tab flipping does not wedge delivery.

Needs a check against a real timestamped backend:

- [x] Live stream: plots update at the usual cadence, freshness pill sits near pipeline lag.
- [x] Multiple visible plots repaint together (no stagger).
- [x] Tab switch shows latest data promptly; hidden tabs cost no compute.

### DIFF
--- a/docs/developer/adr/0005-frame-gated-session-flush.md
+++ b/docs/developer/adr/0005-frame-gated-session-flush.md
@@ -1,6 +1,6 @@
 # ADR 0005: Frame-gated per-session plot flush
 
-- Status: accepted
+- Status: accepted (amended 2026-07-02, see bottom)
 - Deciders: Simon
 - Date: 2026-06-25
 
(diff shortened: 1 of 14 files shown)


SimonHeybrock:
Note from the perf follow-up (#1055, detailed proposal in [#1046 comment](https://github.com/scipp/esslivedata/issues/1046#issuecomment-4897619291)), no change requested here: once delivery is pull-on-frame, `flush_frames`' per-grid `commit` becomes the single point where "a new displayable frame exists" is known — which makes it the natural site to *wake* sessions with a data-free `doc.add_next_tick_callback` tick, instead of sessions detecting the generation advance by 100 ms polling. Feasibility is verified (panel 1.9.3: woken tick runs in correct session context, hold/freeze work; see #1046). Two composition points worth keeping in mind while this branch evolves: the min-frame-interval throttle anticipated in the description and the wake compose cleanly (throttle caps *compute* cadence, wake removes *display-side* polling), and `commit(grid_id)` staying the last step of a grid's flush is what makes a wake-on-commit never observe a half-built frame — already the case here, just worth preserving.

---

SimonHeybrock:
LGTM, merging into base branch since we undo some changes there, to allow for clean combined review.

Issue #1046, opened 2026-07-05
Dashboard: converge on single-writer state + versioned pull (root cause behind the audit bug clusters)


SimonHeybrock:
## Summary

The recent audits (#1039, #1044, plus standalone issues #1040, #955, and the assessment in #1042) are not independent piles of bugs. Sorting the dashboard findings by *where the buggy code lives* rather than by symptom, almost every confirmed race, leak, and waste item sits in one kind of code: **synchronous push machinery — callbacks that fan out across threads and sessions carrying data or lifecycle events**, with thread-safety retrofitted per call-site (the `ingestion_guard` covers only the commit path; every entry point added later — add-layer-while-running, heartbeat deactivation, the stale-session reaper — bypassed it).

Conversely, the parts of the dashboard the audits found *clean* (DataService transactions, version-polling widgets per `.claude/rules/dashboard-widgets.md`, the notification queue) are exactly the parts built on the opposite model: **single authoritative state + version counter + every consumer pulls on its own clock**. And #1007 (subscriber skipped when list mutated during notification) was the same disease already fixed once.

This issue tracks converging the remaining three push planes onto that model, as one sequenced program rather than three unrelated refactors. A draft ADR making the model explicit is linked below.

### Plane 1 — data delivery (`DataService` push-per-delta)

Every ingestion burst eagerly extracts, copies, and computes for every subscriber, on the ingestion thread, under the service lock.

Direct consequences: the compute-waste list in #1044 (ingestion outpacing display, all-keys re-extraction, unconditional copies, extraction under the lock), the extractor-retention ratchet, the need for the viewer gate in #1043 and its documented races, and #1039 findings 2 (register-while-running races ingestion) and 9 (compute-gate stale overwrite).

Fix: **#1045 (pull-on-frame inversion)** — ingestion only marks keys dirty; the frame flush pulls for dirty+viewed layers. Per the analysis in #1044, this *deletes* the gate, the stash, and the documented races rather than guarding them, and shrinks what the #1009 lock must protect to buffer writes.

### Plane 2 — job/lifecycle notification (`JobOrchestrator` → `LayerSubscription` → `PlotOrchestrator`)

Per the assessment in #1042, ~500+ lines (subscription registry, `LayerSubscription`, two-phase plot creation, previous/current JobSet choreography, restore machinery) exist only to transport one fact — "the `job_number` for workflow W changed" — because an ephemeral uuid sits inside the dashboard's subscription identity.

[...]

Issue #1042, opened 2026-07-02
Assessment: Job identity and the JobOrchestrator/PlotOrchestrator complexity


SimonHeybrock:
# Assessment: Job identity and the JobOrchestrator/PlotOrchestrator complexity

Status: working assessment, 2026-07. Not an ADR — this document analyses the
current implementation and evaluates directions; a decision would be recorded
as an ADR once a direction is chosen.

## The question

Feeding (changing) streams into plot (re)creation feels far more complicated
than the high-level description suggests ("obtain current stream identity,
route it through a plotter"). How much of that complexity is home-made — in
particular the random `job_number` uuid minted on every commit — and what
would a switch to a one-in-one-out job model (job identity per
`(workflow, source)`, generation inferred from `start_time`, i.e. what NICOS
sees after ADR 0006) actually buy?

## 1. What `job_number` actually does today

The uuid is not one mechanism; it is load-bearing in six distinct roles.
Any change has to account for each one separately.

1. **Wire routing key.** `ResultKey = WorkflowId + JobId(source_name,
   job_number) + output_name`, serialized as the Kafka stream name
   (`JobResult.stream_name`, `Orchestrator.forward`). The dashboard's
   `DataService` is keyed by the full `ResultKey`.
2. **Generation token.** A fresh uuid per commit means a fresh set of keys:
   fresh `DataService` buffers (no cross-generation mixing in
   `WindowAggregatingExtractor` history), and unambiguous discrimination of
   late results from a stopped/replaced job (`ActiveJobRegistry.is_active`
   filter in `Orchestrator.forward`).
3. **Rebinding trigger.** Because the subscription key contains an ephemeral
   uuid, every consumer must *learn* the new keys on every commit. The entire
   notification chain — `JobOrchestrator._subscriptions` /
   `_workflow_subscriptions` / `WorkflowCallbacks`, `LayerSubscription`, and
   `PlotOrchestrator._on_all_jobs_ready` — exists to transport one fact:
   "the current job_number for workflow W changed".
4. **Command addressing.** Stop/reset are sent per `JobId`
   (`JobOrchestrator._send_job_commands`), although the backend already
   supports whole-workflow selectors (`JobCommand.workflow_id`, used by NICOS
   per ADR 0006).
5. **Retention of stopped-job data.** Data outlives the job it is keyed by,
   so retention must be choreographed: `WorkflowState.previous` vs
   `current`, the deferred `cleanup()` vs immediate `deactivate()` split in
[...]

Issue #1039, opened 2026-07-02
Bugs in dashboard for triage


SimonHeybrock:
Results from audit that need triage below.

---
I'll audit the dashboard code for subtle correctness bugs. Let me start by mapping the dashboard structure.

Now I'll launch five parallel audit agents, each covering a subsystem cluster, focused strictly on correctness bugs.

All five auditors are running. Meanwhile I'll review the remaining uncovered files myself (config adapters, registries, misc utilities).

Those two look fine. Checking the other uncovered files.

The utility files I reviewed myself (configuration_adapter, data_roles, plotter_registry, derived_devices, format_utils, save_filename, range_hook) look correct — no findings there. The five audit agents are still working through the core pipeline, buffers, orchestration, plotting, and widgets/session code; I'll verify and consolidate their findings as they report back.

The orchestration auditor reported back with several significant findings. Let me verify the key claims against the code before accepting them.

Confirmed — extractors are only ever added, never removed. Two of five auditors are in and verified; the orchestration findings (unguarded cross-thread mutation, mid-replay persistence truncation, extractor accumulation) and the buffer findings (datetime64 trim crash, live ring-buffer views) all check out against the code. Waiting on the core-pipeline, plotting, and widgets auditors before writing up the consolidated report.

The plotting auditor found a high-severity ROI-wipe bug. Verifying it and the slicer tuple/list comparison directly:

Confirmed — `_initial_rois` is parsed once from config and never updated, so every new browser session/reload republishes the config-time ROI set and wipes user-drawn ROIs on the backend. Four of five audits are now in and the major findings are verified. Waiting on the final widgets/session auditor before writing the consolidated report.

All five auditors are done. Let me spot-verify the widgets auditor's two high-severity claims before consolidating.

Both widget claims are confirmed by direct reading. Two last spot-checks on the remaining medium claims:

All findings are verified. The eviction claim and the DataService race are confirmed by direct reading (`__setitem__` on an existing key doesn't refresh insertion order; `register_subscriber`/`_get_extractors` touch shared state locklessly across threads). Here is the consolidated audit.

---

# Dashboard correctness audit

I audited all 81 files under `src/ess/livedata/dashboard/` (five parallel deep-read passes over the data pipeline, buffers, orchestration, plotting, and widgets/session code, plus my own review of the utility modules). Every finding below was re-verified against the source; the two buffer bugs were additionally **reproduced numerically** against scipp 26.3.1. Line numbers refer to the current `main` checkout.

[...]

Issue #955, opened 2026-06-01
Dashboard: stale-session cleanup mutates Bokeh document from a background thread (unsafe)


SimonHeybrock:
## Problem

Stale-session cleanup mutates Bokeh `Document` state from a background thread, which is not thread-safe.

The background orchestrator update thread (`DashboardServices._update_loop`, daemon thread) calls `session_registry.cleanup_stale_sessions()` on every iteration:

```python
# dashboard_services.py (_update_loop, background thread)
self.orchestrator.update()
self.session_registry.cleanup_stale_sessions()
```

For each stale session this runs `SessionUpdater.cleanup()` → `PeriodicCallback.stop()`:

```python
# session_updater.py
def cleanup(self) -> None:
    if self._periodic_callback is not None:
        self._periodic_callback.stop()
```

`panel.io.callbacks.PeriodicCallback.stop()` mutates the session's Bokeh **Document** callback collections:

```python
# panel/io/callbacks.py (stop)
if self._doc and self._cb ...:
    if self._doc._session_context:
        self._doc.callbacks.remove_session_callback(self._cb)
```

Bokeh document mutation must happen on the document's IOLoop thread, under the document lock. Performing it from the background orchestrator thread is a thread-safety violation: it races with the IOLoop and can (a) fail to actually cancel the IOLoop-bound 1 Hz `periodic_update` callback, leaving it firing after the session is gone, and (b) corrupt the document model graph during concurrent update/removal.

This path is the defense-in-depth cleanup for **unclean** disconnects, where Bokeh's `on_session_destroyed` never fires (browser crash, network drop, laptop sleep). Clean closes go through `on_session_destroyed` on the IOLoop and are fine.

## Symptoms consistent with this

- Production warnings: `Dropping a patch because it contains a previously known reference (id=...) ... updating a model on one side of a communications channel while it was being removed on the other end`.
- Leaked per-session `periodic_update` callbacks continuing to recompute plots after a session is gone (CPU spent with zero active sessions).
- On Bifrost PROD, a hard single-core CPU spin (no logs, flat memory) began immediately after the last session's teardown following an abnormal websocket close (`code=None`). Root cause of that specific freeze is not yet confirmed via a live profile, but this unsafe cross-thread teardown is a prime suspect.

## Fix direction

[...]

PR #1048, opened 2026-07-05
docs: draft ADR 0007 — single-writer state with versioned pull as the dashboard concurrency model

SimonHeybrock:
Draft ADR consolidating the root cause behind the audit bug clusters, per the analysis in #1046.

The 2026-07 audits (#1039, #1044, plus #1040, #955, and the assessment in #1042) located essentially every confirmed cross-thread race, session leak, and category of compute waste in one kind of code: synchronous push machinery (callbacks fanning out across threads and sessions carrying data or lifecycle events), while the code built on the version-polling pattern from `.claude/rules/dashboard-widgets.md` was audited clean. This ADR proposes making single-writer state + versioned pull the explicit, default concurrency and delivery model for the dashboard, so the three in-flight proposals (#1045 pull-on-frame, #1042 Option B, #1040's polling direction) are recognized as one architectural move and future features inherit the safe pattern by construction.

Status is `proposed` — this PR is meant as the concrete text to discuss and amend; the sequencing/execution program lives in #1046.

Refs #1046. Related: #1039, #1040, #1042, #1044, #955, #1043, #1045, #1009.

https://claude.ai/code/session_01PXpjcjaNvus3Qb1WEJefvQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01PXpjcjaNvus3Qb1WEJefvQ)_

### DIFF
--- a/docs/developer/adr/0007-single-writer-versioned-pull.md
+++ b/docs/developer/adr/0007-single-writer-versioned-pull.md
@@ -71,7 +71,14 @@ Concretely:
   consumers poll the version on their own clock (session poll, frame flush)
   and pull an immutable snapshot when it advanced. Callbacks, where they
   survive, carry no data — at most "something changed" wake-ups that
-  degrade to the next poll tick if lost.
+  degrade to the next poll tick if lost. The sanctioned instance is the
+  ingestion thread scheduling a data-free session tick via
+  `doc.add_next_tick_callback` once `flush_frames` has committed a grid: only
+  the *scheduling* crosses threads, the woken tick body runs in the owning
+  session's document context, and a lost or duplicated wake is harmless
+  because the tick is idempotent and generation-gated. ADR 0005's rejection
+  of `add_next_tick_callback` scopes to data-carrying push and to
+  `pn.state.execute`, not to wake-ups — see its 2026-07-06 amendment.
 - **Data flows by pull at display cadence.** Ingestion records what changed
   (dirty keys); extraction, copy, and compute happen when a consumer with a
   viewer pulls — not eagerly per delta on the ingestion thread. "No viewers,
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
Suggestion from the wake-up analysis ([details in #1046](https://github.com/scipp/esslivedata/issues/1046#issuecomment-4897619291)): the decision bullet *"Callbacks, where they survive, carry no data — at most 'something changed' wake-ups that degrade to the next poll tick if lost"* is exactly the right clause — consider naming its intended concrete instance explicitly: the ingestion thread scheduling a data-free session tick via `doc.add_next_tick_callback` after a `flush_frames` commit.

Two reasons:

1. **Apparent ADR conflict.** ADR 0005's "Alternatives considered" reads as a blanket rejection of `add_next_tick_callback` from the ingestion thread. Without a pointer, 0005 and 0007 appear to contradict each other on whether wake-ups are allowed. A scoped amendment to 0005 would resolve it: the rejection applies to *data-carrying* push and to `pn.state.execute` (which cannot resolve session context off the IOLoop), not to wake-ups that leave all session-bound mutation on the owning session's IOLoop — the trigger moves from clock to event, the flush itself stays where ADR 0005 requires it.
2. **The mechanism is verified**, not hypothetical: on panel 1.9.3 / bokeh 3.9.1, a background-thread `doc.add_next_tick_callback(tick)` runs the tick in the correct session context (`pn.state.curdoc` resolves, `pn.io.hold()` + `doc.models.freeze()` behave), across concurrent sessions, and scheduling into a destroyed document raises catchably (clean lazy unregistration). Experiment description in #1046.

Consequences section could then also state the payoff: with the wake-up as the flush trigger, worst-case freshness is bounded by *data cadence plus milliseconds* (fallback poll only as safety net), so the 100 ms fast poll — and its per-tick freeze recompute and layer scan (#1055 Target 3, #1044 scalability note) — can be retired rather than optimized.

---

docs/developer/adr/0007-single-writer-versioned-pull.md
```diff
@@ -0,0 +1,139 @@
+# ADR 0007: Single-writer state with versioned pull as the dashboard concurrency model
+
+- Status: accepted
+- Deciders: Simon
+- Date: 2026-07-05
```
SimonHeybrock:
```suggestion
- Date: 2026-07-20
```

SimonHeybrock (commented):
LGTM

Issue #1047, opened 2026-07-05
Backend: validate data-derived values at the adapter boundary (timestamps as the service clock)


SimonHeybrock:
## Summary

A second, smaller cluster from the audits — distinct from the dashboard push/pull program in #1046, but with the same shape of fix: several backend findings share the root cause that **adapters pass data-derived values across the trust boundary unvalidated, and each consumer guards (or doesn't) ad hoc**. The most dangerous instance is timestamps: the batcher uses data-derived timestamps as its *only* clock, with no sanity bound.

## Findings that trace to this root cause

- **#1038 finding 1 (High, reproduced):** a single far-future timestamp in the first consumed batch permanently wedges `SimpleMessageBatcher` (and the production `AdaptiveMessageBatcher`) — no batch is ever emitted again, messages pile up unbounded, and restarting re-consumes the same offset and wedges again. All four wire schemas (ev44, f144, da00, ad00) feed unvalidated data-derived timestamps into the batcher. The lag machinery *detects* future payloads and logs at ERROR, but nothing drops or clamps them.
- **#1038 finding 2 (Medium):** ev44 flatbuffers with absent event vectors return scalar `0` from the accessors, causing an `AttributeError` deep in the adapter — the message is silently dropped instead of hitting the intended timestamp fallback. An explicit empty-vector check at the boundary would recover these.
- **#1037 finding 4 (Medium):** multi-pulse ev44 on the monitor path is silently mis-stamped (all events get `reference_time[-1]`), while the detector path raises `NotImplementedError` for the same input. The two paths should agree, and the boundary is the place to enforce the invariant once.

## Proposed fix

One validation/normalization layer at the adapter boundary (where raw flatbuffers become domain messages), instead of per-consumer guards:

- **Bound data-derived timestamps against wall clock**: reject (or clamp, with a counter/log) payload timestamps more than N seconds in the future; optionally fall back to the Kafka broker `CreateTime` the adapters already read for lag reporting. This protects the batcher, the job scheduler, and run-reset logic at once.
- **Structural checks per schema**: empty/absent event vectors (ev44), `reference_time_index` consistency (multi-pulse), so malformed-but-recoverable messages are handled deliberately rather than via `except Exception` drops.
- Surface rejects/clamps in the existing stream stats so persistent upstream corruption is visible rather than silent.

## Why this is high value beyond the current findings

[...]

Issue #1038, opened 2026-07-02
Message-related bugs in backend for triage


SimonHeybrock:
Results from audit that need triage below.

---
I'll audit the backend message path (Kafka consumption → adapters → processor → job manager → workflows → sink) for messages that can crash services, deadlock them, or poison individual jobs. Let me start by mapping the code.

The backend here is `core/`, `kafka/`, `handlers/`, and `services/` (the dashboard is the frontend). I'll read the critical message path myself, starting with the core loop.

Core loop mapped. Now the batching, job management, and timestamp layers.

Command processing is well-guarded. Now the Kafka source/consumer and sink — exceptions there escape the service loop and are fatal.

Let me check the timeseries and detector handlers (per-job data paths) and the da00/f144 compat edge cases.

Let me verify whether the future-timestamp batcher hazard is guarded or tested anywhere, and confirm the ev44 flatbuffer edge case.

Confirmed the primary finding with a direct reproduction. Let me do a couple of final checks on the other data-timestamp entry points and the run-control path to characterize the blast radius.

I have enough to report. Here's the audit.

## Backend message-corruption audit

I traced the full backend ingest path — Kafka consumer → `AdaptingMessageSource`/adapters → `OrchestratingProcessor.process` → `MessageBatcher` → `MessagePreprocessor` (accumulators) → `JobManager`/`Job` → workflows → `KafkaSink`. The good news first: nearly every stage already isolates bad data. Adapter failures are caught per-message in `AdaptingMessageSource.get_messages` (kafka/message_adapter.py:592), preprocessing failures are caught per-stream in `MessagePreprocessor.preprocess_messages` (orchestrating_processor.py:142), per-job accumulate/finalize are wrapped in `Job.add`/`Job.get` (core/job.py:374, 424), command handling is wrapped in `ConfigProcessor` (config_handler.py:44), and the sink swallows non-fatal serialize/produce errors. So most corruption is contained to a single dropped message, stream, or job.

There is one important exception that sits on the shared path and is *not* guarded.

### Finding 1 (High): a single far-future timestamp permanently wedges the batcher — service-wide

[...]

Issue #789, opened 2026-03-12
Plots sometimes do not show up in some sessions


SimonHeybrock:
Observed with two active sessions: Session A sees the plot (recently started workflow), session B still showing "waiting for data". Refreshing the tab (which essentially makes a new session C) fixes the issue. It seems like we have a small hole or race in the plot creation logic, not triggering reliably.

I should be pointed out that session B showed *some* plots (in this case in other tabs), so it was not the entire session that was broken.

Edit: There is a chance that this is the same bug fixed in #810, but unfortunately I cannot remember if what I saw here was using `hv.Layout`, so it might not be.

SimonHeybrock:
Plausible root cause identified via the 2026-07 audits: the two-phase plot creation / job-rebinding machinery has confirmed cross-thread races (#1039 finding 3 — e.g. a stop notification landing after `_notify_workflow_available` leaves layers stuck in a state that only `data_arrived` can't exit, so plots stay blank in some sessions until recreated). That machinery is slated for deletion under #1042 Option B, tracked in the consolidation program #1046 — this issue should be re-checked (and hopefully closed) once those slices land.

---
_Generated by [Claude Code](https://claude.ai/code)_

SimonHeybrock:
Wave 3 (#1064, #1065, merged) deleted the machinery implicated here: `LayerSubscription`, the `WAITING_FOR_JOB` phase, and the `JobOrchestrator` subscription registry with both notify loops (#1039 finding 3). Layer run-state is now derived by single-threaded polling (`PlotOrchestrator.sync_job_states`) with version-guarded rebuilds, so the "stuck waiting-for-data until `data_arrived`" path no longer exists.

The symptom was never reproducible locally, so this is a close by code-reasoning, not by empirical verification — please reopen if it recurs in the field. One caveat: a known, accepted micro-race remains (commit+stop within one poll interval retains the pre-commit frame as stopped). If that ever surfaces it will look like a stale frame, not a missing plot — a differently-shaped bug, worth a fresh issue.

PR #1053, opened 2026-07-06
tests: hostile-wire liveness harness and multi-session browser smoke test

SimonHeybrock:
## Motivation

Two validation gaps let whole classes of errors through untested, both aimed at boundaries the ongoing dashboard restructuring (#1046) will not move:

1. **The backend's worst known error class is invisible to the test suite.** Service-level tests run with `NaiveMessageBatcher`, so the production `AdaptiveMessageBatcher` — which the audit (#1038) showed is permanently wedged by a single far-future data timestamp, silencing the entire service — is never exercised. Nothing feeds malformed or insane payloads through the real adapter wiring.
2. **Multi-session dashboard behavior is manual-only.** Nearly every concurrency PR carries the same unchecked item ("manual: check with 2+ browser sessions", e.g. #1009, #1043), and #789 is exactly that regression class in the field.

## What

**Hostile-wire corpus + liveness harness** (runs in CI, ~7 s total):

- `tests/helpers/hostile_wire.py`: reusable corpus of hostile payloads in two families — *malformed* (garbage, truncated flatbuffers, hand-built ev44 with absent event vectors, wrong schema, unmapped source) and *well-formed but insane* (far-future data timestamps for all four data schemas, pre-epoch timestamps, disagreeing `time_of_flight`/`pixel_id` lengths). New corruption modes get added here and both test layers pick them up.
- `tests/kafka/adapter_robustness_test.py`: pins the existing per-message containment in `AdaptingMessageSource` (regression guards, pass today) and the accepted-by-design cases (mismatched vectors on the plain monitor path, da00 dtype fallback), plus strict `xfail`s asserting data-derived timestamps must not cross the adapter boundary unbounded.
- `tests/services/hostile_input_liveness_test.py`: runs the monitor service with the production-default batcher and asserts one invariant — after consuming any single hostile payload, the service still publishes results for subsequent well-formed data. The reproduced #1038-finding-1 wedge is a strict `xfail`; the surviving variants (mid-stream future timestamp, ancient/pre-epoch timestamps mid-stream and in the first batch, mismatched vectors, all malformed cases) pass and pin current containment behavior.

The strict `xfail`s double as **ready-made acceptance tests for #1047**: when the adapter-boundary validation lands they flip loudly to XPASS and pytest forces the markers' removal, so the known holes cannot silently rot. Any resolution shape (drop, clamp, broker-timestamp fallback) satisfies the assertions as written. There is deliberately no "too old" analogue of the far-future invariant — backlog consumption makes genuinely old timestamps legitimate, so pre-epoch values are exercised for liveness only.

Scope note: this PR contains only cases with an unambiguous contract and a verified reproduction. Extensions needing a contract decision or behavior characterization first (timeseries/f144 path, rate-aware batcher, multi-pulse ev44, NaN f144, detector service) are tracked in #1054 with the reasoning spelled out there.

**Multi-session browser smoke test** (local-only, `pytest -m browser`, ~60 s):

- `Dashboard.connect_many(n)` in `scripts/drive_dashboard.py` opens n isolated sessions (own context/websocket/Bokeh session) on one server, ordered oldest-first for late-joiner scenarios. The Chromium binary is overridable via `PLAYWRIGHT_CHROMIUM_EXECUTABLE` for containers whose playwright package doesn't match the provisioned browsers.
- `tests/dashboard/multisession_smoke_test.py` drives the Kafka-free fake backend with two sessions and walks the manual checklist once: the late joiner sees the same tabs and populated plots, both sessions keep receiving data updates (fingerprinting rendered `ColumnDataSource` contents), and tab switching in one session stalls neither. Deliberately behavioral — it observes rendered data, not internals — so it stays valid across the delivery-mechanism refactors (#1045, #1046) and can help verify them.
- New `browser` pytest marker, excluded from the default run and the CI tox envs. The marker definition and exclusions mirror the in-flight #1030 exactly, so the branches merge cleanly in either order.

## Test plan

- [x] `adapter_robustness_test.py` + `hostile_input_liveness_test.py`: 20 passed, 8 xfailed (~6 s); the wedge xfail was verified to reproduce end-to-end (output frozen forever vs. steady growth in an unpoisoned control).
- [x] `pytest -m browser tests/dashboard/multisession_smoke_test.py` passes against the fake backend; `drive_dashboard.py --launch --map` still works after the `connect` refactor.
- [x] Full fast suite green (3912 passed, 8 xfailed); the only failures are pre-existing environmental ones (blocked `pooch` downloads for dream/estia data in the sandbox).
- [ ] Optional reviewer check: run `pytest -m browser` on a dev machine with Playwright installed.

https://claude.ai/code/session_01WYME35Uh2EmdyEhmhjcXYM

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,6 +19,12 @@ python -m venv .venv && source .venv/bin/activate && pip install -e ".[test]"
 - Tests in `tests/` mirror `src/ess/livedata/` structure, files follow `*_test.py`
 - All unit tests run without Kafka -- use fakes from `fakes.py`
 - `pytest` uses `--import-mode=importlib`
+- Robustness harnesses: `tests/helpers/hostile_wire.py` is the corpus of
+  malformed/insane wire payloads consumed by the adapter- and service-level
+  robustness tests (add new corruption modes there). Known holes are strict
+  `xfail`s referencing their issue; fixing the issue flips them loudly.
+- `pytest -m browser` runs local Playwright UI tests (fake backend, e.g. the
+  multi-session smoke test); excluded from the default run and CI.
 
 ```sh
 python -m pytest                      # fast tests only (~25s)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ addopts = """
 --strict-markers
 --import-mode=importlib
 --benchmark-skip
--m "not integration and not slow"
+-m "not integration and not slow and not browser"
 -ra
 """
 testpaths = "tests"
@@ -112,6 +112,7 @@ markers = [
   "instrument: Specify instrument name for parameterized tests",
   "services: Specify which services to use in integration_env fixture (monitor, detector, or reduction)",
   "slow: Slow tests (>2s) excluded by default, run in CI via tox",
+  "browser: Browser-driven (Playwright) UI tests; run locally, excluded from CI",
 ]
 filterwarnings = [
   "error",
--- a/scripts/drive_dashboard.py
+++ b/scripts/drive_dashboard.py
@@ -15,6 +15,9 @@ Two ways to use it:
           dash.goto_tab("Detectors")
           dash.screenshot("plots.png")
 
+  For multi-session scenarios (state is process-global, widgets per-session),
+  :meth:`Dashboard.connect_many` opens n isolated sessions on one server.
+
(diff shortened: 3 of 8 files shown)


SimonHeybrock (commented):
LGTM

PR #1056, opened 2026-07-07
Default to sciline NaiveScheduler in backend services

SimonHeybrock:
## Motivation

Sciline silently selects dask's threaded scheduler for every `Pipeline.compute()` whenever dask is importable — and it is always importable, since both esslivedata and essreduce depend on it. `StreamProcessor` computes on internal `Pipeline` copies and offers no `scheduler` argument, so backend jobs have been paying the dask graph-conversion and dispatch machinery on every `accumulate()` and `finalize()` (twice per job per ~1 s batch) without any way to opt out — and without any benefit, since production ran dask on a synchronous executor anyway (parallelism happens at the job level via `--job-threads`).

Benchmarked per-cycle cost (accumulate + finalize, median, warm):

| scheduler | monitor wf (14-node graph, 10k events) | detector view 128×128 (29-node graph, ~100k events) |
|---|---|---|
| dask threaded + sync pool (previous default) | 3.6 ms | 14.8 ms |
| dask threaded + real thread pool | 5.2 ms (p90 spikes ≫) | 16.8 ms |
| sciline `NaiveScheduler` | **2.6 ms** | **12.6 ms** |

The dask overhead is ~40 µs of pure bookkeeping per graph node per compute call. With `NaiveScheduler`, every job's per-batch latency drops by ~1–3 ms; outputs are byte-identical (verified with `sc.identical` across all monitor and detector-view targets).

## Approach

Per-instance injection is impossible without changing essreduce (`StreamProcessor.__init__` builds fresh plain `sciline.Pipeline()` objects internally, so a `Pipeline` subclass does not survive), and removing the dask dependency wouldn't help either (essreduce's metadata pulls it back in, and sciline auto-picks it whenever importable). The remaining seam is sciline's default-scheduler selection site.

New module `ess.livedata.core.sciline_scheduler`:
- `configure_sciline_scheduler(mode)` overrides `sciline.task_graph.DaskScheduler`, the factory `TaskGraph` instantiates when no scheduler is passed — this is sciline's documented default-selection behavior, redirected process-wide.
- Because this patches a sciline-internal name, a **startup probe** builds a trivial pipeline and verifies the expected scheduler was actually selected, so a future sciline refactor breaks service startup loudly instead of silently reverting to dask.

CLI (`DataServiceRunner`):
- New `--scheduler {naive,dask-sync,dask-threaded}`, default `naive`.
- `dask-sync` reproduces the previous default exactly (threaded scheduler + `SynchronousExecutor` pool), `dask-threaded` the previous `--no-sync-scheduler`, so production can A/B or roll back without a code change.
- The deprecated `--sync-scheduler`/`--no-sync-scheduler` flags still parse and map to the two dask modes (with a deprecation warning), so existing deployment configurations keep working; an explicit `--scheduler` wins.

## Notes

- If sciline later ships an improved scheduler upstream, it can be added as another mode here — in the current release (25.11.1) only `NaiveScheduler` and `DaskScheduler` exist.
- `NaiveScheduler` holds intermediate results until the computation finishes rather than freeing them eagerly; irrelevant at our graph sizes (14–29 nodes).
- Two `tests/services/data_reduction_test.py` DREAM failures and two `registered_workflow_factories_test.py` roundtrip failures in my environment are pre-existing (blocked external download of instrument data files; they fail identically on `main`).

https://claude.ai/code/session_01XuFKpNkobLPnbg62B4TNnc

---
_Generated by [Claude Code](https://claude.ai/code/session_01XuFKpNkobLPnbg62B4TNnc)_

### DIFF
--- a/docs/developer/design/backend-service-architecture.md
+++ b/docs/developer/design/backend-service-architecture.md
@@ -207,5 +207,5 @@ Services use `ExitStack` for automatic resource cleanup on service exit or error
 
 ## Building Services with DataServiceBuilder
 
-`DataServiceBuilder` constructs services consistently with `OrchestratingProcessor` by default. For command-line services, `DataServiceRunner` (see `service_factory.py`) wraps a builder and adds standard CLI arguments (`--instrument`, `--dev`, `--log-level`, `--sync-scheduler`, `--job-threads`). Services can publish initialization messages on startup for workflow specifications or configuration values.
+`DataServiceBuilder` constructs services consistently with `OrchestratingProcessor` by default. For command-line services, `DataServiceRunner` (see `service_factory.py`) wraps a builder and adds standard CLI arguments (`--instrument`, `--dev`, `--log-level`, `--scheduler`, `--job-threads`). The `--scheduler` option selects the scheduler sciline uses for workflow execution (default `naive`, which avoids per-cycle dask overhead; see `ess.livedata.core.sciline_scheduler`). Services can publish initialization messages on startup for workflow specifications or configuration values.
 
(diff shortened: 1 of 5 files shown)


SimonHeybrock:
LGTM

PR #1059, opened 2026-07-20
Validate data-derived timestamps at the adapter boundary

### LINKED ISSUE #1047: Backend: validate data-derived values at the adapter boundary (timestamps as the service clock)

SimonHeybrock:
## Why

The batcher hardening keeps services alive against pathological timestamps, but it cannot resolve every case: a lone far-future timestamp in the startup backlog is indistinguishable from real traffic without arrival-time evidence, and the simple batcher wedges on it (the strict xfail from #1122). The adapter boundary has that evidence — it sees the broker `CreateTime` and the consuming host's wall clock — so validation belongs there.

## What changed

Data-derived envelope timestamps implausibly far ahead of the wall clock (bound: 30 s; NTP skew is milliseconds, a broken clock is hours out, and three days of production logs contain no future-dated payload at all) are clamped at the adapter boundary. The clamp target is the broker `CreateTime` when itself plausible, the consumer's wall clock as a last resort — the order matters when consuming a backlog, where a wall-clock substitute would be a future outlier relative to the surrounding traffic and re-poison the batchers from within the bound.

The change establishes an explicit contract: **the envelope timestamp is livedata's transport clock** (batching, scheduling, run resets) and may be clamped freely, while **payload times are the device's claim** and reach science consumers unmodified. A wrong device clock stays visible and attributable in the data instead of being silently replaced by an estimate; consumers must tolerate wrong device times regardless, since a clock running behind passes any future bound. Concretely:

- No payload is rewritten on clamp, including f144.
- `event_time_zero` is now built from the payload's `reference_time` (threaded through the `Events` dataclasses) instead of inheriting the envelope, closing the path by which the clamp used to rewrite pulse times covertly. The envelope stands in only for messages carrying no pulse time.
- pl72/6s4t run start/stop times are the exception that proves the rule: they are control flow (they schedule job resets that fire when data time reaches them, so a far-future value would park a reset that never fires and lose the run transition) and are bounded like the envelope.

Clamping rather than dropping keeps data flowing during commissioning, when devices with broken clocks are expected. For a clamped ev44 stream this is a degraded mode with a known taxonomy: physics based on `event_time_offset` (most live reduction) is unaffected; correlation with slow f144 metadata holds at arrival-time accuracy while consumption is live; pulse-precision correlation is unrecoverable — no timestamp policy can restore it once the producer clock is broken.

Clamps are counted per stream and surfaced in the backend stream stats at warning severity, with a clamped-message column in the backend status widget. Adapters without a `StreamCounter` (dashboard-side routes) log once per stream instead of once per message.

While in the adapter, ev44 payloads whose event vectors are absent rather than empty now reach the broker-timestamp fallback intended for them, instead of dying on an `AttributeError` and being dropped (#1038 finding 2).

## Notes for review

An unmapped stream sending far-future timestamps is reported as unmapped rather than as a clamp: Kafka subscription is per-topic, and a foreign producer's broken clock should not raise warnings in a service that ignores its data.

A known residual hole: a message whose payload timestamp and `CreateTime` are both insane, consumed mid-backlog, degrades to the wall-clock fallback and is then a future outlier relative to the backlog's epoch; the rate-aware batcher recovers via re-anchoring, the simple batcher stays silent until catch-up completes.

A possible follow-up, deliberately not in scope: estimating a per-stream clock offset and clamping to `timestamp - offset`, which would preserve the pulse grid and upgrade coarse correlation to near-exact for constant-offset clocks.

Closes #1047. Also resolves findings 1 and 2 of #1038.

## Test plan

The strict xfail acceptance tests from #1053 and from #1122 now pass and their markers are removed. The far-future liveness tests exercise both clamp targets end-to-end. Adapter-level tests assert delivery with the clamped envelope, untouched payloads, the broker-time-vs-wall-clock cascade, run-time bounding, and once-per-stream warning rate-limiting without a counter.

- [ ] Backend status widget: confirm the clamped-message column and its summary badge render sensibly for a stream with clamps.

### DIFF
--- a/src/ess/livedata/core/job.py
+++ b/src/ess/livedata/core/job.py
@@ -119,6 +119,10 @@ class StreamStat:
     source_name: str
     stream: str | None  # Resolved stream name, or None if unmapped
     count: int
+    # Messages whose implausibly-far-future timestamp was clamped to the wall
+    # clock (see KafkaAdapter._clamp_future). A subset of ``count``, so a
+    # stream whose every timestamp was clamped reports count == clamped.
+    clamped: int = 0
 
 
 @dataclass(frozen=True, slots=True)
--- a/src/ess/livedata/core/message_batcher.py
+++ b/src/ess/livedata/core/message_batcher.py
@@ -27,9 +27,10 @@ class MessageBatch:
 # batchers' only clock, so window placement must not follow one outlier
 # arbitrarily far into the future: an over-anchored window stalls delivery
 # until wall time catches up, while an under-anchored one self-heals forward
-# as newer traffic arrives.  The cap holds regardless of where a message
-# came from: sources are not trusted to deliver sane timestamps, and the
-# batcher is the one place every message passes through.
+# as newer traffic arrives.  The adapter boundary clamps far-future
+# timestamps to an arrival estimate; this cap is defense in depth for values
+# inside the adapter's bound -- including wall-clock clamps, which land ahead
+# of the traffic when consuming a backlog -- and for non-Kafka sources.
 MAX_TIMESTAMP_AHEAD_BATCHES = 3
 
 
@@ -55,7 +56,9 @@ def plausible_anchor(timestamps: list[Timestamp], batch_length: Duration) -> Tim
     traffic is not the safe default it appears to be: a window only walks
     forward one batch length per call, so an anchor stranded in a disjoint
     past epoch takes astronomically many calls to catch up, where one
-    stranded ahead merely waits out its own distance.
+    stranded ahead merely waits out its own distance.  Resolving these is
+    the adapter boundary's job, which clamps an implausible timestamp to an
+    arrival estimate rather than weighing it against its neighbours.
     """
     ordered = sorted(timestamps)
     horizon = MAX_TIMESTAMP_AHEAD_BATCHES * batch_length
@@ -242,8 +245,8 @@ class SimpleMessageBatcher(MessageBatcher):
         a message parked implausibly far ahead advances the window on every
         call -- once per poll rather than once per batch length. The window
         then overruns the live data and comes to rest at the outlier, and
-        nothing closes a batch until wall time catches up: the #1038 wedge.
-        Refusing to advance past
+        nothing closes a batch until wall time catches up: the #1038 wedge,
+        merely bounded by the adapter's future bound. Refusing to advance past
         the frontier of the data in hand keeps the window with the traffic,
         and the frontier moves forward as real messages arrive.
 
--- a/src/ess/livedata/core/orchestrating_processor.py
+++ b/src/ess/livedata/core/orchestrating_processor.py
@@ -379,6 +379,7 @@ class OrchestratingProcessor(Generic[Tin, Tout]):
                     window_seconds
                 )
                 self._log_stream_lag(self._stream_stats_provider.drain_lag())
+                self._log_clamped_streams(self._pending_stream_stats)
 
             logger.info(
                 'processor_metrics',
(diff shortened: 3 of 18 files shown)


SimonHeybrock:
Closing in favour of #1141, which fixes what is left of this after #1122 landed in its final shape.

Re-checking the motivation against `main`:

- **Liveness no longer needs it.** The wall-clock backstop bounds a poisoned window at roughly ten batch lengths and costs one stall per poison message, and the rate-aware batcher is now the default. The remaining hole is specific to `SimpleMessageBatcher`, which is opt-in and slated for removal.
- **The data-clock damage this was sized against is currently unreachable.** `_fire_pending_resets` only ever sees an empty list, because `with_run_control_route` is a no-op for #795, and no `JobSchedule` carrying an `end_time` is constructed anywhere in `src/`. Both arm again later, but neither is live today — which also means the pl72/6s4t bounding commit here guards a disabled path with a 30 s bound that would clamp a legitimately scheduled future stop time.
- **The dashboard-side clamp cannot do anything.** `MessagePump` forwards only `stream` and `value`; the envelope timestamp is discarded. That takes the counter-less rate-limiting commit with it, which existed only to quiet warnings from a clamp with no effect.
- **The clamp would blind the detector we already have.** `_clamp_future` runs before `_record_lag`, so when the clamp target is the broker timestamp the recorded lag becomes exactly 0 — a stream that today logs `stream_lag` at ERROR drops to INFO. That is the measurement #1133 wants to build creep detection on.

What did survive is a permanent, in-band defect: a batch built from a lone far-future message anchors on it, and `Job._start_time` then latches a start time outside the stream's timeline for the job's lifetime, leaving every later result with `start_time > end_time`. #1141 corrects that where the evidence is (a job cannot begin after its accumulation ends) without a wall clock or a new layer.

The parts of this PR that stand on their own are separable and unaffected: per-source clock-offset telemetry belongs with #1133 and should extend the existing lag reporting rather than add a parallel counter; the ev44 absent-event-vector handling is #1124, closed separately.

PR #1061, opened 2026-07-20
Remove departed subscribers' extractors from buffer retention

SimonHeybrock:
Part of #1044 (extractor-removal symmetry), sequenced as the prerequisite for #1042 Option B slice (i) per the corrected amendment A1 on that issue.

`DataService.unregister_subscriber` removed the subscriber from the notification list but left its extractors in `_BufferState.extractors` forever, so extractor lists grow and max retention ratchets up over repeated register/unregister cycles on a long-lived key. Today this is bounded because every commit mints fresh keys and old buffers die at cleanup; under the stable dashboard keys of #1042 slice (i) buffers become immortal and the growth unbounded over multi-day operation, so this must land first.

Naive symmetric removal (`set_extractors` with the survivors) would regress plot edits: an unregister/re-register cycle on the same key passes through a survivors-empty state whose Temporal→Single downgrade keeps only the last time slice, wiping timeseries history the backend never resends (delta emission). The unregister path therefore applies survivor extractors with the type downgrade disabled — buffer type is sticky-upward for the buffer's lifetime (resets at buffer deletion) — and empty-survivor states retain previous retention requirements, since shrinking to zero would let an append trim history inside the transient cycle.

Known residual window (documented in the docstring, accepted for now): the unregister/re-register cycle is not atomic, so if a surviving subscriber needs a shorter timespan than the re-registering one, an append landing in the gap may trim history to the shorter window. Making the cycle atomic is plot-orchestrator API surgery that #1042 slices (ii)+ reshape anyway.

## Test plan

- New regression test: unregister/re-register cycle on a populated timeseries key preserves full history (fails on the naive fix).
- Retention shrinks to survivors and extractor count stays constant over repeated cycles; empty-survivor state keeps buffer type and requirements.
- Existing concurrent ingestion/subscriber-churn stress test now exercises the new path.

### DIFF
--- a/src/ess/livedata/dashboard/data_service.py
+++ b/src/ess/livedata/dashboard/data_service.py
@@ -263,6 +263,16 @@ class DataService(MutableMapping[K, V]):
         """
         Unregister a subscriber, stopping it from receiving updates.
 
+        Drops the subscriber's extractors from buffer retention: each affected
+        buffer is reconfigured to the surviving subscribers' requirements, so
+        retention cannot ratchet up over repeated register/unregister cycles.
+        Buffer type never downgrades here (sticky-upward, see
+        :py:meth:`TemporalBufferManager.set_extractors`), so buffered history
+        survives an unregister/re-register cycle on the same key. The cycle is
+        not atomic, though: if a surviving subscriber needs a shorter timespan,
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

Issue #1037, opened 2026-07-02
Bugs in backend for triage


SimonHeybrock:
Results from audit that need triage below.

---

# Backend correctness audit — esslivedata

**Scope:** `core/`, `kafka/`, `handlers/`, `services/`, and backend-relevant `config/` (~13.5k lines). Method: I reviewed the core message-flow layer (orchestrator, job manager, batchers, job lifecycle) line by line myself, fanned out three parallel audit agents over `kafka/`, `handlers/`, and `services/`+`config/`, and then re-verified every candidate finding directly in the source (several were also verified empirically against the installed `ess.reduce`). Baseline: fast test suite passes (3896 passed; 4 failures are sandbox-network-only, from tests downloading DREAM files off `public.esss.dk`).

### 1. Downstream workflows mutate shared zero-copy preprocessor buffers in place — a data race under the default `job_threads=5`

`ToNXevent_data.get()` returns a DataArray that aliases the accumulator's reusable buffers (`handlers/to_nxevent_data.py:157-202`), including the `_WeightsBuffer` ones-array, with a documented "callers must not modify" contract. But `ess.reduce`'s `assemble_monitor_data`/`assemble_detector_data` call `_add_variances`, which does `da.copy(deep=False)` and then `var.variances = var.values` **in place** on the underlying Variable — i.e. directly on the shared buffer. Two aggravating factors:

- The preprocessor keeps **one** accumulator per stream, and `_filter_data_for_job` (`core/job_manager.py:727-740`) hands the **same object by reference** to every job consuming that stream. The monkey-patched `group_event_data` even returns the input object unchanged (`handlers/_patch_group_event_data.py:32`), so all detector jobs on one source share one DataArray.
- Jobs run concurrently in a `ThreadPoolExecutor` (`core/job_manager.py:582,691`), and `service_factory.py:64` defaults to `job_threads=5`.

So with ≥2 jobs on the same monitor/detector source, multiple threads concurrently attach variances to a Variable other threads are reading — every cycle on the detector path (fresh grouped object per batch). Today this is silent because the written variances (1.0) equal what each job would have computed, but it is undefined-behavior territory in scipp (partially initialized variance arrays visible to readers, or a crash), and any future in-place operation in a workflow would corrupt sibling jobs' data outright. **Confidence: high** that the contract-violating shared mutation exists (empirically verified); observable corruption needs an unlucky interleave.

[...]

PR #1062, opened 2026-07-20
Key the dashboard data plane by stable DataKey

### LINKED ISSUE #927: Bounded memory increase from retained state.previous buffers
Follow-up to #926.

After #926, stopping a workflow retains the previous job's buffered data so that plots created or reconfigured while the workflow is stopped can render the last known state. Data is dropped only when the job is pushed out of ``state.previous`` by a later stop/commit.

This means each workflow can hold up to 2 generations of buffered data (``state.current`` + ``state.previous``) instead of 1. The cost is bounded — linear in workflow count, capped at one extra generation per workflow, capped per buffer by the configured retention window or geometry — but real:

- SingleValueBuffer plots: negligible (kilobytes).
- TemporalBuffer plots: bounded by ``required_timespan × data rate``.
- Per-pixel detector outputs: bounded by detector geometry. DREAM-scale outputs (~120 banks × ~10k bins) can sit around ~100–200 MB resident per stopped workflow.

The retained data lingers until the user commits a new job, stops again, or removes the plot/cell — there is currently no explicit way to release it.

## What may be worth adding

A UI affordance to evict ``state.previous`` explicitly (e.g. a "Clear stopped data" action on the workflow widget). The plumbing already exists:

- ``ActiveJobRegistry.cleanup(state.previous.job_number)`` to drop the buffers.
- Null out ``state.previous`` in ``JobOrchestrator``.

``ActiveJobRegistry.cleanup`` already batches its deletes inside a ``DataService`` transaction, so it is safe to call with plots still subscribed to those keys — they will see a single empty notification and keep their last cached plot state. That defensive transaction wrap (also added in #926) was put in place anticipating exactly this caller.

## Decision

I think it is fine to ship #926 without this — the extra generation is bounded and most setups won't notice. Re-open if memory pressure shows up in practice on production dashboards.

SimonHeybrock:
Implements #1042 Option B slice (i) (Wave 2 of #1046), following the slice (i) design and amendments recorded on #1042. Stacked on #1061 (extractor-removal symmetry), which it relies on. **Nothing changes on the wire, in the backend, or for NICOS.**

The dashboard's data plane is now keyed by the stable `DataKey(workflow_id, source_name, output_name)` — the triple ADR 0006 already uses for stable extraction — instead of the wire `ResultKey`, which embeds the per-commit `job_number` uuid. `Orchestrator.forward` maps wire→stable at ingest. The uuid is demoted to two roles: a generation filter (only the workflow's current `job_number` is admitted; stale messages are counted and logged once per uuid) and a per-key provenance stamp, resolvable to the committed config via `ActiveJobRegistry.resolve_config` and readable atomically with data via `DataService.snapshot_with_stamps`.

`ActiveJobRegistry` becomes a per-workflow generation registry: `(current, last)` generations with their configs. `begin_generation` is the commit flip — rotate, clear the workflow's buffers (the only eviction point), prune job statuses of the generation leaving the window — under the ingestion guard, with the clear atomic to pulls (single DataService lock hold). The flip moves **before** `send_batch`, closing the window where the first new-generation result raced activation. Stop retains data under the stable keys; `JobStatus` filtering accepts `(current, last)` so a just-replaced job can still report final states. Buffer clearing preserves extractor registrations, buffer type, and retention (the #1061 symmetry with downgrade disabled), so windowed plots survive a commit but can never aggregate across a parameter change.

Decision D7 is implemented as agreed: a byte-identical recommit clears too (uniform semantics, matching the fresh buffers per-commit keys used to provide).

ROI publishing was the one consumer needing the wire identity: `ROIPublisher` now takes `(workflow_id, source_name)` and resolves the current `JobId` at publish time via a resolver injected once the `JobOrchestrator` exists; no active job → skip with a warning; unwired resolver → fail-fast `RuntimeError`.

Deliberately unchanged (slices (ii)–(v) per #1046): `LayerSubscription`, two-phase plot creation, the `JobOrchestrator` subscription registry, current/previous `JobSet` persistence and restore. The per-commit pipeline rebuild now happens with identical keys; #1061 makes its unregister/register cycle net-zero on buffers.

Known behavior recorded per amendment A2: a send swallowed by the sink after the flip (#1037 finding 7) leaves cleared buffers with no job to fill them — a blank plot with no error, at parity with the previous deactivate-then-send ordering. Recovery signal is pending-command expiry (Wave 1.5 in #1046).

## Test plan

- Race merge gate: `TestConcurrentForwardAndCommit::test_commit_ingest_race_preserves_single_generation_invariant` — barrier-synchronized ingest/commit race asserting post-flip emptiness plus a free-running no-mixed-generations checker. Validated by mutation: removing both the registry-internal lock and the commit-path guard makes it fail; each layer alone suffices (defense in depth).
- Generation semantics: clear-on-recommit (including byte-identical, D7), retain-on-stop, late-old-data rejection, other-workflow isolation, status window semantics, stamp lifecycle, `clear_keys` atomicity.
- E2E against the fake backend via Playwright (`scripts/drive_dashboard.py` kit): plots fill through the stable-key plane; stop retains the displayed plot; restart clears (colorbar max drops an order of magnitude right after recommit) and the plot resumes accumulating.
- [ ] Manual session against a real backend: recommit a running workflow with changed parameters; draw an ROI after the recommit and confirm the backend applies it to the new job.

Closes #927: the memory doubling it tracked (current + previous generation resident per workflow after a recommit) is structurally gone — at most one generation is buffered per workflow at any time, and the commit-flip clear drops the backing arrays. Its secondary suggestion (an explicit "clear stopped data" affordance) is a future feature to be added on request; a stopped workflow's last generation stays displayed until that workflow's next commit, by design.


SimonHeybrock:
LGTM, preparation for full cleanup

PR #1064, opened 2026-07-20
Static plot creation + polled run-state; delete subscription registry

SimonHeybrock:
Part of #1042 (Option B, Wave 3 on #1046, slices ii and iii). With the data plane keyed by stable `DataKey`s (#1062), a layer's keys are computable from its `PlotConfig` alone — the entire callback apparatus that transported "the job_number for workflow W changed" to the plot layer becomes unnecessary and is deleted: `LayerSubscription`, the two-phase plot creation (`WAITING_FOR_JOB` → `on_ready` → create plotter → setup pipeline), and the `JobOrchestrator` subscription registry (`subscribe_to_workflow`, `WorkflowCallbacks`, both notify loops). This closes the #1039 finding 3 races by deleting the racy machinery rather than locking it.

**What replaces each deleted behavior** (H3 ledger):

- *Plot creation timing*: plotter + pipeline are set up unconditionally at `add_layer` time. The initial dirty mark pulls retained data, so a layer added while its workflow is stopped renders the last results and starts as STOPPED — this replaces `LayerSubscription`'s previous-job replay without needing `get_previous_job_number`.
- *Job start/stop notifications*: `PlotOrchestrator.sync_job_states()`, polled from the orchestrator update loop (before `flush_frames`, so a commit and its blank land in the same pass). It compares each layer's workflows' active job numbers against the last observed: changed and all running → presentation reset; changed and not all running → STOPPED (frame retained). Multi-workflow gating (correlation layers reset only when *all* workflows run, freeze when *any* stops) matches the old `LayerSubscription` semantics.
- *Per-commit pipeline rebuild* (incidentally blanked plots and reset autoscale): an explicit presentation reset per amendment A3 — on generation change the plotter is replaced, driving the layer to a blank `WAITING_FOR_DATA` instead of keeping a stale frame from the cleared buffers. The pipeline is untouched (keys are generation-independent), so extractor lifecycle is now owned solely by layer add/remove (#1061 symmetry).
- *UI states*: `WAITING_FOR_JOB` is gone; states are derived from run-state + data arrival. A never-run workflow's layer shows "Workflow not running" (formerly "Waiting for workflow").

Known micro-race, documented in `sync_job_states`: a commit immediately followed by a stop within one poll interval (~100 ms) is observed as a plain stop and retains the pre-commit frame. Accepted as unreachable in practice.

The merge-gate race test (`restart_cleanup_test.py`, run with `-m ""`) passes unchanged; the commit-path ingestion guard and registry lock are untouched.

## Test plan

- New pinning tests: `TestSyncJobStates` (recommit → blank fresh plotter; pipeline survives reset; stop freezes with retained frame; poll idempotence; restart-after-stop) and `TestSyncJobStatesMultiWorkflow` (all-running gating, single stop transition).
- `plot_orchestrator_test.py` now shares one `DataService` between pipelines and `ActiveJobRegistry` (as in production), so commit-clear is observable in tests.
- Full suite + browser smoke test (`pytest -m browser`) pass.
- Verified visually via `scripts/drive_dashboard.py`: stop retains the frame; restart resets accumulation (colorbar max drops vs. a never-stopped panel).


tests/dashboard/plot_orchestrator_test.py
```diff
@@ -2917,3 +2930,276 @@ def test_generation_is_scoped_per_grid(
...
         assert orchestrator.frame_generation(grid_a) == 1
         assert orchestrator.frame_generation(grid_b) == 0
+
+
+class TestSyncJobStates:
+    """sync_job_states drives layer lifecycle from polled workflow run-state.
+
+    Pins the replacements for the deleted subscription callbacks: a commit
+    observed by the poll resets the layer presentation (clear-at-commit,
+    #1042 A3), a stop freezes it with the last frame retained.
```
SimonHeybrock: This comment references history and does not seem useful.

SimonHeybrock (commented):
LGTM

PR #1065, opened 2026-07-20
Drop previous-JobSet tracking and persistence

SimonHeybrock:
Part of #1042 (Option B, Wave 3 on #1046, slice iv). Stacked on #1064.

The previous-JobSet machinery existed to let `LayerSubscription` replay a stopped workflow's job to late-created layers. With static plot creation (#1064) retention is free — a stopped workflow's data stays under its stable keys and any new layer binds to it directly — so `WorkflowState.previous`, `get_previous_job_number`, and the `previous_job` persistence key have no remaining consumers.

`ActiveJobRegistry.restore` now takes only the current generation. Rationale for dropping restored-`last`: after a dashboard restart the only thing a `last` generation could do is admit status heartbeats of a job stopped before the restart — but a stopped backend job no longer heartbeats, and the retained data (in-memory) did not survive the restart either. The *runtime* `last` generation is untouched: `begin_generation`/`deactivate` still rotate into it, backing provenance stamps (`resolve_config`) and the just-replaced-job status filter.

`current_job` persistence and `restore` stay — they are how a dashboard restart re-recognizes a running backend job, until heartbeat adoption (slice v, Wave 5, gated on D8).

## Test plan

- `test_stop_workflow_persists_stopped_state` now pins that a stop leaves neither `current_job` nor `previous_job` in the store.
- Integration test rewritten: recommit persists only the new current job; restart restores it.
- Full suite and the `restart_cleanup_test.py -m ""` race gate pass.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -84,7 +84,7 @@ python -m ess.livedata.dashboard.reduction --instrument dummy --transport fake \
 Add `--auto-start` (requires `--transport fake`) to commit every staged workflow on
 launch so plots render with no interaction. Regenerate a fixture by configuring via the
 UI, then copying the persisted `workflow_configs.yaml` (strip the runtime
-`current_job`/`previous_job` keys, keep `jobs`) and `plot_configs.yaml` from the config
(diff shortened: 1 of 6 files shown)


SimonHeybrock (commented):
LGTM

PR #1066, opened 2026-07-21
Remove dead WorkflowController data_service dependency

SimonHeybrock:
Trivial cleanup flagged in the #1042 Wave 3 handoff: `WorkflowController` stored a `DataService` it never used (dead since before slice (i)). Also deletes an unused fixture in `plot_grid_tabs_test.py` that would have raised `TypeError` if exercised (stale `correlation_histogram_controller` argument).


SimonHeybrock (commented):
LGTM

PR #1067, opened 2026-07-21
Expire pending commands whose acknowledgement never arrives

SimonHeybrock:
`KafkaSink.publish_messages` silently swallows serialization and buffer-full errors, so a dropped start/stop/reset command never produces a `CommandAcknowledgement` and its `PendingCommandTracker` entry lived forever: the workflow badge showed "PENDING / Waiting for backend..." indefinitely with no feedback. This is the designated recovery signal for the send-failure mode recorded on #1042 (amendment A2), tracked on #1046 and as the #1039 low finding "pending commands never expire".

The background update loop now expires commands unacknowledged for 30s (deliberately shorter than the 60s heartbeat-staleness constants, which answer a different question), logging a warning and surfacing the existing command-error toast. Workflow state and the status badge are deliberately untouched — point fix; extend to the badge only if the failure mode shows up in practice. A late ack after expiry is ignored gracefully (pinned by test). The tracker takes an injectable clock so expiry is tested without real waits.

### DIFF
--- a/src/ess/livedata/dashboard/dashboard_services.py
+++ b/src/ess/livedata/dashboard/dashboard_services.py
@@ -148,6 +148,10 @@ class DashboardServices:
             start = time.monotonic()
             try:
                 self.orchestrator.update()
+                # Expire pending commands whose acknowledgement never arrived, so
+                # a silently dropped send surfaces as an error toast rather than a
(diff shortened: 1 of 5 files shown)


SimonHeybrock (commented):
LGTM

PR #1068, opened 2026-07-21
dashboard: Suspend hover tooltips while an ROI edit tool is active

### LINKED ISSUE #1063: dashboard: Disable hover info in plots when ROI edit tools are active?
The hover info added in #1032 (showing x,y, and value) gets in the way when editing (drawing or moving) a rectangle or polygon. Can we disable it while these tools are active (and re-enable when deactivating the edit tool)? Not that the mere presence of the tools must not disable the info.

SimonHeybrock:
Fixes #1063

The hover readout (x, y, value) obscures the plot area exactly where the user is drawing or dragging an ROI. Activating a rectangle or polygon edit tool now deactivates the figure's hover tools; leaving the tool restores their previous state, so hover that the user switched off by hand stays off. Merely having the edit tools on the toolbar has no effect on hover.

Bokeh's `Tool.active` is a client-side-only property — tool activation never reaches the server, and `Toolbar.active_drag` is not written back on user clicks — so nothing on the Python side can observe the edit tool becoming active. The toggle is therefore wired in JS, installed via a `CustomJS` on the figure's `inner_width`, which BokehJS writes on every layout pass and which serves as a reliable "figure has rendered" trigger.

The hook is applied to every plot cell alongside the existing SaveTool and autoscale hooks; it is a no-op on figures that carry no edit tool or no hover tool.

## Test plan

Unit tests cover the wiring, the no-op cases, and idempotency across re-renders, plus an integration test that renders a real collated HoloViews overlay and confirms the hook sees the hover tool from the image layer and the BoxEdit tool from the ROI layer on the shared figure.

The JS behaviour itself was verified against real BokehJS in headless Chromium on a standalone document: hover active initially, inactive while the edit tool is active, active again after deactivating.

Still to check manually in the dashboard:

- [x] Drawing and moving a rectangle ROI on a detector view shows no hover tooltip.
- [x] Drawing and moving a polygon ROI shows no hover tooltip.
- [x] Hover works again after deactivating the edit tool.
- [x] Hover switched off manually stays off after using an edit tool.
- [x] Hover is unaffected on a plot where the ROI tools are present but not activated.

### DIFF
--- a/src/ess/livedata/dashboard/hover_suspend.py
+++ b/src/ess/livedata/dashboard/hover_suspend.py
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+"""Suspend plot hover tooltips while a Bokeh edit tool is active.
+
+The hover readout (x, y, value) obscures the plot area exactly where the user
+is drawing or dragging an ROI. This module wires the figure so that activating
+any edit tool (BoxEdit for rectangles, PolyDraw for polygons) deactivates the
+hover tools, and deactivating it restores their previous state. Merely having
+the edit tools on the toolbar does not affect hover.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Any
+
+# Marks a figure as already wired, so the repeated renders of a DynamicMap do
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
Works and LGTM

PR #1083, opened 2026-07-21
Read cell title from TextInput.value, not value_input

### LINKED ISSUE #1072: Saving cell properties without editing the title erases the existing custom title
**Severity:** High

`CellPropertiesModal` pre-fills its title field from the existing custom title:

```python
self._text = pn.widgets.TextInput(
    value=current_title if has_user_title else '',
    ...
)
```

(`src/ess/livedata/dashboard/widgets/cell_properties_modal.py:94-99`)

But `_persist_title()` reads a different attribute:

```python
def _persist_title(self) -> None:
    new_title = (self._text.value_input or '').strip() or None
    self._orchestrator.set_cell_title(self._cell_id, new_title)
```

(lines 143-145)

Panel's `TextInput.value_input` only gets populated by actual browser keystrokes in the field — it is not synchronized from `value` on programmatic pre-fill. So opening the modal on a cell titled "Foo" and clicking Save (or "Add layer…", which also calls `_persist_title` per lines 156/162) without typing anything reads `value_input` as `None`, which `_persist_title` then persists as the cleared title (`None`). The custom title "Foo" is silently erased for all sessions and written out to `plot_configs.yaml`.

**Impact:** any user who opens the cell-properties modal purely to add a layer or check settings, without touching the title field, wipes the cell's custom title on save. This is a common workflow (the modal is also used just to add layers), making the erasure frequent and easy to trigger unintentionally.

**Fix hint:** read `self._text.value` (kept in sync with the field's current displayed content) instead of `value_input`, or initialize `value_input` alongside `value` at construction time.

**Test hint:** instantiate the modal with `has_user_title=True` and a non-empty `current_title`, call `_persist_title()` without setting `value_input`, and assert the orchestrator's stored title is unchanged (not cleared to `None`).

Extracted from #1039.

SimonHeybrock:
Fixes #1072: saving the cell-properties modal (or adding a layer) without typing in the title field erased the existing custom title, because `value_input` only populates on real keystrokes while the pre-fill lives in `value`. Reading `value` covers untouched, edited, and cleared fields; the `value_input or value` fallback was rejected since it would break clearing (empty `value_input` is falsy and would fall through to the stale title).

Also switches the modal's buttons to the `label=`/`color=` Button API — the deprecated kwargs warn on Panel 1.9.3 and `filterwarnings=error` made the modal unconstructable in tests, which is why it had none. New test file pins the regression plus the typed and cleared cases.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/cell_properties_modal.py
+++ b/src/ess/livedata/dashboard/widgets/cell_properties_modal.py
@@ -106,13 +106,11 @@ class CellPropertiesModal:
         self.modal.open = True
 
     def _build(self) -> pn.Modal:
-        self._add_layer_button = pn.widgets.Button(
-            name='Add layer…', button_type='default'
-        )
+        self._add_layer_button = pn.widgets.Button(label='Add layer…', color='default')
         self._add_layer_button.on_click(lambda _: self._on_add())
-        save_button = pn.widgets.Button(name='Save', button_type='primary')
+        save_button = pn.widgets.Button(label='Save', color='primary')
         save_button.on_click(lambda _: self._finish(save=True))
-        cancel_button = pn.widgets.Button(name='Cancel')
+        cancel_button = pn.widgets.Button(label='Cancel')
         cancel_button.on_click(lambda _: self._finish(save=False))
 
         modal = pn.Modal(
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1084, opened 2026-07-21
Fix dead 2D-coordinate branch in slicer flatten mode

### LINKED ISSUE #1074: Slicer flatten mode never preserves the 2D coordinate due to a tuple/list comparison bug
**Severity:** Medium

`SlicerPresenter._flatten_outer_dims` (`src/ess/livedata/dashboard/slicer_plotter.py:227-257`) is meant to keep a genuine 2D coordinate when flattening two of three dimensions, if one exists:

```python
flatten_dims = [d for d in dims if d != keep_dim]   # line 239, a list
...
if (
    coord := data.coords.get(flatten_dims[1])
) is not None and coord.dims == flatten_dims:        # line 253
    flat_dim = flatten_dims[1]
else:
    flat_dim = '/'.join(flatten_dims)
return data.flatten(dims=flatten_dims, to=flat_dim)
```

`coord.dims` on a scipp `Variable`/`DataArray` coordinate is always a `tuple`, but `flatten_dims` is constructed as a `list` (line 239). `tuple == list` is always `False` in Python regardless of contents, so the `coord.dims == flatten_dims` check can never be true — the "keep the 2D coord" branch is dead code, and the flattened axis always falls back to `'/'.join(flatten_dims)`, discarding any real 2D coordinate and showing meaningless `0..N-1` integer indices on the flattened axis instead.

**Fix hint:** compare against `tuple(flatten_dims)`, or build `flatten_dims` as a tuple from the start.

**Test hint:** construct a 3D `DataArray` with a genuine 2D coordinate spanning the two dims that get flattened, call `_flatten_outer_dims` in 'flatten' mode, and assert the result's flattened dimension carries that coordinate (not synthesized integer indices).

Extracted from #1039.

SimonHeybrock:
Fixes #1074: `coord.dims` (a tuple) was compared against a list — always `False` — so the slicer's flatten mode never preserved a genuine 2D coordinate spanning the flattened dims and always fell back to synthesized integer indices on the flattened axis. Regression test pins the coordinate-preserving path; it fails against the old comparison.

### DIFF
--- a/src/ess/livedata/dashboard/slicer_plotter.py
+++ b/src/ess/livedata/dashboard/slicer_plotter.py
@@ -250,7 +250,7 @@ class SlicerPresenter(PresenterBase):
         # the relevant coords.
         if (
             coord := data.coords.get(flatten_dims[1])
-        ) is not None and coord.dims == flatten_dims:
+        ) is not None and coord.dims == tuple(flatten_dims):
             flat_dim = flatten_dims[1]
         else:
             flat_dim = '/'.join(flatten_dims)
--- a/tests/dashboard/plots_test.py
+++ b/tests/dashboard/plots_test.py
@@ -979,6 +979,28 @@ class TestSlicerPlotter:
         assert isinstance(result, hv.Image | hv.QuadMesh)
         assert result.data['values'].shape == (40, 10)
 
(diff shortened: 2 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1085, opened 2026-07-21
Refresh recency on config-store key update

### LINKED ISSUE #1076: Dashboard config store's LRU eviction never refreshes recency on update, behaving as FIFO
**Severity:** Low

Both `InMemoryConfigStore.__setitem__` (`src/ess/livedata/dashboard/config_store.py:86-96`) and `FileBackedConfigStore.__setitem__` (lines 153-166) evict "oldest" entries by insertion order once `max_configs` is exceeded:

```python
def _evict_oldest(self) -> None:
    num_to_remove = max(1, int(len(self.data) * self._cleanup_fraction))
    oldest_keys = list(self.data.keys())[:num_to_remove]
    for key in oldest_keys:
        del self.data[key]
```

Both classes are documented as implementing "LRU (Least Recently Used) eviction" (class docstrings, lines 54 and 108), but `__setitem__` on an already-present key just calls `super().__setitem__(key, value)`, which — for a plain `dict`/`UserDict` — does **not** move that key to the end of iteration order. Python dicts only reorder on deletion+reinsertion, not on updating an existing key's value in place. So a config that was updated moments ago still sits at its original insertion position and can be evicted ahead of entries nobody has touched in a long time — the opposite of what LRU is supposed to guarantee.

**Impact:** for a store under eviction pressure (`max_configs` reached), recently-edited configurations can be silently dropped from persisted state (`plot_grids`, workflow configs, etc.) while genuinely stale, untouched entries survive.

**Fix hint:** on update of an existing key, explicitly move it to the end of recency order before checking the eviction threshold — e.g. `del self.data[key]` then `super().__setitem__(key, value)` again, or switch the backing structure to something with `move_to_end` (like `collections.OrderedDict`).

**Test hint:** create a store with `max_configs=3`, insert keys A, B, C, then update A's value, then insert D (triggering eviction); assert A survives (it was most-recently touched) and the oldest untouched key is evicted instead.

Extracted from #1039.

SimonHeybrock:
Fixes #1076: both config stores document LRU eviction but evicted by plain dict insertion order, which is unchanged when an existing key is updated in place — so a config edited moments ago could be evicted ahead of entries untouched for days. Pop-before-reinsert moves the updated key to the end of iteration order. `OrderedDict` was deliberately avoided: `yaml.safe_dump` in the file-backed store dispatches on `type(data)` and would not serialize it. Regression tests for both store classes fail on the old behavior.

### DIFF
--- a/src/ess/livedata/dashboard/config_store.py
+++ b/src/ess/livedata/dashboard/config_store.py
@@ -88,6 +88,9 @@ class InMemoryConfigStore(UserDict[str, dict[str, Any]]):
         if not isinstance(key, str):
             raise TypeError(f"ConfigStore keys must be str, got {type(key).__name__}")
         with self._lock:
+            # Drop any existing entry first so re-insertion below moves the key
+            # to the end, keeping insertion order aligned with recency of use.
+            self.data.pop(key, None)
             super().__setitem__(key, value)
 
             # Automatic LRU eviction if limit exceeded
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1086, opened 2026-07-21
Return a snapshot from JobService.job_statuses

### LINKED ISSUE #1079: JobService.job_statuses exposes the live status dict, risking a RuntimeError during concurrent iteration
**Severity:** Low

`JobService.job_statuses` (`src/ess/livedata/dashboard/job_service.py:39-42`) returns the live internal dict directly:

```python
@property
def job_statuses(self) -> dict[JobId, JobStatus]:
    """Access to all stored job statuses."""
    return self._job_statuses
```

`status_updated()` (lines 44-49), which mutates this same dict via `self._job_statuses[job_status.job_id] = job_status`, is driven by incoming job-status messages (heartbeats) and can run concurrently with UI-side consumers. The one call site that iterates it, `WorkflowStatusWidget._compute_status` (`src/ess/livedata/dashboard/widgets/workflow_status_widget.py:896`), does `for job_status in self._job_service.job_statuses.values(): ...` on the UI/session thread. If a heartbeat inserts a new `JobId` key while this iteration is in progress, Python raises `RuntimeError: dictionary changed size during iteration`, aborting that widget refresh pass (an update is lost, though the widget recovers on the next poll).

**Fix hint:** have the `job_statuses` property return a shallow copy (`dict(self._job_statuses)`) rather than the live mapping, so callers iterate a stable snapshot.

**Test hint:** assert the property returns a snapshot decoupled from subsequent `status_updated()` calls (mutating after taking the snapshot does not change the snapshot).

Extracted from #1039.

SimonHeybrock:
Fixes #1079: the property handed out the live dict while `status_updated` mutates it on the ingestion thread, so a UI-thread caller iterating it could hit `RuntimeError: dictionary changed size during iteration` when a heartbeat lands mid-pass. The sole consumer (`WorkflowStatusWidget._compute_status`) only iterates, so a shallow copy suffices.

### DIFF
--- a/src/ess/livedata/dashboard/job_service.py
+++ b/src/ess/livedata/dashboard/job_service.py
@@ -38,8 +38,13 @@ class JobService:
 
     @property
     def job_statuses(self) -> dict[JobId, JobStatus]:
-        """Access to all stored job statuses."""
-        return self._job_statuses
+        """Snapshot of all stored job statuses.
+
+        Returns a copy: ``status_updated`` runs on the ingestion thread, so
+        handing out the live dict would let UI-thread callers iterate it while
+        a heartbeat inserts a new key.
+        """
+        return dict(self._job_statuses)
 
     def status_updated(self, job_status: JobStatus) -> None:
         """Update the stored job status and record timestamp."""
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1087, opened 2026-07-21
Cast trim timespan to int64 for datetime64 time coords

### LINKED ISSUE #1069: TemporalBuffer trim crashes on datetime64 time coordinates, permanently breaking the buffer
**Severity:** High

`TemporalBuffer._trim_to_timespan` (`src/ess/livedata/dashboard/temporal_buffers.py:422-443`) computes:

```python
timespan = sc.to_unit(sc.scalar(self._required_timespan, unit='s'), latest_time.unit)
cutoff = latest_time - timespan
```

`timespan` stays a float64 scalar. When the buffer's time coordinate is `datetime64` (as opposed to an int64 offset), scipp rejects subtracting a float64 from a datetime64: `DTypeError: datetime arithmetic requires int64`.

datetime64 time coords do reach this buffer in normal operation — e.g. NXlog-derived streams (`sc.epoch('ns') + times`) and da00-decoded data preserve datetime64. The dashboard already handles this exact case elsewhere: `src/ess/livedata/dashboard/extractors.py:196-198` explicitly casts to int64 before datetime64 arithmetic ("datetime arithmetic requires int64, not float64"), but `temporal_buffers.py` was never given the equivalent cast.

**Impact:** once a windowed buffer with a datetime64 time coordinate reaches `max_capacity` and the trim path first runs, `_trim_to_timespan` raises. Every subsequent `add()` for that buffer key then fails the same way — the plot silently stops updating, permanently, until the dashboard is restarted. Existing tests for this buffer type only use int64 time coordinates, so the crash is not caught by CI.

**Fix hint:** cast `timespan` (or `latest_time`) to a matching int64/datetime64-compatible representation before the subtraction, mirroring the pattern in `extractors.py:196-198`.

**Test hint:** add a `_trim_to_timespan` test where `data.coords['time']` is `dtype='datetime64'`, filling the buffer past `max_capacity` to trigger the trim path, and asserting no exception and correct trimming.

Extracted from #1039.

SimonHeybrock:
Fixes #1069: scipp rejects datetime64-minus-float64, so `TemporalBuffer._trim_to_timespan` raised `DTypeError` the first time a buffer with a datetime64 time coordinate reached capacity — and every subsequent `add()` failed the same way, permanently freezing the plot until dashboard restart. The fix mirrors the int64 cast already used for datetime arithmetic in `extractors.py`. The regression test triggers the trim path with datetime64 coords and also pins that a subsequent `add()` keeps working (the "permanent" half of the bug).

### DIFF
--- a/src/ess/livedata/dashboard/temporal_buffers.py
+++ b/src/ess/livedata/dashboard/temporal_buffers.py
@@ -440,6 +440,9 @@ class TemporalBuffer(BufferProtocol[sc.DataArray]):
         timespan = sc.to_unit(
             sc.scalar(self._required_timespan, unit='s'), latest_time.unit
         )
+        # datetime64 arithmetic requires int64, not float64
+        if latest_time.dtype == sc.DType.datetime64:
+            timespan = timespan.astype('int64')
         cutoff = latest_time - timespan
 
         # Find first index to keep
--- a/tests/dashboard/temporal_buffers_test.py
+++ b/tests/dashboard/temporal_buffers_test.py
@@ -49,6 +49,20 @@ def make_single_slice(x_values, time_value, time_unit='s'):
     )
 
 
+def make_single_slice_datetime(x_values, time_value, time_unit='s'):
+    """Create a single time slice DataArray with a datetime64 time coord.
+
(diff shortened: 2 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1088, opened 2026-07-21
Route tab-to-grid lookups through the tabbed-grid list

### LINKED ISSUE #1071: Disabled plot grids desynchronize tab-to-grid index mapping, freezing visible plots
**Severity:** High

`PlotGridTabs` keeps disabled grids in `self._grid_widgets` (a dict of `GridId -> PlotGrid`) but does not create a visible tab for them (`src/ess/livedata/dashboard/widgets/plot_grid_tabs.py:335-374`, `_reconcile_grid_tabs`, comment at line 350: "Disabled grids are hidden from tabs but we keep the widget reference"). Meanwhile, `_get_active_grid_id` (lines 292-310) and `_remove_grid_tab` (lines 283-290) both resolve tab positions by indexing directly into `list(self._grid_widgets.keys())`, which still includes the hidden disabled entries.

**Concrete sequence:** with grid A disabled and grid B enabled, only one tab exists for B, but `list(self._grid_widgets.keys())` is `[A, B]`. Viewing (the only) tab resolves `grid_idx = 0`, which maps to `A` — the disabled, non-visible grid — not `B`, the grid actually on screen. Consequences:

- `_get_active_grid_id()` returns the wrong grid, so `update_pipe()` never runs for the grid that's actually visible and its interest/viewer token is never acquired — the visible tab's plots silently freeze (compute may stop entirely for that grid).
- `_remove_grid_tab()` (lines 283-290) computes `tab_index` the same way, so removing a grid can pop the wrong tab from the Bokeh `Tabs` widget, or raise `IndexError` (swallowed elsewhere), leaving a stale grid's tab on screen.

Existing tests only exercise disabling the *last* grid in the list, which happens to not skew earlier indices, so the bug escapes test coverage.

**Fix hint:** the index mapping from Bokeh tab position to `GridId` must only consider *enabled* grids (i.e., filter `self._grid_widgets` by enabled state, or maintain a separate ordered list of the grids that actually have tabs) before doing positional lookups.

**Test hint:** create three grids where the *first* or a *middle* grid is disabled (not the last), and assert `_get_active_grid_id()` and `_remove_grid_tab()` resolve the correct `GridId` for each visible tab index.

Extracted from #1039.

SimonHeybrock:
Fixes #1071: disabled grids keep their widget in `_grid_widgets` (preserving state across re-enable) but have no tab, so positional lookups indexing `list(_grid_widgets.keys())` skewed once a disabled grid preceded an enabled one — the visible tab resolved to the wrong grid, silently freezing its plots (`update_pipe` never ran for the grid actually on screen), and removing a grid could pop the wrong tab. Three positional call sites were affected (`_get_active_grid_id`, `_remove_grid_tab`, and `_on_grid_created`'s auto-switch, one more than the issue named); all now go through a single helper listing grids that actually have tabs.

Tab membership is derived from the live tab panels rather than the orchestrator's enabled flag: `remove_grid` drops the grid from the orchestrator before `on_grid_removed` fires, so the enabled flag is already gone for the grid whose tab must still be popped (this broke the removal path in a first attempt using the flag).

New tests disable the first/middle grid of three — the existing suite only ever disabled the last one, which happens not to skew; that gap is why this escaped.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
+++ b/src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
@@ -280,13 +280,35 @@ class PlotGridTabs:
                 cell=cell,
             )
 
+    def _tabbed_grid_ids(self) -> list[GridId]:
+        """GridIds that currently have a tab, in tab order.
+
+        Disabled grids keep a widget in ``_grid_widgets`` (preserving state
+        across re-enable) but have no tab. Positional lookups between a Bokeh
+        tab index and a GridId must go through this list, never through
+        ``_grid_widgets`` directly, or a disabled grid preceding an enabled one
+        skews the mapping. Membership is read from the live tabs -- the grid
+        panels actually appended -- so it stays correct even mid-removal, when
+        the orchestrator has already dropped the grid being removed.
+        """
+        panel_to_grid = {
+            id(grid.panel): grid_id for grid_id, grid in self._grid_widgets.items()
+        }
+        tabbed: list[GridId] = []
+        for tab in self._tabs[self._static_tabs_count :]:
+            grid_id = panel_to_grid.get(id(tab))
+            if grid_id is not None:
+                tabbed.append(grid_id)
+        return tabbed
+
     def _remove_grid_tab(self, grid_id: GridId) -> None:
         """Remove a grid tab."""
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1089, opened 2026-07-21
Share live ROI state across sessions instead of republishing per session

### LINKED ISSUE #1070: New dashboard sessions republish the initial ROI set, wiping user-drawn ROIs on the backend
**Severity:** High

`_create_edit_handler()` in `src/ess/livedata/dashboard/roi_request_plots.py:635-673` unconditionally publishes the config-time ROI set at the end of handler creation:

```python
current_rois: dict[int, ROIType] = dict(self._initial_rois)   # line 648
...
self._publish_rois(current_rois)                              # line 672
```

`_create_edit_handler()` is called once per browser session (from `create_presenter()`, lines 753-758 and 926-931). `_initial_rois` is parsed once from config at plotter construction (line 550) and never updated as the user edits ROIs. Publishing an ROI dict is destructive: an empty or stale dict is treated by the backend as "clear all requested regions" (`roi_publisher.py:48`).

**Concrete sequence:** user A opens a session and draws ROIs; the backend accumulates ROI spectra for them. User B opens the dashboard, or A reloads the tab — either creates a new `_create_edit_handler()` call, which republishes `_initial_rois` (typically empty), silently deleting all of A's ROIs on the backend. A's browser still shows the (now-orphaned) rectangles until their next edit resurrects the mismatch.

**Impact:** any second session (multi-user viewing, or a single user reloading) destroys ROI state for everyone. This is a plotter-per-layer, multi-session-shared component (`session_layer.py:87`), so the hazard triggers on ordinary multi-tab usage, not just concurrent distinct users.

**Fix hint:** ROI state and the "have I published yet" flag need to live at plotter-instance scope (shared across sessions, not the per-session edit-handler closure), and publishing should only happen once at plot-config creation/edit — never at session/presenter creation. Session presenters and edit baselines should read from that shared live state, not from `_initial_rois`, so a session opened after edits sees the current ROIs rather than the config-time set.

**Test hint:** create a plotter, call `create_presenter()` twice (simulating two sessions) after simulating a user edit via the first presenter's `on_edit` callback, and assert the second `create_presenter()` call does not trigger a new `_publish_rois` call that reverts to `_initial_rois`.

Extracted from #1039.

SimonHeybrock:
Fixes #1070: each browser session's edit handler kept a closure copy of the ROI set seeded from the config-time initial ROIs and unconditionally published it at presenter creation. Publishing replaces the backend's requested set, so any new session or tab reload wiped everyone's user-drawn ROIs.

ROI state now lives at plotter scope, shared by all sessions' edit handlers under a lock (edit events fire on each session's own document thread; the presenter-seeding read is lock-free since the handler rebinds a fresh dict atomically). The initial set is published once per plotter lifetime, on the first presenter creation after `compute()` has set the data key — that ordering is guaranteed because `create_presenter` only runs once `has_displayable_plot()` is true, which requires the same `compute()` call that sets the key. The once-flag is only set when a message actually went out, so an unwired publisher retries on the next presenter creation. New presenters seed their drawn shapes from the live state, so a session opened after edits shows the current ROIs.

Known residual (pre-existing, filed separately): a dashboard restart rebuilds plotters and republishes the config-time set, clobbering backend-side edits, because edits are never persisted back to the plot config.

### DIFF
--- a/src/ess/livedata/dashboard/roi_request_plots.py
+++ b/src/ess/livedata/dashboard/roi_request_plots.py
@@ -17,7 +17,8 @@ ROI request plotters follow the two-stage compute/present pattern:
    forwards the raw data. Called once when data arrives.
 
 2. create_presenter(): Creates a presenter with HoloViews config and an edit
-   handler callback. The callback captures per-session ROI state in a closure.
+   handler callback. The callback reads and updates the plotter's live ROI
+   state, which is shared across all sessions.
 
 3. Presenter.__init__(): Creates session-bound Pipe, DynamicMap, and edit
    streams. Each browser session gets its own presenter instance.
@@ -31,6 +32,7 @@ comparison, skip logic, publishing) stays in the plotter via the edit callback.
 
 from __future__ import annotations
 
+import threading
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, Generic, Protocol, TypeVar, runtime_checkable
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
Confirmed in the UI with two browsers. LGTM

PR #1094, opened 2026-07-21
Express frame-aspect sizing via opts, attach hook JS once

### LINKED ISSUE #1092: Fixed-data-aspect plots flicker to a collapsed frame on every data update
## Symptom

On a plot with `aspect_type="Fixed data aspect ratio"`, the plot briefly jumps to a collapsed width (or height, depending on `stretch_mode`) and then snaps back to the correct aspect. It is most noticeable when a job starts and initial data arrives.

## What actually happens

It is not a one-off at job start — it happens on **every** data update. Recorded Bokeh figure geometry (data-aspect, Fill width, dummy fixture, fake backend):

```
32676  Figure.inner_width    h=400 iw=429 ih=316   <- hook seeds 400
32795  Figure.inner_height   h=513 iw=429 ih=429   <- CustomJS corrects, aspect right
33987  Figure.inner_height   h=400 iw=429 ih=316   <- data update: reset to 400
34040  Figure.inner_height   h=513 iw=429 ih=429   <- CustomJS corrects again
34990  ...400...  35067  ...513...  36010  ...400...  36091  ...513...
```

The seed is 400px, so on a cell whose correct size is much larger the intermediate frame reads as "collapsed".

## Root cause

`hook()` in `frame_aspect.py` is a HoloViews hook, and HoloViews runs `_execute_hooks` from **`update_frame`**, not only from `initialize_plot` (see `holoviews/plotting/bokeh/element.py`, the call at the end of `_update_plot`). So every incoming frame re-runs the hook, which re-seeds `fig.height = 400` / `fig.width = 400`.

That seed is a *real* server-side change, because the size the `CustomJS` sets in the browser syncs back to the server model. So each update patches the figure back to 400px and the JS snaps it out again — one frame of wrong geometry per update.

## Why the obvious fix does not work

Tagging the figure so the hook body runs once collapses the plots to zero height instead. The repeated hook execution is currently **load-bearing**: Panel stamps the enclosing pane's `sizing_mode` (`stretch_both`) onto the figure *after* the initial render, and only a subsequent hook run restores `stretch_width`. Splitting the body (re-assert `sizing_mode` on every run, seed + attach `CustomJS` once) also fails, collapsing width instead.

So the hook and Panel are duelling over `sizing_mode`, and the flicker is a symptom of that duel. A fix likely has to stop Panel from stomping in the first place — set the pane's `sizing_mode` where the `pn.pane.HoloViews` is constructed in `widgets/cell.py` to match the configured `StretchMode`, rather than letting the hook fight it — not merely make the hook idempotent.

Possibly related: #580 (closed). A Fill-height run during this investigation showed the same zero-width collapse described there.

## Reproduction

Fake backend, no Kafka:

1. Copy `tests/dashboard/ui_config_fixtures/dummy` to a scratch config dir and replace `aspect_type: Square` with `aspect_type: Fixed data aspect ratio`.
2. Launch `python -m ess.livedata.dashboard.reduction --instrument dummy --transport fake --port 5017 --config-dir <dir> --no-fetch-announcements`.
3. Open two Playwright sessions: one parked on the Detectors tab, one on Workflows to press play. Tab switching would destroy the models under observation, and `requestAnimationFrame` is throttled in a background tab, so poll-based recorders miss the transient.
4. In the viewer session, before any job runs, install a recorder via `Bokeh.documents[0].on_change(...)` filtering `Figure` / `Range` events on `width`, `height`, `inner_*`, `outer_*`, `sizing_mode`, `start`, `end`.
5. Press play in the driver session and dump the recorded trace.

The 400/513 oscillation above appears once per update, indefinitely.

### LINKED ISSUE #1093: Frame-aspect hook leaks a CustomJS callback set on every data update
## Problem

`hook()` in `frame_aspect.py` attaches a fresh `CustomJS` on every invocation:

```python
callback = CustomJS(args={"fig": fig, **js_args}, code=...)
fig.js_on_change("inner_width", callback)
fig.js_on_change("inner_height", callback)
if listen_ranges:
    fig.x_range.js_on_change("start", callback)
    fig.x_range.js_on_change("end", callback)
    fig.y_range.js_on_change("start", callback)
    fig.y_range.js_on_change("end", callback)
```

HoloViews runs `_execute_hooks` from `update_frame`, not only from `initialize_plot`, so this runs once per incoming data frame. Nothing ever detaches the previous callback.

## Measurement

Counting `js_property_callbacks` in the browser after ~20s of updates on the dummy fixture (fake backend, data-aspect plots):

```
js callbacks per figure: [{'fig': 42, 'xr': 42}, {'fig': 42, 'xr': 42}]
```

Growth is unbounded and linear in job runtime. Every one of them fires on every layout/range change, so the cost is quadratic in the number of updates over a session. Long-running dashboard sessions are the exposed case.

## Notes

This is independent of the visual flicker in #1092, though both stem from the same "hooks re-run on every update" fact. A one-shot guard around the attach fixes this cleanly on its own; the flicker needs more than that (Panel re-stamps `sizing_mode` after the initial render, so the hook body cannot simply become one-shot in full).

SimonHeybrock:
Fixes #1092
Fixes #1093

## What and why

Plots with a non-free aspect flickered to a collapsed frame on every data update, and the frame-aspect hook leaked a CustomJS callback set per update (quadratic client-side cost over a session).

Instrumenting all server-side writes to `Figure.sizing_mode/height/width` showed the root cause is a duel between the hook and HoloViews — not Panel as #1092 hypothesized: `ElementPlot._update_plot` recomputes plot properties from the *opts* (`responsive=True` with no dimensions → `stretch_both`) and re-applies them to the figure, while the hook wrote `stretch_width`/`stretch_height` plus a 400px seed on every execution (HoloViews runs hooks from `update_frame`, not only `initialize_plot`).

The fix expresses sizing where HoloViews looks for it: `responsive=True` plus a fixed cross dimension (`height` for Fill width, `width` for Fill height), from which HoloViews derives `stretch_width`/`stretch_height` itself and re-applies it stably on every update. The cross dimension doubles as the initial seed and is applied only at initialization, so the browser-adjusted size survives updates. The hook shrinks to a one-shot CustomJS attach guarded by a figure tag, fixing the leak. Leaf sizing opts propagate to `OverlayPlot`, so multi-layer cells behave the same.

## Test plan

- New unit tests drive a real Pipe-fed DynamicMap through updates: callback count stays constant, and a browser-adjusted height survives updates (both fail on the previous implementation).
- Playwright harness (fake backend, data-aspect config): server-side writes drop to three per figure at init with zero afterwards; figure geometry bit-stable across ~18s of updates; JS callback count constant vs. growth to ~32/figure in 20s before.
- [x] Manual browser verification of data-aspect plots across job start and updates.

### DIFF
--- a/src/ess/livedata/dashboard/frame_aspect.py
+++ b/src/ess/livedata/dashboard/frame_aspect.py
@@ -1,19 +1,30 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
-"""HoloViews hook that enforces frame aspect ratios on Bokeh plots.
+"""HoloViews sizing opts that enforce frame aspect ratios on Bokeh plots.
 
 HoloViews' aspect options (``aspect="square"``, ``data_aspect``, etc.) do not
 produce correctly shaped data areas (frames) when plots use ``responsive=True``
 inside Panel containers.  This is an upstream bug spanning Bokeh, HoloViews,
 and Panel.
 
-The workaround uses HoloViews *hooks* that:
+The workaround, assembled by :func:`make_frame_aspect_opts`, has two
+cooperating parts:
 
-1. Switch the Bokeh figure to ``stretch_width`` or ``stretch_height``
-   (chosen by the user via :class:`StretchMode`) so it fills the container
-   along one axis.
-2. Attach a ``CustomJS`` callback that adjusts the *other* dimension so the
-   frame has the correct shape.
+1. Sizing opts: ``responsive=True`` plus a fixed cross dimension (an initial
+   ``height`` for :attr:`StretchMode.width`, ``width`` for
+   :attr:`StretchMode.height`).  HoloViews derives ``stretch_width`` (resp.
+   ``stretch_height``) from this combination, so the figure fills the
+   container along one axis.  Expressing the sizing mode through opts rather
+   than writing it to the figure is essential: HoloViews recomputes plot
+   properties from the opts whenever they change (``ElementPlot's
+   _update_plot``), which would overwrite figure-level writes.
+2. A hook attaching a ``CustomJS`` callback that adjusts the cross dimension
+   in the browser so the frame has the correct shape.  HoloViews runs hooks
+   from ``update_frame`` (i.e. on every data update), not only from
+   ``initialize_plot``, so the hook tags the figure and only acts once per
+   figure: repeated attaching would leak a callback set per update, and
+   re-seeding the cross dimension would collapse the figure to the seed size
+   for one layout frame per update.
 
 Two hook variants exist:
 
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1095, opened 2026-07-21
Split session teardown into two tiers for reaper safety

### LINKED ISSUE #955: Dashboard: stale-session cleanup mutates Bokeh document from a background thread (unsafe)

SimonHeybrock:
## What

Splits session teardown into two tiers so the stale-session reaper never mutates Bokeh document state from the background thread, and so shared-state severing runs even when a session's IOLoop tick never fires.

- **Tier 2 — shared-state severing** (`SessionUpdater.cleanup` inline, `register_cleanup_handler`): releases the notification-queue slot, lifecycle subscriptions, and the session's viewer/interest tokens. Thread-safe, unconditional, independent of tier 1.
- **Tier 1 — document mutation** (`register_document_teardown_handler` + `PeriodicCallback.stop()`): on the reaper path scheduled onto the session's IOLoop via the captured `Document.add_next_tick_callback` (scheduling into a destroyed document is tolerated); the clean `on_session_destroyed` path runs it inline.

`PlotGridTabs.shutdown()` is split along the same line into `sever()`/`dispose_widgets()`, self-registered on the `SessionUpdater`.

## Why

The reaper called `updater.cleanup()` → `PeriodicCallback.stop()` on the background update thread — a document mutation off the IOLoop, racing the session's model graph. This is the prime suspect for the Bifrost PROD single-core CPU spin after abnormal websocket closes (leaked `periodic_update` callbacks recomputing plots for dead sessions) and matches the "Dropping a patch ... previously known reference" production warnings. The reaper fires precisely when the session is abnormal, so an IOLoop tick may never run — hence severing must not depend on it.

Grid-widget teardown was additionally wired only to `on_session_destroyed`, so abnormal disconnects leaked viewer/interest tokens and lifecycle subscriptions (old audit finding: hidden layers keep computing). Routing both tiers through the `SessionUpdater` closes that leak on both paths.

## Test plan

- Regression tests fail when the fix is reverted: reaper must not stop the periodic callback on the calling thread; tier-2 severing happens even if the document tick never runs; destroyed-document scheduling is tolerated; clean path still tears down inline.
- End-to-end reaper test with real `PlotGridTabs` + orchestrator + `SessionRegistry`.
- [ ] Field check on a real backend: abnormal disconnect (kill browser) leaves no lingering CPU load and no "Dropping a patch" warnings.

Fixes #955

### DIFF
--- a/src/ess/livedata/dashboard/dashboard.py
+++ b/src/ess/livedata/dashboard/dashboard.py
@@ -161,6 +161,7 @@ class DashboardBase(ServiceBase, ABC):
             session_registry=self._services.session_registry,
             notification_queue=self._services.notification_queue,
             username=pn.state.user,
+            document=pn.state.curdoc,
         )
 
     def _start_periodic_callback(
--- a/src/ess/livedata/dashboard/reduction.py
+++ b/src/ess/livedata/dashboard/reduction.py
@@ -181,10 +181,10 @@ class ReductionApp(DashboardBase):
             session_updater=session_updater,
         )
 
-        # Release this session's interest tokens when the browser session ends,
-        # so hidden layers stop computing promptly rather than waiting for the
-        # SessionLayer weakref finalizer to run at some later GC.
-        pn.state.on_session_destroyed(lambda _ctx: plot_grid_tabs.shutdown())
+        # PlotGridTabs registers its own two-tier teardown on the session
+        # updater (see PlotGridTabs.__init__), so it runs on both the clean
+        # disconnect and the stale-session reaper paths, releasing this
+        # session's interest tokens when the browser session ends.
 
         # Register refresh with visibility gate: skip updates when the
         # Workflows tab (index 0) is not the active tab.
--- a/src/ess/livedata/dashboard/session_registry.py
+++ b/src/ess/livedata/dashboard/session_registry.py
@@ -169,11 +169,14 @@ class SessionRegistry:
                         now - info.last_heartbeat,
                     )
 
-        # Clean up updaters outside lock to avoid potential deadlocks
+        # Clean up updaters outside lock to avoid potential deadlocks. This runs
+        # on the background update thread, off the sessions' IOLoops, so
+        # document-mutating teardown must be marshalled onto each session's
+        # IOLoop rather than run here (see SessionUpdater.cleanup).
         for session_id, updater in stale_sessions:
             if updater is not None:
                 try:
-                    updater.cleanup()
+                    updater.cleanup(defer_document_teardown=True)
                 except Exception:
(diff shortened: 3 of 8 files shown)


SimonHeybrock (commented):
LGTM

Issue #1097, opened 2026-07-21
Field verification (real backend) of the dashboard delivery restructure and point fixes


SimonHeybrock:
The dashboard delivery restructure (#1046: single-writer versioned pull per ADR 0007, job adoption per ADR 0008) and the associated point fixes were verified against the fake backend and Playwright only. This was the checklist for one verification session against a real backend. Most items are now covered by automated tests (#1112 real-Kafka integration job in CI; #1113 browser UI tests, run in CI by #1118; #1114 and #1119 backend-semantics integration tests), so the remaining manual session shrinks to the short list at the bottom.

## Checklist

- [x] Late joiner (old #789 class): open a second session while a workflow runs — it shows populated plots immediately. *Automated: `tests/dashboard/multisession_smoke_test.py` (browser).*
- [x] Clear-at-commit: a recommit drops the colorbar max by the accumulated factor immediately; stop retains the frame; restart resets it. *Backend semantics automated: `tests/integration/clear_at_commit_test.py` (recommit drops the accumulated total).*
- [ ] Command-ACK expiry (#1067): expect *no* expiry toasts in normal operation. Any toast is a real lost send — record it here. *The expiry mechanism itself is automated (`tests/integration/command_expiry_test.py`); what remains manual is the soak observation that no toasts appear at real rates. Note: #1112 fixed a real lost-command bug (manual partition assignment resolved `latest` lazily, silently skipping messages produced before the first fetch) — a likely cause of exactly such toasts.*
- [ ] ROI sharing (#1089): draw ROIs in session A; open session B (sees them); reload A — backend ROI spectra survive both. *Backend half automated: `tests/integration/roi_spectra_test.py` publishes ROIs via the real `ROIPublisher` path and asserts `roi_spectra` results track the ROI set. Remaining manual: cross-session UI visibility and reload, one two-session glance.*
- [x] datetime64 buffers (#1087): a windowed timeseries/logdata plot keeps updating after running past buffer capacity. *Automated: sustained-wraparound test in `tests/dashboard/temporal_buffers_test.py`.*
- [x] Disabled grids (#1088): disable the *first* grid of several — remaining tabs resolve and update; then remove a middle grid. *Automated: `tests/dashboard/browser_ui_test.py`.*
- [x] Cell title (#1083): open the properties modal on a titled cell, Save without typing — title intact. *Automated: `tests/dashboard/browser_ui_test.py`.*
[...]

PR #1098, opened 2026-07-21
Move grid lifecycle from cross-session push to version polling

### LINKED ISSUE #1040: PlotOrchestrator shared state has no thread-safety discipline (background thread vs UI; cross-session widget push)

SimonHeybrock:
## What

Converts the last dashboard widgets still on the cross-session push model — the grid tabs and grid manager — to the versioned-pull pattern of ADR 0007, and gives the shared topology dicts an explicit thread-safety discipline. This completes the dashboard-side restructure tracked by #1046.

- `PlotOrchestrator` mutators bump a single topology version; the subscribe/notify lifecycle machinery is deleted. Each session's `PlotGridTabs` polls the version from its own periodic tick and reconciles tabs under its own document lock, restoring the active tab by identity. Cell composition changes (layer add/remove/reconfigure, title) are detected per tick via cell signatures; vanished cells are swept and disposed. Cells of a merely disabled grid survive for re-enable.
- Tab-focus semantics change deliberately: auto-switching to a new grid happens only in the session that created it (`PlotGridManager` reports local creations to its owning `PlotGridTabs`); other sessions just gain the tab. Previously grid creation set `tabs.active` in every session.
- Thread-safety (hazard 1 of #1040): the IOLoop thread remains the sole topology writer; a narrow `RLock` makes the ingestion thread's multi-dict traversals see consistent snapshots. The lock stays off persist/compute/pipeline paths; the discipline (lock order, leaf-first deletes) is documented on the class.
- `rename_grid`/`move_grid`/`set_grid_enabled` now tolerate ids removed by another session inside the poll window instead of raising.

## Why

Lifecycle notifications fanned out to every session's widgets on the *mutating* session's thread: session A's `add_grid` rebuilt session B's Bokeh document outside B's document lock (intermittent cross-session corruption), and hijacked B's active tab. Per-layer data already moved to versioned pull; lifecycle was the remaining push surface. The topology dicts were additionally read by the ingestion thread with no consistency guarantee across the three mappings.

## Test plan

- Lifecycle-notification tests rewritten as topology-version tests; widget tests drive the poll explicitly.
- New cross-session regression tests: grid created via session A focuses the tab in A only, B's active tab unchanged (fails on the push model); direct orchestrator creation steals no focus; `replace_grid` keeps position without stealing focus; cell reconcile add/remove/retitle; ingestion read paths tolerate concurrent layer removal; disabled-grid cells are not disposed.
- Full fast suite passes; `pytest -m browser` multisession smoke test passes on this branch.
- [ ] Field verification items on #1097 (cross-session tab isolation with a real backend).

Fixes #1040

### DIFF
--- a/src/ess/livedata/dashboard/plot_orchestrator.py
+++ b/src/ess/livedata/dashboard/plot_orchestrator.py
@@ -476,7 +476,7 @@ class PlotOrchestrator:
         grid = PlotGridConfig(title=title, nrows=nrows, ncols=ncols)
         with self._topology_lock:
             self._grids[grid_id] = grid
-            self._bump_topology_version()
+        self._bump_topology_version()
         self._persist_to_store()
         self._logger.info(
             'Added plot grid %s (%s) with size %dx%d', grid_id, title, nrows, ncols
@@ -502,7 +502,7 @@ class PlotOrchestrator:
 
             with self._topology_lock:
                 del self._grids[grid_id]
-                self._bump_topology_version()
+            self._bump_topology_version()
             self._persist_to_store()
             self._logger.info('Removed plot grid %s (%s)', grid_id, title)
 
@@ -551,7 +551,7 @@ class PlotOrchestrator:
         keys.insert(new_index, keys.pop(current_index))
         with self._topology_lock:
             self._grids = {k: self._grids[k] for k in keys}
-            self._bump_topology_version()
+        self._bump_topology_version()
         self._persist_to_store()
         self._logger.info('Moved grid %s by %d positions', grid_id, delta)
 
@@ -632,7 +632,7 @@ class PlotOrchestrator:
             items = list(self._grids.items())
             items.insert(position, (new_grid_id, new_grid))
             self._grids = dict(items)
-            self._bump_topology_version()
+        self._bump_topology_version()
 
         self._persist_to_store()
         self._logger.info(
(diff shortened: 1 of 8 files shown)


SimonHeybrock:
LGTM

PR #1099, opened 2026-07-21
Add developer glossary

SimonHeybrock:
Adds an authoritative glossary for project terminology. We have accumulated enough overloaded words (role, stream, view, orchestrator, frame, config, ...) that reading prose or reviewing PRs requires tribal knowledge to resolve which sense is meant; the glossary pins down each term and ends with a disambiguation section for words carrying several meanings.

The glossary lives next to the code so grep and directory listings surface it during code work: `src/ess/livedata/glossary.md` (cross-cutting + backend + disambiguation) and `src/ess/livedata/dashboard/glossary.md`. `docs/developer/glossary.md` is a stub of myst includes rendering both as one page in the docs. Class names get entries only when the class is a concept referenced outside its module (or collides with another term) — wiring classes are folded into parent entries.

Writing it surfaced places where existing docs and docstrings reference concepts that no longer exist; those are corrected here so nothing in the docs tree contradicts the glossary: `dashboard-architecture.md` claimed `DataService` is keyed by `ResultKey` (it is keyed by `DataKey`), the `StreamKind` list in `message-flow-and-transformation.md` was missing three members, and docstrings referenced the removed `SessionPlotManager` and `ConfigHandler` classes.

The audit findings were collected in #1100; the decided renames small enough to land here are included in this PR (see table), while the handler-layer rename is split into #1104 (stacked on this branch) and the stale `stream-registry.md` replacement into #1103.

## How to refer to things now

| Was | Is | Notes |
|---|---|---|
| `Job(processor=...)` / `Job._processor` | `Job(workflow=...)` / `Job._workflow` | "processor" stays reserved for the `Processor` protocol driven by `Service` |
| "slot" (a plot layer's data-source key) | *data role* (`primary`/`x_axis`/`y_axis`) | wording + `plot_orchestrator.py` internals; matches `data_roles.py` and the glossary |
| `DataSourceConfig.view_name` overwritten in place with the resolved backend field name | `ResolvedDataSource.output_name` (runtime-only) | `view_name` now always means the user-facing `OutputView` name; persisted YAML schema unchanged |
| `JobManager.service_name` | *(removed)* | dead state, hardcoded `'data_reduction'`; the real name lives on `JobFactory`/`OrchestratingProcessor` |
| `pooch.os_cache('beamlime')`, benchmark URLs under `scipp/beamlime/` | `'esslivedata'` / `scipp/esslivedata/` | server serves both names via symlink; old local caches are simply orphaned |
| `StreamRole` | `Windowing` | the `since_start`/`per_update` flavor of an output field is windowing, not a role; "role" now means data role or aux-input role only |
| `OutputView.streams` | `OutputView.fields` | it maps windowing to backend pydantic *field names* (`field_for` said so all along); not persisted or sent over the wire |
| dashboard `Orchestrator` (`dashboard/orchestrator.py`) | `MessagePump` (`dashboard/message_pump.py`) | the last unqualified "orchestrator"; remaining ones (`OrchestratingProcessor`, `JobOrchestrator`, `PlotOrchestrator`) are all qualified |

Not in this PR but part of the same effort: `*HandlerFactory` → `*PreprocessorFactory`, `handlers/` → `preprocessors/`, `core/handler.py` → `core/preprocessor.py`, `LatestValueHandler` → `LatestValueAccumulator`, `ConfigProcessor` → `CommandDispatcher` (`core/command_dispatcher.py`) — all in #1104.

Draft because the glossary is only useful if it matches how the team actually uses these words — please read it as a proposal for the authoritative senses, not a description I am certain of.

## Test plan

- `tox -e docs` builds; the only warning is an unrelated intersphinx network failure in the sandbox. The stitched glossary page renders with correct section order (cross-cutting, backend, dashboard, disambiguation) and working anchors.
- Wheel contents unaffected: `[tool.setuptools.package-data]` uses explicit per-package globs that do not match the new `.md` files.
- Full fast suite green after each rename batch.


SimonHeybrock (commented):
LGTM

PR #1102, opened 2026-07-22
Fix _job_states leak on workflow recommit (#1080)

### LINKED ISSUE #1080: JobOrchestrator leaks one job-state entry per source on every workflow restart-commit
**Severity:** Low

`JobOrchestrator` tracks the latest reported state per job in `self._job_states: dict[JobId, JobState]` (`src/ess/livedata/dashboard/job_orchestrator.py:153`). Entries are written in `on_job_status_updated` (line 791) and are only ever removed in `_deactivate_workflow` (line 768), which is called from an explicit `stop_workflow()` or once the backend confirms all jobs of a workflow are fully stopped.

`commit_workflow()` (lines 335-448) — used both for first-time start and for *recommitting an already-running workflow* — does not go through `_deactivate_workflow` when replacing an active job set. On a recommit, it sends fire-and-forget stop commands for the old job IDs (lines 369-382) and then directly overwrites the workflow's tracked job set:

```python
state.current = job_set   # line 410, no _job_states cleanup for the old job_ids
```

Because `JobId` embeds a per-commit `job_number`, each recommit's job IDs are new dictionary keys — the old job IDs' entries in `_job_states` are never popped. Worse, once `state.current` has moved on, `on_job_status_updated`'s guard `if job_status.job_id not in job_ids: return` (job_ids being computed from the *new* `state.current`) means the old job's own final "stopped" heartbeat — which would otherwise be the natural point to clean it up — is dropped before it's even recorded. The old entry is permanently orphaned.

**Impact:** for a long-running dashboard where workflows are recommitted repeatedly (routine operation — parameter changes, restarts), `_job_states` grows by one entry per source per recommit, unbounded, for the lifetime of the dashboard process. Individual entries are small (an enum value keyed by a small dataclass), so this is a slow memory leak rather than an acute one, but it is unbounded over multi-day operation.

**Fix hint:** when `commit_workflow()` replaces an active `state.current` with a new job set, pop the old job set's IDs from `_job_states` (the same cleanup `_deactivate_workflow` already does), rather than only doing so on explicit stop / full-stop confirmation.

**Test hint:** commit a workflow, record `_job_states` size after some `on_job_status_updated` calls, commit the *same* workflow again (simulating a restart-commit) with fresh job IDs, and assert `_job_states` does not retain entries for the now-superseded job IDs.

Extracted from #1039.

SimonHeybrock:
`commit_workflow` replaces the active job set but left the superseded generation's entries in `_job_states`. Because `job_number` is fresh per commit, the old jobs' final heartbeats fall outside the current set, so `on_job_status_updated` ignores them and they never self-clean. Tracked states therefore grew by one entry per source on every restart-commit — a routine operation (parameter changes, restarts) — unbounded over the lifetime of the dashboard process.

The fix pops the old job set's IDs on recommit, mirroring the cleanup `_deactivate_workflow` already performs on explicit stop. This area was heavily refactored since the issue was filed (stable-key generation model), so line numbers in #1080 are stale, but the leak was still live — confirmed with a regression test against real components that fails on `main` and passes with this change.

Closes #1080.

### DIFF
--- a/src/ess/livedata/dashboard/job_orchestrator.py
+++ b/src/ess/livedata/dashboard/job_orchestrator.py
@@ -505,6 +505,14 @@ class JobOrchestrator:
             message_id, workflow_id, "start", expected_count=len(state.staged_jobs)
         )
 
+        # Drop the superseded generation's tracked states. The old jobs have
+        # been stopped above; their final heartbeats are no longer in the
+        # current set, so on_job_status_updated would ignore them and never
+        # clean these up (the same cleanup _deactivate_workflow performs).
+        if state.current is not None:
+            for job_id in state.current.job_ids():
+                self._job_states.pop(job_id, None)
+
         state.commit(job_set)
 
         # Flip the generation, clear the workflow's buffers, and notify
--- a/tests/dashboard/restart_cleanup_test.py
+++ b/tests/dashboard/restart_cleanup_test.py
@@ -252,6 +252,29 @@ class TestRestartCleanup:
             == job_ids_3[0].job_number
         )
 
+    def test_recommit_does_not_leak_superseded_job_states(self, workflow_spec):
+        """A recommit drops the superseded generation's tracked job states.
+
+        Regression for #1080: job_number is fresh per commit, so recommit
+        keys are new; the old jobs' final heartbeats fall outside the current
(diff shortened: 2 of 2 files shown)


SimonHeybrock (commented):
LGTM

Issue #1100, opened 2026-07-21
Naming and terminology inconsistencies (glossary audit)


SimonHeybrock:
While writing the developer glossary (#1099), I audited the repo for inconsistent naming. Trivial corrections (stale docstrings, wrong key type in dashboard-architecture.md, incomplete StreamKind list) are fixed in that PR. The remaining findings need a decision or non-trivial work; listed here grouped by theme. Severity in brackets.

## Stale design doc

- [high] `docs/developer/design/stream-registry.md` documents types and methods that do not exist: `LogContextBinding`, `Instrument.log_context_bindings`, `Instrument.get_context_keys`, `TransformValueLog`, `StreamProcessorWorkflow.context_keys`. The code has `ContextBinding` (`config/stream.py`), `Instrument.context_bindings`, `resolve_context_keys`, `ValueLog`; ADR 0003 explicitly removed `context_keys` from the workflow protocol. The doc predates ADR 0003/0004 and needs a rewrite (or trimming down to what the ADRs don't cover).

## Code naming — decisions needed

- [high] **The "handler" layer has no handlers.** `DetectorHandlerFactory`, `ReductionHandlerFactory`, `LogdataHandlerFactory` subclass `JobBasedPreprocessorFactoryBase` and produce `Accumulator`s; no `Handler` class exists anywhere in the backend. Related: `LatestValueHandler` (`handlers/accumulators.py`) is an accumulator, and `handlers/config_handler.py` contains `ConfigProcessor`. Suggest renaming the factories to `*PreprocessorFactory` and `LatestValueHandler` to `LatestValue`, and aligning module names; alternatively bless "handler" as the term and rename the base class. The glossary currently says "avoid handler for new names".
- [high] **`Job` stores its `Workflow` in `self._processor`** (`core/job.py`, param `processor: Workflow`). "Processor" already means the `Service`-driven protocol (`core/processor.py`), and `ess.reduce` has `StreamProcessor` as a third sense. Rename to `_workflow`.
- [med] **`JobManager.service_name = 'data_reduction'` is hardcoded** (`core/job_manager.py`) and wrong for `monitor_data`/`detector_data`/`timeseries` workers; the real name lives on `JobFactory` and `OrchestratingProcessor`. Looks like dead or misleading state — remove or wire through.
[...]

PR #1103, opened 2026-07-22
Replace stale stream-registry design doc with docstrings and ADR

SimonHeybrock:
## Summary

`docs/developer/design/stream-registry.md` documented types and methods that no longer exist -- `LogContextBinding`, `Instrument.log_context_bindings`, `Instrument.get_context_keys`, `TransformValueLog`, `StreamProcessorWorkflow.context_keys` -- superseded by `ContextBinding`, `Instrument.context_bindings`, `resolve_context_keys`, and `ValueLog` per ADR 0003, which also dropped `context_keys` from the workflow protocol. The project favors in-code documentation once a design is implemented, so rather than rewrite the doc, its still-accurate content is folded into the docstrings of the code it describes (`stream.py`, `nexus_helpers.py`), and the one genuine decision it captured -- generating `streams_parsed.py` from a NeXus geometry file at development time instead of parsing NeXus at runtime or hand-maintaining the registry -- becomes a retroactive ADR 0009.

Part of #1100

## Test plan

`tox -e docs` builds cleanly (only a pre-existing, network-caused intersphinx warning, unrelated to this change) and `python -m pytest -n auto` passes.


SimonHeybrock (commented):
LGTM

PR #1104, opened 2026-07-22
Rename and split backend handler layer into preprocessors and workflows

SimonHeybrock:
## Summary

The backend's "handler" layer never produced Handler objects, and it conflated two distinct concepts. This PR renames the layer to what it actually is, then splits the two concepts it was hiding.

**Rename.** The `*HandlerFactory` classes subclass `JobBasedPreprocessorFactoryBase` and produce `Accumulator`s; no `Handler` class exists anywhere in the backend. Per the glossary audit, "handler" is retired as backend terminology: `core/handler.py` becomes `core/preprocessor.py`, and the concrete factories become `DetectorPreprocessorFactory`, `ReductionPreprocessorFactory`, and `LogdataPreprocessorFactory`. `config_handler.py`'s `ConfigProcessor` was never a preprocessor -- it dispatches `Command` messages from the commands topic to `JobManagerAdapter` -- so it moves to `core/command_dispatcher.py` and is renamed `CommandDispatcher`.

**Split.** Renaming the directory alone would have swapped a vague label for a precise but wrong one: the layer held both the preprocessor factories (raw stream -> workflow input) and the scientific reduction workflows (detector views, monitor and wavelength-LUT workflows, and the `Workflow` protocol and factory themselves). The glossary treats Workflow and Preprocessor as separate concepts, so the packages now do too. `preprocessors/` keeps the factories, accumulators, and raw->input adapters; a new `workflows/` holds the reduction layer, its specs, and the workflow-only helpers `geometry_signal` and `dynamic_transforms`.

`timeseries.py` straddled both layers and is split: `TimeseriesStreamProcessor` moves to `workflows/`, `LogdataPreprocessorFactory` stays in `preprocessors/`. The standalone geometry CLIs (`download_geometry`, `upload_geometry`), which belong to neither layer and have no importers, move to `scripts/` alongside `make_geometry_nexus`.

One naming deviation: `LatestValueHandler` becomes `LatestValueAccumulator` rather than `LatestValue`, because `accumulators.py` already has an unrelated `LatestValue` class (a `streaming.Accumulator` for `StreamProcessor` workflows) that the literal target name would have collided with.

Part of #1100.

## Test plan

`python -m pytest -n auto` passes (4039 passed, 47 skipped, 8 xfailed).


SimonHeybrock:
LGTM

Issue #1055, opened 2026-07-06
Dashboard performance: profiled bottlenecks and ranked fix targets (tab switching, plot creation, idle CPU)


SimonHeybrock:
# Summary

We profiled the dashboard end-to-end (server Python via py-spy, browser via Chrome CPU profiles, long-task timing, websocket byte counts, and DOM mutation tracing) to find where the remaining time goes after our earlier upstream contributions (HoloViews 1.23.0, Panel 1.8.8/1.9.3). Setup: `--transport fake` with the committed dummy fixture (**4 plots** on the Detectors tab), holoviews 1.23.1 / panel 1.9.3 / bokeh 3.9.1, Chromium via Playwright.

Headline numbers with just 4 small plots and one session:

| Symptom | Measurement |
|---|---|
| Idle on plot tab, **client** | ~30% of the browser main thread busy; one ~225 ms blocking task *per second*; ~840 `<style>` rewrites/s |
| Idle on plot tab, **server** | ~11% of a CPU core per session; ~90% of that is provably redundant work |
| Switch **into** plot tab | 1.7–1.9 s of blocked browser main thread (4 repeats, consistent); ~250–700 KB websocket payload |
| Switch to widget-only tab | ~0–60 ms |

Almost all of this concentrates in **four upstream mechanisms** (Bokeh ×3, Panel ×1), each fixable in one place, plus two cheap app-side mitigations. Details + repro below so each item can be split into its own fix/PR.

> **Update 2026-07-06:** follow-up analysis of the 100 ms session tick, its relation to the in-flight delivery restructuring (#1043/#1045/#1046/ADR 0007), and a verified event-driven alternative is in [this #1046 comment](https://github.com/scipp/esslivedata/issues/1046#issuecomment-4897619291). It changes the recommended fix for Target 3's app side — see the update notes inline below.

# How to think about server vs. client (for non-experts)

Every plot on screen is a tree of **Bokeh models** — Python objects (axes, ranges, glyphs, toolbars, layout boxes; ~50–100 per plot) mirrored 1:1 in the browser. The pipeline is:

1. **Server (Python)**: build or patch the model tree.
2. **Serialize**: encode changes as JSON/binary patches, send over the websocket.
3. **Client (JS)**: decode patches, update the mirrored models.
4. **Client (JS)**: bokehjs builds *views* for models, computes **layout** (who gets which pixels) and resolves **styles** (colors, fonts — via the browser's CSS engine).
[...]

PR #1105, opened 2026-07-22
Dashboard: event-driven session flush (wake-on-commit) retires the 100 ms poll

### LINKED ISSUE #1096: Dashboard: event-driven session flush (wake-on-commit) to retire the 100 ms poll
Extracted from the design comment on #1046 (2026-07-06), updated for what has landed since. Supersedes the app-side "skip hold/freeze on no-op ticks" item from #1055 Target 3 — see "Relation to #1055" below.

## Problem

The 100 ms session poll is only a *detection* clock for `FrameClock` generation advances (ADR 0005) — plots already flush at data cadence (~1 Hz). What the 10 Hz tick costs: mean-50/max-100 ms added display lag, one unconditional `hold`+`freeze` model-graph recompute per tick per session even when idle (measured in #1055 Target 3: ~11% of a core per idle session, ~90% redundant), and the O(grids×cells×layers) lock-taking scan flagged in #1044's scalability notes. The goal is "flush right after data arrives, then nothing until the next frame" — an event-driven trigger. ADR 0007 (accepted) sanctions this in principle: *"Callbacks, where they survive, carry no data — at most 'something changed' wake-ups that degrade to the next poll tick if lost."* This issue is the concrete instance of that clause.

## Mechanism — verified working

ADR 0005's alternatives section rejected `doc.add_next_tick_callback` from the ingestion thread, but the failure that rejection encodes was **data-carrying push / direct cross-thread session mutation** (and `pn.state.execute`, which genuinely cannot resolve session context from a background thread). A **wake-up-only** callback preserves ADR 0005's actual invariant: all session-bound mutation still happens on the session's IOLoop, in document context — only the *scheduling* crosses threads, which is Bokeh's documented pattern for updating documents from threads.

Verified empirically on panel 1.9.3 / bokeh 3.9.1 with a minimal two-session app: a background thread bumps a generation at 1 Hz and schedules a no-arg tick via `doc.add_next_tick_callback`. Inside the woken tick, `pn.state.curdoc` resolves to the correct session's document, `pn.io.hold()` + `doc.models.freeze()` behave normally, and both sessions update at 1 Hz with **zero** session work between frames. Scheduling into a destroyed document raises a catchable exception, so lazy unregistration works.

## Design sketch

- A small `WakeupHub` (or an extension of `SessionRegistry`, which already tracks sessions): sessions register `(doc, tick)` in session context at startup; unregister on `on_session_destroyed` **and** on the stale-session reaper path — both paths now exist and are exercised by the two-tier teardown from #955 — plus lazily when scheduling raises.
- Ingestion thread, after `flush_frames` commits a grid: for each registered session with no wake pending, set a pending flag and `doc.add_next_tick_callback(tick)`. Waking every session on any grid's commit is fine — the tick is generation-gated per *active* grid, so wrong-tab wakes are ~free no-ops, and the pending flag coalesces bursts.
- Tick body = today's `periodic_update` pass, unchanged: idempotent and version/generation-gated, so lost or duplicate wakes are harmless — exactly ADR 0007's degradation property, with the fallback poll as the safety net.
- The periodic callback survives at ~1–2 s instead of 100 ms, for what genuinely needs a clock: browser-heartbeat check, notification poll, freshness-pill stall aging, catch-up. Post-click widget feedback does not regress to that cadence: UI handlers already run in-session and can trigger an immediate tick. Later (optional), shared-state version bumps (e.g. the topology version from #1040) could poke the same hub for snappy cross-session widget refresh.

## Effect

- Commit→browser latency: ~1 ms of scheduling instead of mean 50 / max 100 ms. The remaining floor is the ingestion loop's 50 ms drain cadence — irrelevant at 1 Hz publish cadence; could later become a blocking consumer poll if it ever matters.
- Per-session background work drops 10×: the "sessions × layers × 10 Hz lock acquisitions" wall from #1044 becomes 1 Hz, and the per-tick freeze-pop recompute runs once per data frame instead of ten times per second. Bears directly on #753 (lag accumulation) and #718 (performance follow-ups).

## Prerequisites — all met

Pull-on-frame made `flush_frames` the single authoritative "grid has a new displayable frame" point — the natural wake site (landed; #1062/#1065). Session unregistration rides the clean and reaper teardown paths from #955 (landed). #1040 (grid lifecycle to version polling, in flight) is independent but its topology version is a candidate second wake source.

## ADR changes required

- Scoped amendment to ADR 0005: narrow the alternatives-rejection to data-carrying push; the flush *trigger* moves from clock to event while the flush itself stays on the session thread.
- A sentence in ADR 0007 naming the frame-flush wake-up as the sanctioned form of its no-data-callback clause.

## Relation to #1055

#1055 Target 3's app-side recommendation was "skip `hold`/`freeze` on ticks that end up doing nothing" — a guard inside the 10 Hz tick that avoids the per-tick model-graph recompute when no generation advanced. This issue supersedes that as the end state: with wake-on-commit there *are* no idle ticks to skip (the residual 1–2 s housekeeping tick can keep an unconditional, now-cheap pass). The skip-guard remains a trivial interim mitigation if this lands late. The upstream Bokeh items in #1055 (CSS custom-property reads, style invalidation, tab-switch rebuild cost) are unaffected by cadence and stay worthwhile — each real wake still pays one recompute.

---

SimonHeybrock:
Fixes #1096.

## What

Sessions no longer poll shared state at 10 Hz. A `WakeupHub` lets the ingestion loop (after each drain pass) and UI-driven version bumps (workflow staging/commit/stop, gate toggle, plot topology, notification push) schedule data-free wake ticks onto each session's IOLoop, per ADR 0005's 2026-07-06 amendment and ADR 0007's no-data-callback clause. The tick body stays the existing version-gated pull pass, so lost or duplicate wakes are harmless.

Ticks are gated by cheap per-handler `has_work` predicates. Wake ticks run only handlers whose predicate fires and skip the hold+freeze batch entirely when none does, so a wake meant for another session's tab costs no model-graph recompute — this keeps the #1055 Target 3 skip-guard as a companion rather than superseding it (see the [analysis on the issue](https://github.com/scipp/esslivedata/issues/1096#issuecomment-5042816352): without it, multi-source multi-grid deployments could exceed today's 10 recomputes/s). The periodic callback survives at 1 s as a housekeeping tick, predicate-gated like a wake tick, except that every 5 s one runs every handler unconditionally: that full pass is the clock for wall-clock-driven displays no version counter can signal (status staleness at its 60 s threshold, heartbeat readouts) and bounds how long a predicate hole can keep a widget stale. Sub-5 s time-based work (freshness-pill stall aging at 2 s) is encoded as a time term in its handler's predicate; a stalled stream sends no wakes, while healthy data resets the term each frame. Events that change *what is visible* rather than what changed — a tab switch, closing a config modal — request an immediate full tick, since no predicate can observe "became visible".

The hub's per-session pending flag coalesces wake bursts and is cleared before the tick body runs, so a commit landing mid-tick schedules a fresh wake instead of waiting for the poll. Sessions unregister on both teardown tiers; scheduling into a destroyed document drops the registration lazily. Both load-deferred pieces of setup — hub registration and the housekeeping callback itself — wait for the browser session to load, because either one running before the client's first sync can leave the session rendering permanently empty.

## Why

The 100 ms poll was only a detection clock, but cost mean-50/max-100 ms display lag plus one unconditional hold+freeze model-graph recompute per tick per session even when idle (~90% redundant, measured in #1055), and the O(grids×cells×layers) scan flagged in #1044.

## Measured impact

Dummy instrument, fake transport, auto-started workflows publishing at 1 Hz; sessions idle on the named tab. Measured with a profiling harness patching `FrameClock.commit` (commit time), `PlotOrchestrator.frame_generation` (first observation per session → latency), and `DocumentModelManager.recompute` (freeze-exit recompute count); process CPU includes the in-process fake backend.

| Scenario | Metric | main | this PR |
|---|---|---|---|
| 1 session, Detectors | recomputes | 9.9/s | 1.2/s |
| | commit→observation latency (mean/max) | 42/73 ms | 0.8/2.5 ms |
| | process CPU | ~44% | ~20% |
| 3 sessions, all Detectors | recomputes | 28/s | 3.6/s |
| | latency (mean/max) | 62/184 ms | 46/127 ms |
| | process CPU | ~81% | ~31% |
| 1 Detectors + 2 Workflows | recomputes | — | 1.6/s |
| | latency (mean/max) | — | ~1/9 ms |

Recompute rates decompose exactly as designed: one per data wake per session viewing the grid, plus 0.2/s per session for the 5 s full pass — e.g. 3.6/s = 3 wakes + 3×0.2, and 1.6/s = 1 wake + 3×0.2 (wrong-tab wakes are free). Notes: with three sessions all rendering the same 1 Hz grid, latency in both builds is dominated by the shared IOLoop serializing the three per-frame renders, not by detection — the wake mechanism removes the detection delay (visible in the 1-session and mixed-tab rows) and cannot remove render serialization.

These numbers predate the review round below. The Detectors rows are unaffected by it, but sessions parked on **Workflows** now refresh at the observed job-status rate instead of the 5 s full pass — that is the point of the start-feedback fix, and still far below the old unconditional 10/s. The mixed-tab row would rise accordingly if re-measured.

## Review round

A final review pass found the tick gates too narrow in three places and the hub's coalescing flag able to latch permanently. All are fixed here, each with a regression test.

- **Stopping a workflow left its plots looking live for up to 5 s.** A stop bumps no topology version and, unlike a restart, produces no further frame, so neither existing gate term could see it. `PlotDataService` now exposes an aggregate version over the per-layer counters, and the plot gate tracks it. The counter advances only for transitions that took effect, so a no-op `data_arrived` on an already-READY layer cannot arm the gate on every data message.
- **Start feedback regressed from ~1 s to ~6 s.** A committed job reaching RUNNING bumps no workflow state version, so the row sat on its post-commit rendering until the next full pass. `JobService` gained a version counter and `WorkflowStatusWidget.is_stale` a job-status term alongside the structural one.
- **Returning to a hidden tab showed content frozen since it was last visible.** The visibility-gated predicates correctly report "no work" while hidden, and becoming visible again is not a change any predicate can observe. Tab switches and modal close now request a full tick rather than a gated one.
- **A lost wake callback could silence a session permanently.** The pending flag was cleared only by the tick that consumed it. Bokeh routes `add_next_tick_callback` through the document's held-event list while a hold is active — which is every tick — and `unhold` copies and rebinds that list non-atomically, so a foreign-thread append in between is dropped. The housekeeping tick now re-arms the flag, bounding the damage to one period. A superseded hub entry also no longer runs its tick after re-registration.
- **`unregister_custom_handler` compared handlers by identity**, which never matches separately obtained bound methods.

The two documented predicate holes are consequently closed; what remains on the full pass is genuinely wall-clock-driven text with no message behind it.

## Test plan

- Unit tests for `WakeupHub` (burst coalescing, pending-flag clear-before-run, self-heal re-arm, superseded entries, lazy drop on destroyed documents, unregistration) and for the wake/gated-housekeeping/full-pass tick tiers in `SessionUpdater` (deterministic via injected clock).
- `PlotGridTabs` widget tests run through the wake gate, so any change source the poll pass renders but the predicate misses fails loudly instead of silently lagging; `TestWakeGateContract` pins the quiescence direction and the per-source re-arm, including job stop and plotter swap.
- Multi-session Playwright smoke test and the `-m browser` UI suite pass.
- Manual UI checks from the original round:
  - [x] Plots update promptly and in sync while data flows; no staleness after leaving a tab open for minutes
  - [x] Start/stop workflow: row feedback is immediate in the acting session and prompt in a second session
  - [x] Tab switch renders the newly shown grid's latest data without noticeable delay
  - [x] Toasts appear promptly in all sessions
  - [x] Freshness pill ages a stalled stream at its usual 2 s cadence; status badge staleness (backend stopped) appears within ~5 s
  - [x] Job restart resets the affected plot within ~5 s (or on its first new frame)
- Manual re-check for the review round:
  - [x] Stopping a workflow resets/freezes its plots promptly, not after a ~5 s pause
  - [x] After starting a workflow the row reaches RUNNING promptly, without sitting on the post-commit state for several seconds
  - [x] Switching to **Workflows** after time away shows current badges and timing immediately, not the values from when the tab was last visible
  - [x] Same for **System Status** (session list, backend status)
  - [x] Closing a plot config modal renders the grid's latest data without waiting
  - [x] A long session (tens of minutes, several tabs, some idle) stays responsive — no tab that gradually stops updating promptly


SimonHeybrock (commented):
Could not identify any further problems in UI testing. Code grew more than I would have liked, but cannot think of a way to simplify.

src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
```diff
@@ -256,8 +262,16 @@ def __init__(
...
         self._last_topology_version = self._orchestrator.topology_version()
 
-        # Register handler for periodic polling
-        session_updater.register_custom_handler(self._poll_for_plot_updates)
+        # Register the poll pass; the predicate lets wake ticks skip it (and
+        # the batched hold+freeze) when nothing visible changed.
+        session_updater.register_custom_handler(
+            self._poll_for_plot_updates, has_work=self._has_pending_work
+        )
+        # A tab switch changes what the poll pass must flush but bumps no
```
nvaytet, line 270: For the tabs that are not displayed: are they updating every 5 seconds in the background or do they do nothing until they are displayed again?

src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
```diff
@@ -256,8 +262,16 @@ def __init__(
...
+        # shared version, so request an immediate in-session tick rather than
+        # waiting for the housekeeping cadence.
+        self._session_updater = session_updater
+        self._tabs.param.watch(self._on_active_tab_changed, 'active')
```
nvaytet, line 274: Once the tab switch triggers update via the wake tick, how noticeable is the delay after you have switched tab? Does it take a second or so for everything to update?

I remember a project a saw a while ago (can't find it now) that dealt with such a delay by updating slightly before the user clicked, using the event when the mouse hovered over the tab instead of the click event.
Could be something to think about if we need to improve the UX? That said, I am not sure it's a great idea as you may be triggering a lot of unecessary work by just hovering back and forth over the tabs but not clicking anything...

src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
```diff
@@ -635,10 +690,16 @@ def _poll_for_plot_updates(self) -> None:
...
-        intervals is acceptable for config UI updates.
+        Bokeh models. Skipped layers keep their dirty flag set; a tab switch
+        requests its own tick (:meth:`_on_active_tab_changed`), which sends the
+        newly visible layers' latest cached state.
         """
+        # Snapshot before reading any layer state, and record it only once the
+        # pass completes (below): a concurrent lifecycle change is then picked
+        # up by the next pass, and an exception escaping mid-pass leaves the
+        # gate armed so the next tick retries.
+        layer_version = self._plot_data_service.version
```
nvaytet, line 712: Why is it defined here and saved after the pass completes? Is it because the `self._plot_data_service.version` can change between here and where it is saved below? If not, can we just move this down to L890?

src/ess/livedata/dashboard/widgets/workflow_status_widget.py
```diff
@@ -1146,6 +1147,22 @@ def _on_commit_click(self) -> None:
             lambda: self._orchestrator.commit_workflow(self._workflow_id),
         )
 
+    def is_stale(self) -> bool:
+        """True when a refresh would show something the browser has not seen.
```
nvaytet, line 1160: Not sure I get the terminology: I thought 'stale' would mean something that has not changed in a long time. But did I inderstand correctly that here it means something has changed but it has not been displayed yet?

src/ess/livedata/dashboard/widgets/workflow_status_widget.py
```diff
@@ -1351,6 +1370,7 @@ def _sync_gate_checkbox(self) -> None:
         :meth:`JobOrchestrator.set_gate_enabled` is a no-op for equal values, so
         there is no feedback loop.
         """
+        self._last_gate_version = self._orchestrator.get_gate_version()
```
nvaytet, line 1386: For my understanding, are these versions integers or are they something like datetimes?

src/ess/livedata/dashboard/job_service.py
```diff
@@ -51,6 +63,7 @@ def status_updated(self, job_status: JobStatus) -> None:
         logger.debug("Job status updated: %s", job_status)
         self._job_statuses[job_status.job_id] = job_status
         self._job_status_timestamps[job_status.job_id] = time.time_ns()
+        self._version += 1
```
nvaytet, line 66: Is the version an integer that grows indefinitely? I guess it never overflows in python...

src/ess/livedata/dashboard/plot_data_service.py
```diff
@@ -318,10 +355,7 @@ def data_arrived(self, layer_id: LayerId) -> None:
         layer_id:
             Layer ID to update.
         """
-        with self._lock:
-            state = self._layers.get(layer_id)
-            if state is not None:
-                state.data_arrived()
+        self._apply(layer_id, lambda state: state.data_arrived())
```
nvaytet:
```suggestion
        self._apply(layer_id, lambda state: state.data_arrived(), create=False)
```
to make sure we don't depend on the default arg value of `self._apply`? (same below?)

src/ess/livedata/dashboard/wakeup_hub.py
```diff
@@ -0,0 +1,126 @@
...
+
+    :meth:`wake_all` may be called from any thread — the ingestion loop after
+    a drain pass, or a session's UI thread after a shared version bump. A wake
+    carries no data: it schedules the session's tick onto that session's
+    IOLoop via ``Document.add_next_tick_callback`` (Bokeh's documented pattern
+    for cross-thread scheduling), and the tick re-reads shared state in
+    session context. Ticks are idempotent and version-gated, so lost or
+    duplicate wakes are harmless; the housekeeping poll is the safety net.
+
+    A per-session pending flag coalesces bursts: while a wake is scheduled but
```
nvaytet, line 42: Did you find that combining many quick wake requests into one actual run was necessary? (if so in  which cases?)
Or are we adding extra complexity which was not so needed?

Is it related to multi-sessions where multiple users could be doing things at the same time?

src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
SimonHeybrock: Nothing until they are displayed again -- and it is stronger than "no repaint": a hidden grid's layers are *deactivated* (`activate_layer(..., False)`), so unless another session is viewing the same layer they do not even compute, and with `dynamic=True` there are no Bokeh models to update anyway. Their buffers keep filling, so nothing is lost. The 5 s full pass only keeps their *structure* reconciled (a cell someone else added or removed).

On the way back, the 0->1 activation rebuilds the layer synchronously and that same tick pushes it, so you see current data, not a 5 s-old rendering.

Fair thing to ask of the code, though -- the docstring only described what happens to the *visible* tab. Written down now.

src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
SimonHeybrock: Yes, exactly that. The ingestion thread bumps `PlotDataService.version` while the pass runs, so reading it at L890 would also absorb transitions that landed *during* the pass and were never rendered -- the gate would then be closed on them until the next unconditional full pass, i.e. up to 5 s late. Recording it late has a second benefit: an exception escaping mid-pass leaves the gate armed, so the next tick retries.

The comment said this far too tersely; it now names the concurrent writer.

src/ess/livedata/dashboard/widgets/workflow_status_widget.py
SimonHeybrock: Good catch, and it is worse than an unclear name: `stale` already means your reading in this same code -- `JobService.is_status_stale` is "no heartbeat within the timeout". Two meanings, one word, in one PR.

Renamed to `has_pending_work` (also matching the `has_work` tick predicates it feeds), and `stale` now has a glossary entry reserving it for age. The two are genuinely independent: staleness sets in by the clock with no version bump behind it, which is precisely why it is rendered by the periodic full pass rather than by a wake.

src/ess/livedata/dashboard/widgets/workflow_status_widget.py
SimonHeybrock: Plain `int`s. Incremented by the single writer, and only ever compared for *equality* against a session's last-seen value -- never ordered, subtracted, or read as a time. A session only asks "did this move since I rendered it?".

Since this came up in two places, the dashboard glossary now has a *Version counter* entry saying so.

src/ess/livedata/dashboard/job_service.py
SimonHeybrock: It only grows, and Python ints are arbitrary-precision, so there is no wrap-around to design for. Nothing depends on the magnitude anyway -- only on "same or different" -- so even a pathological rate would be harmless.

src/ess/livedata/dashboard/plot_data_service.py
SimonHeybrock: Taken, and pushed one step further: `create` is now a required keyword-only argument, so no call site inherits a default. There is no sensible default to have -- whether an unknown layer id may be created depends on whether that caller can legitimately arrive first (`job_started` and `error_occurred` can, `data_arrived` and `job_stopped` cannot).

src/ess/livedata/dashboard/wakeup_hub.py
SimonHeybrock: Not a multi-session thing -- wakes fan in per session with a single browser open, from two directions. The ingestion loop wakes after every drain pass that consumed anything and only sleeps when a pass was *empty*, so while catching up on a backlog it wakes far faster than the old 10 Hz. And a single UI action can bump several versions in one call stack (a staging transaction's commit, a grid template's per-cell topology bumps). Each queued tick that finds work pays a full `hold`+`freeze` pass, so a burst without the flag buys back exactly the redundant recompute this PR removes.

Honest caveat: I never isolated a measurement of the flag itself. In the steady 1 Hz scenario in the description there is roughly one wake per frame anyway, so it saves nothing *there*; the burst cases above are why it is in. It costs a bool and a dict lookup, so I kept it. The subtle part of that paragraph is not the coalescing but the `clear_pending` re-arm, and that one is there for correctness (a lost `add_next_tick_callback`), not for cost.

The docstring now gives this reason rather than just describing the mechanism.

src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
SimonHeybrock: Good question, and I had not measured it, so I did: Playwright driving the dummy fixture's Detectors grid (2 images, fake backend at 1 Hz), timing from the browser's own `click` on the tab to the first animation frame where both canvases actually contain colormapped pixels.

| | mean | range |
|---|---|---|
| repeat switch (Workflows -> Detectors) | **419 ms** | 305-634 |
| first switch after a page load | **866 ms** | 794-936 |

Split of the warm case: 186 ms click -> the `active` watcher firing server-side, 52 ms for the `full=True` tick itself, then the browser paints at 419 ms (~116 kB of patch in between). The 186 ms is Bokeh's property-change delivery, not this branch -- it measures 119-182 ms for the plot-free tabs too.

So the wake tick is ~50 ms of a ~420 ms switch and finishes ~175 ms *before* the paint, which is the part I care about: fresh data lands in the **first** painted frame, so there is no "grid appears, then fills in" two-stage flash. Caveat in the other direction: this is localhost, two 64x64 images, headless Chromium -- a real instrument's grid will be slower in both the tick and the transfer.

On the hover idea: I like it, but I think it would buy the wrong 12% here. Detection is already ahead of the paint, so the only thing hovering could pre-pay is layer activation (inside that 50 ms). The dominant cost is Bokeh model creation plus patch delivery, and that cannot be moved earlier without giving up `dynamic=True` -- which is exactly what makes a hidden tab cost nothing (#1055 Target 3, #1044). Your own worry applies too: activation is refcounted, so flicking across tabs would start and stop computes for every grid on the way. If tab switches turn out to feel slow on a real instrument, I would go after the tick and render cost rather than the trigger.

One honest oddity while measuring: the very first switch in a *freshly started server process* took 4.8 s, of which 3.8 s was one server tick -- first-time HoloViews/Bokeh plot construction in that process. It never recurred once the process was warm, and it is not specific to tab switching (the same cost appears on the first plot a process ever builds), but it is a real first-user-of-the-day effect worth knowing about.

SimonHeybrock: > the very first switch in a freshly started server process took 4.8 s, of which 3.8 s was one server tick

Note: I do not think it is that long with a real browser, in my experience.

nvaytet (approved):
One last thing: while the [gist](https://gisthost.github.io/?60a6d20f01b1697a2da762bd7c3f5046) was useful, I found the language in it very hard to understand. It was so filled with jargon specific to various parts of software development (many of which I am not familiar with), that I actually had to ask Claude to translate it into 'plain english' for me 😄 

I generally find it is sometimes also an issue in comments in the code written by claude, although not quite to the same extent. The language may be concise and probably accurate, but often difficult to follow if you are not an expert.

SimonHeybrock:
> One last thing: while the [gist](https://gisthost.github.io/?60a6d20f01b1697a2da762bd7c3f5046) was useful, I found the language in it very hard to understand. It was so filled with jargon specific to various parts of software development (many of which I am not familiar with), that I actually had to ask Claude to translate it into 'plain english' for me 😄
> 
> I generally find it is sometimes also an issue in comments in the code written by claude, although not quite to the same extent. The language may be concise and probably accurate, but often difficult to follow if you are not an expert.

I suppose we would need to tell it about the readers (our team members) so it can better gauge how to calibrate the language. Or at least tell it that reviewers may not be deeply familiar with the project.

PR #1107, opened 2026-07-23
Fix saturated 1-10 colorbar for uniform positive log-scale image data

### LINKED ISSUE #1078: Log-scale color limits force a saturated 1-10 colorbar for uniform positive image data
**Severity:** Low

`Plotter._get_log_scale_clim` (`src/ess/livedata/dashboard/plots.py:628-651`) provides fallback color limits when HoloViews' `LogColorMapper` would otherwise fail on all-`NaN` data:

```python
vmin = float(data.data.nanmin().value)
vmax = float(data.data.nanmax().value)
if np.isnan(vmin) or np.isnan(vmax) or vmax <= vmin:
    return (1.0, 10.0)
return None
```

The `vmax <= vmin` condition (intended to guard the all-`NaN` case, per the docstring) also matches the case `vmax == vmin`, which occurs whenever the (non-NaN) data is uniform — e.g. a detector image that is entirely `1e6` counts, or any constant positive array. In that case the function returns the placeholder `(1.0, 10.0)` instead of `None` (which would let HoloViews pick sensible bounds around the actual constant value). The plot then renders fully saturated with a colorbar reading `1-10`, which is misleading for data that is actually a constant value far outside that range.

**Fix hint:** only trigger the placeholder fallback when `vmin`/`vmax` are actually `NaN` (the genuine failure case `LogColorMapper` can't handle); for the `vmax == vmin` uniform-data case, return sensible bounds around the actual value (e.g. `(vmin * 0.9, vmin * 1.1)` or similar) instead of the unrelated placeholder.

**Test hint:** call `_get_log_scale_clim` with a `DataArray` whose values are all equal to some positive constant (e.g. all `1e6`), and assert the returned bounds are related to that value, not the fixed `(1.0, 10.0)` placeholder.

Extracted from #1039.

SimonHeybrock:
## Summary

`Plotter._get_log_scale_clim` supplies fallback color limits for log-scale images that HoloViews' `LogColorMapper` cannot bound itself. Its guard collapsed two unrelated cases: genuinely unplottable all-NaN data, and uniform positive data (`vmax == vmin`). Both returned the placeholder `(1.0, 10.0)`, so a constant-valued detector image (e.g. all `1e6` counts) rendered fully saturated with a colorbar reading `1-10` — well outside the actual data range.

## Why

The original guard `vmax <= vmin` was meant to catch all-NaN data. Two things were off. First, scipp's `nanmin`/`nanmax` return `+inf`/`-inf` (not NaN) for all-NaN input, so the all-NaN case actually surfaces as `vmax < vmin`; the `==` branch was only ever hit by uniform data, which is plottable. Second, the saturation was *our* forced placeholder, not HoloViews: left to itself HoloViews handles uniform data fine.

We still special-case uniform data, but for a subtler reason: HoloViews' own fallback for a zero-width range is *additive* (`value ± 1`), which produces a non-positive lower bound for small constants (e.g. `0.5 → -0.5`, `1.0 → 0.0`) — invalid on a log scale. A *multiplicative* bracket (`vmin * 0.9`, `vmax * 1.1`) stays positive and gives a readable colorbar. Log-scale masking has already replaced non-positive values with NaN, so a non-NaN constant is guaranteed positive (a constant negative or zero image becomes all-NaN and correctly falls through to the placeholder).

Fixes #1078.

## Test plan

- [x] New unit tests for `_get_log_scale_clim`: all-NaN placeholder; uniform data (with and without NaN) brackets the constant; small uniform constant keeps positive bounds; varying data defers to HoloViews.
- [x] `pytest tests/dashboard/plots_test.py -k "GetLogScaleClim or ImagePlotter"` passes.

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -633,11 +633,12 @@ class Plotter:
         error in _draw_colorbar. Scipp's nanmin/nanmax return +inf/-inf rather
         than NaN for all-NaN input, so the all-NaN case surfaces as vmax < vmin.
 
-        A second degenerate case is uniform data (``vmax == vmin``): HoloViews
-        renders a fully saturated plot with a zero-width color range. Bounds
-        bracketing the constant value give a readable colorbar instead. Log-scale
-        masking has already replaced non-positive values with NaN, so a non-NaN
-        constant is guaranteed positive.
+        Uniform data (``vmax == vmin``) is a second degenerate case. HoloViews
+        does handle it, but via an *additive* fallback (value +/- 1) that yields
+        a non-positive lower bound for small constants (e.g. 0.5 -> -0.5), which
+        is invalid on a log scale. A multiplicative bracket around the constant
+        stays positive. Log-scale masking has already replaced non-positive
+        values with NaN, so a non-NaN constant is guaranteed positive.
 
         Parameters
         ----------
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -625,13 +625,19 @@ class Plotter:
     @staticmethod
     def _get_log_scale_clim(data: sc.DataArray) -> tuple[float, float] | None:
         """
-        Return fallback clim for log scale if data is all NaN.
+        Return fallback clim for log scale when HoloViews cannot pick bounds.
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1108, opened 2026-07-23
Count variances in TemporalBuffer memory-limit sizing

### LINKED ISSUE #1081: TemporalBuffer memory-limit sizing ignores variances, allowing up to 2x the configured budget
**Severity:** Low

`TemporalBuffer._initialize_buffers` computes how many samples fit within the configured memory budget using only the values array's byte size (`src/ess/livedata/dashboard/temporal_buffers.py:400-408`):

```python
bytes_per_element = data.data.values.nbytes
for name in accumulated:
    bytes_per_element += data.coords[name].values.nbytes
if 'time' in data.dims:
    bytes_per_element /= data.sizes['time']

max_capacity = max(1, int(self._max_memory / bytes_per_element))
```

`data.data.values.nbytes` only accounts for the `.values` array. When the underlying scipp `Variable` also carries `.variances` (common for reduced/accumulated detector or monitor data), that second same-sized array is not counted at all. The same omission applies to any accumulated coordinate that itself carries variances. As a result, a buffer holding data-with-variances can use up to roughly twice the memory implied by `self._max_memory`, since `max_capacity` is computed as if only the values array existed.

**Impact:** for instruments/workflows where the plotted data carries variances, per-key buffer memory can silently run ~2x over the configured cap — relevant for the dashboard's per-key memory bounding strategy on multi-day, memory-constrained deployments.

**Fix hint:** include `data.data.variances.nbytes` (when not `None`) and the equivalent for each accumulated coordinate in the `bytes_per_element` calculation.

**Test hint:** initialize a `TemporalBuffer` with `data` that has non-`None` `.variances`, with a `max_memory` budget; assert the resulting `max_capacity` is roughly half of what it would be if variances were ignored (i.e. that the variances array is actually counted).

Extracted from #1039.

SimonHeybrock:
## Summary
`TemporalBuffer._initialize_buffers` sized the per-element memory budget from `.values.nbytes` alone, ignoring `.variances`. Data carrying variances (common for reduced/accumulated detector or monitor data) could therefore use up to ~2x the configured `max_memory` budget, undermining the dashboard's per-key memory bounding strategy on memory-constrained deployments.

- Adds a `_variable_nbytes` helper that includes `.variances.nbytes` when present.
- Applies it to both the data buffer and each accumulated coord buffer.
- Adds a regression test asserting that variances roughly halve the resulting `max_capacity` for a fixed memory budget.

Fixes #1081

## Test plan
- [x] `python -m pytest tests/dashboard/temporal_buffers_test.py` passes, new test fails without the fix
- [x] Full `python -m pytest` suite passes

### DIFF
--- a/src/ess/livedata/dashboard/temporal_buffers.py
+++ b/src/ess/livedata/dashboard/temporal_buffers.py
@@ -13,6 +13,14 @@ import scipp as sc
 T = TypeVar('T')
 
 
+def _variable_nbytes(var: sc.Variable) -> int:
+    """Return the byte size of a variable's values, plus variances if present."""
+    nbytes = var.values.nbytes
+    if var.variances is not None:
+        nbytes += var.variances.nbytes
+    return nbytes
+
+
 class BufferProtocol(ABC, Generic[T]):
     """Common interface for all buffer types."""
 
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1109, opened 2026-07-23
Suppress plot-grid persistence during startup replay and shutdown

### LINKED ISSUE #1073: An exception during startup plot-grid replay permanently truncates the saved layout
**Severity:** Medium

At startup, `PlotOrchestrator._load_from_store()` (`src/ess/livedata/dashboard/plot_orchestrator.py:1604-1633`) replays each persisted grid through the normal mutating API:

```python
for spec in specs:
    grid_id = self.add_grid(spec.title, spec.nrows, spec.ncols)
    for cell in spec.cells:
        cell_id = self.add_cell(grid_id, cell.geometry, user_title=cell.user_title)
        for layer in cell.layers:
            self.add_layer(cell_id, layer.config)
    ...
```

Every one of `add_grid`, `add_cell`-adjacent mutations, and `add_layer` calls `self._persist_to_store()` internally as part of normal operation (e.g. `add_grid` persists at line 478; `add_layer`'s underlying `_setup_layer` persists at lines 1219/1261). `_persist_to_store()` (lines 1635-1645) unconditionally overwrites the `plot_grids` store entry with whatever is currently in memory:

```python
serialized = self._serialize_grids()
self._config_store['plot_grids'] = {'grids': serialized}
```

The whole replay loop is wrapped in one `try/except Exception` (lines 1609-1633) that only logs and swallows the error. If any `add_layer` call raises partway through replay (e.g. a params-schema mismatch from a stale config, or a validation error), the loop aborts — but the grids/cells/layers already replayed before the exception have already overwritten the store via their own `_persist_to_store()` calls. The result: the store now permanently contains only the partial layout that was applied before the crash; the user's remaining tabs are gone even on a clean restart, because there is nothing left in the store to recover them from.

**Related latent hazard, same root cause:** `shutdown()` (lines 1832-1845) calls `remove_grid()` once per grid, and `remove_grid` also persists after each individual removal (line 503). If `shutdown()` is interrupted partway (or even completes normally in an unexpected ordering), the store can transiently or permanently end up as `{'grids': []}`.

**Fix hint:** defer persistence during startup replay (e.g. a flag or context manager that no-ops `_persist_to_store()` while `_load_from_store()` is running), and persist once, only after the full replay completes successfully. A similar deferral should be considered for `shutdown()`'s per-grid removal loop.

**Test hint:** seed a config store with two valid grids and a third grid whose layer config is invalid enough to make `add_layer` raise; run `_load_from_store()`; assert the store still contains all three original grids' definitions afterward, not just the first two.

Extracted from #1039.

SimonHeybrock:
Fixes #1073.

## Problem

Every `PlotOrchestrator` mutator persists the *whole* in-memory model to the config store on each step (`plot_grids` is a single aggregate blob, rewritten in full). There is no way to treat a multi-step operation as one commit, so any composite lifecycle operation that drives those mutators leaks intermediate state to the store:

- **Startup replay** (`_load_from_store`) reapplies grids/cells/layers through the normal API. If an `add_layer` raises partway through — e.g. a stale layer config that parses but is rejected on add — the loop aborts *after* earlier steps have already overwritten the store. The saved layout is permanently truncated to the partial state, so the user's remaining tabs are gone even on a clean restart.
- **`shutdown()`** removes each grid in turn, every removal persisting, ending at `{'grids': []}` and wiping the saved layout. (Latent today — only tests call it — but the same defect.)

## Fix

A single `_suppress_persist()` context manager makes `_persist_to_store()` a no-op within the block; startup replay and shutdown both run under it. Persistence is suppressed, not deferred — nothing is written on exit: the model was just built from the store (replay) or is being torn down (shutdown), so the last committed layout is exactly what should survive, and a raising block leaves the store untouched, preserving the full layout for the next attempt.

## Similar problems

The sibling `JobOrchestrator` follows the same load-then-persist shape but is structurally immune: it persists one store key *per workflow* rather than a single aggregate blob, so a partial load can never clobber the other entries. The cluster is confined to `PlotOrchestrator`'s two aggregate-blob operations, both fixed here.

## Tests

Two regression tests in `TestPersistenceResilience`, each verified to fail without the fix:

- a store seeded with two valid grids straddling a broken one (a cell stacking two non-overlayable layers) — after the failed replay all three grid definitions still survive in the store, including the one never reached;
- `shutdown()` leaves the saved layout intact.

### DIFF
--- a/src/ess/livedata/dashboard/plot_orchestrator.py
+++ b/src/ess/livedata/dashboard/plot_orchestrator.py
@@ -16,7 +16,8 @@ from __future__ import annotations
 import copy
 import threading
 import traceback
-from collections.abc import Mapping, Sequence
+from collections.abc import Iterator, Mapping, Sequence
+from contextlib import contextmanager
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, NewType, Protocol
 from uuid import UUID, uuid4
@@ -388,6 +389,11 @@ class PlotOrchestrator:
         self._instrument = instrument
         self._instrument_config = instrument_config
         self._config_store = config_store
+        # Suppresses per-mutation persistence during composite lifecycle
+        # operations (startup replay, shutdown). See ``_suppress_persist``.
+        # Read/written only on the IOLoop thread, the sole topology writer,
+        # so it needs no lock.
+        self._persist_suppressed = False
         self._plot_data_service = plot_data_service
         self._frame_clock = frame_clock or FrameClock()
         self._logger = structlog.get_logger()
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1112, opened 2026-07-23
Run Kafka integration tests in CI

SimonHeybrock:
The `tests/integration/` harness (fake data sources feeding the real `monitor_data`/`detector_data`/`data_reduction` services plus the dashboard backend over real Kafka) existed but no workflow ever ran it. This adds a CI job with a single-node KRaft service container running `tox -e integration`, in parallel with the unit-test job. Failure output is emitted as check-run annotations because raw log download is firewalled in some dev environments.

Getting the suite green on a pristine broker surfaced real rot, fixed here:

- **Lost-command bug in the Kafka consumer (src)**: manual partition assignment relied on `auto.offset.reset=latest`, which is resolved lazily at first fetch — a message produced between `assign()` and that fetch is skipped silently. On a cold broker this reliably swallowed the workflow start command (no ACK, no job, no data, while heartbeats flowed). Plausible source of lost commands in production; the assignment is now pinned to the broker high watermark, making "everything produced after assignment is consumed" deterministic.
- **Topics were never created by the harness**: consumers fail fast on missing topics and auto-creation only triggers on produce, so the suite only ever passed against brokers holding topics from earlier runs. Fixtures now pre-create the instrument's topics from its `StreamMapping`.
- **Readiness gate matched a log message that no longer exists** ("Kafka consumer ready and polling" vs the actual `kafka_consumer_ready`), silently burning the 10 s timeout on every service start and leaving tests to race service startup.
- **Wait helpers spoke the pre-restructure data plane**: they matched `ResultKey.job_id` on `DataService` keys, which are `DataKey` (workflow, source, output) since the delivery restructure. Jobs are now matched by workflow + source, with a guard test that another workflow's data does not satisfy the wait.
- On timeout the helpers now log DataService keys, job statuses, and per-topic watermark offsets — the diagnostics that pinpointed the lost command.

Part of #1097; enables automating its backend-dependent checklist items (job adoption, clear-at-commit, ROI spectra, ACK expiry) as integration tests in a follow-up PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,8 @@ jobs:
         if: failure()
         shell: bash
         run: |
-          sed -n '/=== FAILURES ===/,$p' integration-output.txt | tail -n 160 > summary.txt
+          sed -n '/=== FAILURES ===/,/Captured stdout/p' integration-output.txt | head -n 140 > summary.txt
+          sed -n '/short test summary info/,$p' integration-output.txt >> summary.txt
           if [ ! -s summary.txt ]; then
             tail -n 60 integration-output.txt > summary.txt
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,21 +99,25 @@ jobs:
         if: failure()
         shell: bash
         run: |
+          set +e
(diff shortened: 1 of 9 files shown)


SimonHeybrock:
LGTM

PR #1113, opened 2026-07-23
Automate Kafka-free items of the #1097 verification checklist

SimonHeybrock:
Automates the checklist items of #1097 that need no Kafka backend, so the recurring manual verification round shrinks to the backend-dependent items.

**Unit test**: `TemporalBuffer` with a datetime64 time coord keeps accepting and correctly trimming data through repeated wraparound past capacity (#1087) — newest retained, oldest dropped, time coord stays monotonic datetime64. The existing datetime64 regression test only covered the first trim.

**Browser tests** (Playwright against the fake backend, each on an isolated server):
- session reload restores tabs and live plot updates;
- a grid created in one session appears in other sessions without changing their active tab (#1040);
- a cell title survives a no-op Save of the cell-properties modal (#1083);
- disabling the first grid and removing a middle grid keep the remaining tabs resolving and updating (#1088).

The shared ColumnDataSource-fingerprint and condition-wait helpers move from the multisession smoke test into `tests/helpers/browser.py`. The cell titlebar pencil gains a per-cell `lt-cell-r{row}c{col}` automation hook (with a unit guard test): a rebuilt cell moves to the end of the DOM, so DOM order cannot address a specific cell.

Browser tests are excluded from CI and the default run; run them locally with:

```sh
python -m pytest -m browser tests/dashboard/
```

Verified locally: three consecutive full browser-suite runs green, fast unit suite green.

Part of #1097.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -59,7 +59,11 @@ no text, `title`, or `aria-label` — leaving nothing semantic for browser autom
 `lt-wf-*` and `lt-grid-*` slug different things on purpose: a workflow has a stable
 *name* identity to slug (`workflow_status_widget.py`'s `_tool_css_class`), but a grid has
 none, so `lt-grid-*` slugs the grid *title* instead (`plot_grid_manager.py`) — the same
-title that also drives the grid's download filename.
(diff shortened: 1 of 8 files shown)


SimonHeybrock:
LGTM

PR #1114, opened 2026-07-23
Integration tests for job adoption, clear-at-commit, ROI spectra, and ACK expiry

SimonHeybrock:
Automates four items of the field-verification checklist in #1097 as real-Kafka integration tests, running the actual service stack (fake sources + monitor_data/detector_data) against the dashboard backend:

- **Job adoption (ADR 0008)**: stop the dashboard while a job runs, create a fresh backend, assert the job is adopted from heartbeat observation (same job_number, ACTIVE, data flows) with no command written (commands-topic watermark unchanged).
- **Clear-at-commit**: let a monitor workflow's cumulative output grow, recommit, assert the first post-commit total drops below the pre-commit value.
- **ROI spectra**: publish rectangle ROIs through the dashboard's ROIPublisher path and assert the detector view's roi_spectra output tracks the published set (first detector-pipeline coverage in the integration suite).
- **Command-ACK expiry**: with no services running, a start command expires after `PENDING_COMMAND_TIMEOUT_SECONDS` and surfaces as the error notification the UI would toast; the job never reports a status.

`DashboardServices.roi_publisher` is made public so tests exercise the same publish path the UI uses; the test backend additionally exposes `roi_publisher` and `notification_queue`. The DIAG timeout diagnostics gain the detector and ROI topics, plus a generic pump-and-wait helper and a topic-watermark probe. CI failure annotations now also capture the pytest ERRORS section (fixture failures were previously unreadable).

One subtlety the clear-at-commit test surfaced: `DataService.clear_keys` empties buffers in place, so after a generation flip a key stays iterable while `__getitem__` raises KeyError until new-generation data arrives. The wait helpers now treat such cleared keys as absent; whether `DataService` should keep violating the Mapping contract there may deserve its own discussion.

Part of #1097.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
           set +e
           sed -e 's/\x1b\[[0-9;]*m//g' integration-output.txt \
             | grep -v 'Consumed 0 messages' > clean.txt
-          awk '/=== FAILURES ===/{found=1} found && n<150 {print; n++}' clean.txt > summary.txt
+          awk '/=== (FAILURES|ERRORS) ===/{found=1} found && n<150 {print; n++}' clean.txt > summary.txt
           grep 'DIAG' clean.txt >> summary.txt
           sed -n '/short test summary info/,$p' clean.txt >> summary.txt
           if [ ! -s summary.txt ]; then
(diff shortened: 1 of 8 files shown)


SimonHeybrock (commented):
LGTM

PR #1115, opened 2026-07-23
Hide cleared keys from the DataService Mapping view

SimonHeybrock:
`clear_keys` (the clear-at-commit generation flip) empties buffers in place so extractor registrations and retention requirements survive — intended. Unintended: the cleared key stayed visible to `__iter__`/`__len__` while `__getitem__` raised `KeyError`, violating the Mapping contract. Any consumer pairing iteration with reads (`items()`, `dict(service)`, the integration wait helpers in #1114) crashed if a recommit landed in between. `__contains__` was already data-based via the Mapping mixin, so membership and iteration also disagreed.

Iteration and `len` now skip empty buffers, probed via a new cheap `is_empty` on the buffer protocol rather than assembling buffered data. "Exists but empty" is only reachable through `clear_keys` (buffer creation and first write share one lock hold), so the observable change is confined to the formerly broken window. The #1114 helper workaround for cleared-but-present keys is reverted.

Found by the clear-at-commit integration test while automating the #1097 checklist.

### DIFF
--- a/src/ess/livedata/dashboard/data_service.py
+++ b/src/ess/livedata/dashboard/data_service.py
@@ -417,14 +417,25 @@ class DataService(MutableMapping[K, V]):
             self._notify()
 
     def __iter__(self) -> Iterator[K]:
-        """Iterate over keys."""
+        """Iterate over keys that hold data.
+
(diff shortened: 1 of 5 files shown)


SimonHeybrock (commented):
LGTM

Issue #1116, opened 2026-07-23
Second round of automated verification checks (follow-up to #1097)


SimonHeybrock:
Follow-up to #1097: automating that checklist (#1112, #1113, #1114) left a set of identified-but-unimplemented checks. The big wins from that analysis are being implemented directly (browser suite in CI; adoption re-stop and service-crash integration tests; DataService concurrency stress test); this issue collects the remainder so they don't evaporate.

## Integration tests (real Kafka, harness in `tests/integration/`)

- [ ] **Stop path end-to-end**: a commanded stop actually stops backend work — the data-topic high watermark stops advancing — not just the status flip in the dashboard.
- [ ] **Rapid recommit under a live producer**: recommit several times in quick succession; only the latest generation's data is admitted, stale-generation drops are counted (`ActiveJobRegistry.record_stale`). The generation filter is unit-tested; the race against a live producer is not.
- [ ] **Poison message liveness end-to-end**: publish a malformed `ev44` (corpus: `tests/helpers/hostile_wire.py`) into the detector/monitor topic mid-job; the service keeps processing subsequent good data. Adapter-level robustness tests exist; the service-over-real-Kafka liveness claim is untested.

## Browser tests (fake transport, `pytest -m browser`)

- [ ] **Session churn**: open/close/kill N sessions in a loop; the server stays responsive and the session count returns to baseline. Behavioral leak detection for reaper teardown (#955/#1095) without needing production CPU observation.
- [ ] **Concurrent edits**: two sessions editing the same grid's properties simultaneously — last-writer-wins without crash or corrupted layout.
- [ ] **Wake-on-commit timing** (once #1096 lands): first plot update within a latency budget after commit; no recompute activity while idle. Nothing currently tests the delivery *timing* that the wake-on-commit work changes.

## Infrastructure enabler

- [ ] **Fake-transport ROI loopback**: `FakeBackendTransport` wires `roi_sink=NullMessageSink()`, so ROIs drawn in the UI are discarded and `roi_spectra` cannot respond. Looping ROI requests back into synthesized `roi_spectra` unlocks browser-level ROI tests — including the cross-session ROI visibility check that currently remains manual on #1097.

## Considered and deliberately excluded

Broker-outage/reconnect chaos tests (docker-pause games in CI, high flake risk for the signal gained) and throughput gates on shared runners (better tracked as a metric than enforced as a gate).

---
_Generated by [Claude Code](https://claude.ai/code)_

[...]

PR #1117, opened 2026-07-23
Stress test: concurrent ingestion, generation flips, and Mapping reads

SimonHeybrock:
DataService is hit concurrently in production — pump-thread ingestion, UI-thread `clear_keys` generation flips, dict-style readers — but `TestThreadSafety` had no case mixing all three. The new stress test hammers them together and asserts the Mapping view never raises: iteration yields only keys holding data (the #1115 contract), `get` absorbs the inherent iterate-then-read race, and `len`/membership stay consistent. The docstring records the design boundary: unguarded `__getitem__` across that window is racy by design; `snapshot` is the atomic API.

Stacked on the #1115 fix (retargets to main when it merges). Part of the #1097 follow-up (#1116 context).

### DIFF
--- a/tests/dashboard/data_service_test.py
+++ b/tests/dashboard/data_service_test.py
@@ -1390,6 +1390,69 @@ class TestThreadSafety:
         assert not errors, f"concurrent access raised: {errors[0]!r}"
         assert service["key1"].value >= 0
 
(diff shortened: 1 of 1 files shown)


SimonHeybrock (commented):
LGTM

PR #1118, opened 2026-07-23
Run the Playwright browser suite in CI

SimonHeybrock:
The Playwright browser tests (`pytest -m browser`: multi-session smoke test and dashboard UI flows, fake backend, no Kafka) ran nowhere in CI and could silently rot. This adds a `browser` tox env (installs playwright and its matching chromium) and a CI job modeled on the integration job: chromium system libraries are installed via `playwright install-deps` outside tox, and failures are surfaced as check-run annotations. Docs claiming the suite is excluded from CI are updated; it remains excluded from the default pytest run and the other tox envs.

Part of the #1097 follow-up work.

Note: browser tests are the most flakiness-prone job in this workflow (real chromium, timing-sensitive UI assertions on 4-core runners). Watch it across a few PRs before considering it as a required check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,57 @@ jobs:
           done
           true
 
+  browser:
+    # Playwright UI tests: a real chromium driving the dashboard served by the
+    # in-process fake backend (no Kafka), covering multi-session sync and grid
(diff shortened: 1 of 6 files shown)


SimonHeybrock:
LGTM

PR #1119, opened 2026-07-23
Integration tests: stop reconciliation and service crash (ADR 0008)

SimonHeybrock:
Two real-Kafka integration tests for the remaining automatable items of the #1097 checklist, against ADR 0008 (job adoption / desired-observed reconciliation).

**Stop reconciliation** (`reconciliation_restop_test.py`): freezing the backend service with SIGSTOP makes a commanded stop sit unacted while the job's observed status stays fresh; reconciliation re-issues the stop after `STOP_REISSUE_INTERVAL_SECONDS`, observed as an extra commands-topic message no user action produced. The checklist's restart variant ("stop commanded before the restart is re-stopped by reconciliation") is not what the architecture does: desired-stopped is not persisted, so after a restart an absent `current_job` record is indistinguishable from a record miss, and ADR 0008 resolves that toward adoption ("record misses degrade, they do not lie"). The second test pins that actual guarantee — the still-running job is adopted record-less (no spurious commands, degraded provenance) and remains stoppable. If re-stop-after-restart is the desired semantics instead, the stop intent would need to be persisted (e.g. a tombstone for the stopped job_number); that is a product change, flagged on #1097.

**Service crash mid-run** (`service_crash_test.py`): a SIGKILLed `monitor_data` worker (no terminal heartbeat) surfaces as stale-while-`running` in `ServiceRegistry` after the 30 s heartbeat timeout — the crash signature, distinct from a clean shutdown whose terminal `stopped` status is never stale. The job-status tier (60 s window) intentionally still shows the last state at that point, and the data-topic watermark stalls immediately and stays put.

Helpers: `wait_for_watermark_stall` (proves a producer went quiet) and a `service_registry` accessor on `DashboardBackend`.

Part of #1097.

### DIFF
--- a/tests/integration/backend.py
+++ b/tests/integration/backend.py
@@ -134,6 +134,11 @@ class DashboardBackend:
         return self._services.job_service
 
     @property
+    def service_registry(self):
+        self._check_available()
+        return self._services.service_registry
+
+    @property
     def plotting_controller(self):
         self._check_available()
         return self._services.plotting_controller
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -75,6 +75,45 @@ def topic_high_watermark(topic: str) -> int:
         consumer.close()
 
 
+def wait_for_watermark_stall(
+    topic: str, *, settle: float = 4.0, timeout: float = 40.0
+) -> int:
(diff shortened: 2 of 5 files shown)


SimonHeybrock (commented):
LGTM

PR #1121, opened 2026-07-24
Show stopped/error status in the cell titlebar pill

### LINKED ISSUE #1120: Default cell header does not show stopped-job indication
The black frames indicated that a job was stopped are only visible when switching to the expanded state, on the per-layer headers:

Image

The lag pill is an indicator that no data is flowing. Is this enough? Do we still want the black frames? Think about whether we can have a more coherent approach.

SimonHeybrock:
Fixes #1120.

## Motivation

The titlebar pill only showed data age, so a stopped job decayed to the same red as a broken pipeline. That signal is wrong twice: stopping is deliberate and the frozen snapshot is valid (no alarm warranted), and once users learn "red often just means stopped" they start ignoring red — alarm fatigue masking real staleness problems. Meanwhile the actual stopped indication (a dark border on per-layer rows) was only visible in the expanded state, and errors on layers that still had a plot were invisible entirely.

## Example

image

## What changed

The pill now speaks one status language at both cell and layer level:

- Any layer errored: red "error" pill in the titlebar (previously invisible while a plot was still shown).
- All dynamic layers stopped: neutral gray "stopped" pill with a square stop glyph, frozen — no age counting on a deliberate snapshot. Static overlays are ignored for the aggregate.
- Otherwise: freshness bands as before, with stopped layers excluded from the age merge so a partial stop cannot drag the band to red while other layers are live.

The stopped border on expanded per-layer rows is replaced by the same pill in miniature ("stopped"/"error" badge at the row's far left, mirroring the titlebar), so the expanded view is detail, not a different encoding. `FreshnessPill` is renamed to `StatusPill` accordingly.

## Test plan

Behavior tests cover the frozen pill (fresh bounds must not overwrite it), the error pill, and partial-stop age filtering, driven through the real poll loop. Verified visually against the fake backend (stopped workflows show gray pills collapsed and expanded, plots retained).

- [x] Manual check on a live instrument: stop a workflow, confirm the titlebar shows the gray "stopped" pill and expanded rows show the badge instead of the border.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/cell.py
+++ b/src/ess/livedata/dashboard/widgets/cell.py
@@ -49,7 +49,7 @@ from .plot_widgets import (
     get_plot_cell_display_info,
     get_workflow_display_info,
 )
-from .styles import Colors, FreshnessPill, StatusColors
+from .styles import Colors, StatusColors, StatusPill
 
 logger = structlog.get_logger(__name__)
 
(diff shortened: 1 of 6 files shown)


SimonHeybrock (commented):
LGTM

PR #1122, opened 2026-07-24
Harden batchers against pathological data-derived timestamps

### LINKED ISSUE #1047: Backend: validate data-derived values at the adapter boundary (timestamps as the service clock)

SimonHeybrock:
## Why

A single message carrying a far-future timestamp could silence a backend service indefinitely (#1038 finding 1). The batchers use data-derived timestamps as their only clock, so such a value became that clock: it anchored the batch window in the future, after which all real traffic looked early, no batch ever closed, and the process kept reporting itself healthy. Restarting re-consumed the same offset and wedged again. Both production batchers were affected; `RateAwareMessageBatcher` — already the configured choice for some instruments — wedged permanently even from a mid-stream timestamp, where `SimpleMessageBatcher` merely degraded.

## What changed

Window placement no longer follows a single outlier arbitrarily far. Batch anchors come from the bulk of the traffic rather than a bare maximum (`plausible_anchor`), the simple batcher refuses to advance its window past the frontier of the data in hand, and the rate-aware batcher diverts disjoint-epoch messages straight to delivery and bounds its high-water mark relative to the active window.

The rate-aware batcher additionally gains a wall-clock liveness backstop (#1123): it takes an injectable monotonic clock, and when no batch has closed for a bounded wall-clock interval while traffic is buffered, the window is re-placed at the plausible anchor of that traffic. This replaces an earlier per-pathology re-anchoring mechanism whose trigger conditions needed individual reachability proofs (one turned out to be dead code); the backstop corrects any placement the buffered traffic cannot close, without diagnosing how the window got misplaced. Wall time gates liveness only, never window placement or sizing, so replay/catch-up batching is unaffected — the module docstring documents this clock policy. `timeout_factor` is now validated against the high-water-mark plausibility horizon (a larger finite value could never fire); an explicit `inf` requests gate-only closure.

No timestamps are modified anywhere: this layer only constrains how window placement reacts to them, so it is safe for any source, Kafka or not.

Two holes remain in the simple batcher, each pinned by a strict xfail: a lone far-future timestamp in the startup backlog (one message on each side of a gap carries no evidence to weigh), and a parked future message that fast-forwards the window through a silence gap. Both need arrival-time evidence where messages enter the system; a follow-up PR adds that validation at the adapter boundary and closes #1047. The rate-aware batcher recovers from both via the backstop.

## Test plan

Unit tests for the anchor selection, the window-advance guard, the backstop (fires only after the wall-clock threshold with traffic buffered; quiet streams never advance the data clock on wall-time evidence), recovery from a poisoned gap jump plus silence gap, and a full forward epoch jump; wall time is driven by an injected fake clock in lockstep with data time. Slowly creeping device clocks (seen in production at ~1 s/day) are pinned separately: delivery flows through every threshold crossing in both directions, a persistent minutes-scale offset with intermittent traffic does not ping-pong, and the window-tracks-the-fastest-clock equilibrium is documented as current behavior — the resulting alignment/data-clock skew is a known quality degradation whose detection belongs at the adapter boundary. The service-level liveness tests run against both `SimpleMessageBatcher` and `RateAwareMessageBatcher`, so the batcher already deployed in production is covered by the same invariants as the default.

### DIFF
--- a/src/ess/livedata/core/message_batcher.py
+++ b/src/ess/livedata/core/message_batcher.py
@@ -22,6 +22,52 @@ class MessageBatch:
     messages: list[Message[Any]]
 
 
+# Largest jump between consecutive timestamps that still reads as continuous
+# traffic, in multiples of the batch length.  Data-derived timestamps are the
+# batchers' only clock, so window placement must not follow one outlier
+# arbitrarily far into the future: an over-anchored window stalls delivery
+# until wall time catches up, while an under-anchored one self-heals forward
+# as newer traffic arrives.  The cap holds regardless of where a message
+# came from: sources are not trusted to deliver sane timestamps, and the
+# batcher is the one place every message passes through.
+MAX_TIMESTAMP_AHEAD_BATCHES = 3
+
+
+def plausible_anchor(timestamps: list[Timestamp], batch_length: Duration) -> Timestamp:
+    """Newest timestamp still connected to the bulk of the given traffic.
+
+    Batch windows are anchored at the newest timestamp seen.  A bare ``max``
+    lets a single pathological far-future timestamp pin the window (and with
+    it the batcher's data-derived clock), stalling delivery until wall time
+    catches up.
+
+    Splitting the sorted timestamps at every jump wider than the horizon
+    yields groups of mutually-connected traffic; the anchor is the newest
+    timestamp of the largest group.  Membership of the bulk is what marks a
+    timestamp as genuine, not distance from an average: a burst spread evenly
+    over many batch lengths stays one group and keeps its true maximum, while
+    a tight cluster plus one stray yields the cluster -- whether the stray
+    sits ahead of the traffic or behind it, as a stream in a disjoint epoch
+    does.
+
+    Ties go to the later group, and a lone timestamp is necessarily its own
+    anchor.  Neither case has evidence to weigh, and anchoring behind the
+    traffic is not the safe default it appears to be: a window only walks
+    forward one batch length per call, so an anchor stranded in a disjoint
+    past epoch takes astronomically many calls to catch up, where one
+    stranded ahead merely waits out its own distance.
+    """
+    ordered = sorted(timestamps)
+    horizon = MAX_TIMESTAMP_AHEAD_BATCHES * batch_length
+    groups = [[ordered[0]]]
+    for timestamp in ordered[1:]:
+        if timestamp - groups[-1][-1] > horizon:
+            groups.append([])
+        groups[-1].append(timestamp)
+    largest = max(len(group) for group in groups)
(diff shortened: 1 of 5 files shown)


SimonHeybrock (commented):
This seems reasonable after several iteration rounds. Fixing this is critical, but I think it is impossible to think through all edge cases, so review is far from confident -- I focussed on reading (new) tests.

PR #1124, opened 2026-07-27
Accept ev44 payloads with absent event vectors

SimonHeybrock:
Fixes finding 2 of #1038.

The ev44 schema does not require the event vectors, so a producer may legitimately omit them. When it does, the flatbuffers-python `*AsNumpy()` accessors return scalar `0` rather than an empty array, both ev44 adapters raise on `.size`, and the adapter-level catch-all drops the message and records it as `<error>` — the Kafka-timestamp fallback that exists for exactly this case is never reached.

Such a payload now adapts to an empty event list with the timestamp taken from the broker. Beyond the monitor adapter named in the finding, the same bug was present in `KafkaToEv44Adapter`, and the multi-pulse guard downstream indexed `reference_time_index[0]` unconditionally, so an empty index would have dropped the message anyway one step later. "No reference time" now means "no pulse", which is not multi-pulse.

The payload consequently leaves the malformed-payload corpus in `tests/helpers/hostile_wire.py`: that corpus's contract is "must be contained/dropped", which is no longer what this payload should get. Its service-liveness coverage is replaced by a dedicated test.

Note this is orthogonal to finding 1 (unvalidated data-derived timestamps, #1047) — those strict xfails are untouched and still fail.

## Test plan

- [x] Existing strict xfail `test_ev44_without_event_vectors_falls_back_to_kafka_timestamp` flips to passing, marker removed
- [x] New coverage for the detector adapter chain, `from_ev44` with no reference times, and service liveness after such a payload
- [x] Full fast suite green, with the #1047 xfails still xfailing

### DIFF
--- a/src/ess/livedata/kafka/message_adapter.py
+++ b/src/ess/livedata/kafka/message_adapter.py
@@ -5,6 +5,7 @@ from dataclasses import replace
 from typing import Any, Generic, Protocol, TypeVar
 
 import numpy as np
+import numpy.typing as npt
 import pydantic
 import scipp as sc
 import streaming_data_types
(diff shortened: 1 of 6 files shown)


SimonHeybrock:
I don't think this is the right approach. Have not seen problems in practice, and using broker timestamp might have unexpected consequences.

PR #1125, opened 2026-07-27
Apply masks as NaN when serializing to da00

### LINKED ISSUE #1106: Masks are silently dropped by scipp_to_da00 (DREAM vanadium-normalized outputs reach dashboard unmasked)
## Masks are silently dropped by `scipp_to_da00` — DREAM vanadium-normalized outputs reach the dashboard unmasked

`scipp_to_da00` (`src/ess/livedata/kafka/scipp_da00_compat.py`) serializes only signal, errors, and coords. Masks on a published DataArray are silently discarded — no warning, no error.

This is not hypothetical: DREAM's `powder_reduction_with_vanadium` normalizes by vanadium via `ess.powder.correction` (`normalize_by_vanadium_dspacing` and friends), which attaches a `zero_vanadium` mask over bins where the vanadium histogram is zero — in live operation plausibly at range edges while statistics accumulate. The division still produces values (inf/garbage) in those bins; the mask is what marks them invalid. With the mask stripped on the wire, the dashboard renders those bins as if they were data.

`i_of_tof` goes through `transform_coords`, which preserves the mask, so all three normalized outputs (`i_of_dspacing`, `i_of_dspacing_two_theta`, `i_of_tof`) are affected.

Options, roughly in order of preference:

- Serialize masks (da00's Variable list can carry them by naming convention, as the schema is generic) and reattach on decode. Needs agreement on a naming scheme since da00 has no mask concept.
- Apply masks before serializing (`da.data = sc.where(mask, nan, da.data)`-style, or drop masked bins into NaN). Loses the mask/data distinction but is honest on screen and cheap.
- At minimum: fail fast or log loudly when a published output carries a mask, instead of the silent drop.

Found during the #1057 wire inventory; independent of the schema work there (a dedicated result schema would face the same question, and its current proposal falls back to da00 for anything it cannot represent — which does not help while da00 itself drops masks).

---

SimonHeybrock:
Fixes the silent mask drop in `scipp_to_da00` reported in #1106.

Masked elements reached the dashboard indistinguishable from valid data. DREAM's vanadium normalization attaches a `zero_vanadium` mask over bins where the vanadium histogram is zero — in live operation plausibly at range edges while statistics accumulate. Those bins arrived as inf/garbage and were plotted as data.

## Why NaN rather than transporting masks

Not because da00 cannot carry masks — its variable list is generic enough to do so by naming convention. The obstacle is the consumer side: nothing in the dashboard reads `DataArray.masks`, neither the buffering and extraction chain nor the plotters, so a transported mask would be dropped one layer later instead. A time-varying mask would additionally break `TemporalBuffer`, whose `sc.identical` metadata check would then fail on every message and reset the buffer.

NaN, by contrast, is honored end to end: transparent image pixels, gaps in curves, finite-only autoscaling and log-scale clim, `nansum`/`nanmean` window aggregation, `-` in tables. Folding masks into the values is therefore honest on screen and needs no dashboard changes.

This discards the mask/data distinction deliberately. Adding mask support across transport, dashboard data chain, and plotters is very much on the table if that distinction turns out to be needed.

Side effect worth noting: `normalize_by_vanadium` divides by zero and produces `inf` in the masked bins, which previously reached the plots. Those are now NaN gaps.

Scope: serialization is a genuine chokepoint — `scipp_to_da00` is the only path any workflow result takes to Kafka, and nothing in the backend consumes `LIVEDATA_DATA`, so NaN cannot propagate into further reduction. Unmasked data is untouched, so there is no copy and no dtype change on the common path. Masked integer data is promoted to float64, since it cannot hold NaN.

Closes #1106.

## Test plan

- [x] Unit tests for NaN values, NaN errors, multi-mask OR, mask broadcast over missing dims, integer promotion, dtype preserved when unmasked, and full da00 round-trip.
- [x] Full fast suite passes.

### DIFF
--- a/src/ess/livedata/kafka/scipp_da00_compat.py
+++ b/src/ess/livedata/kafka/scipp_da00_compat.py
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+import operator
+from functools import reduce
+
 import numpy as np
 import scipp as sc
 from streaming_data_types import dataarray_da00
@@ -21,12 +24,13 @@ def scipp_to_da00(
 ) -> list[dataarray_da00.Variable]:
     # Encode DataArray.name in the 'label' field of the signal variable
     label = da.name if da.name is not None else None
-    if da.variances is None:
-        variables = [_to_da00_variable(signal_name, da.data, label=label)]
+    data = _masked_to_nan(da) if da.masks else da.data
+    if data.variances is None:
+        variables = [_to_da00_variable(signal_name, data, label=label)]
     else:
         variables = [
-            _to_da00_variable(signal_name, sc.values(da.data), label=label),
-            _to_da00_variable('errors', sc.stddevs(da.data)),
+            _to_da00_variable(signal_name, sc.values(data), label=label),
+            _to_da00_variable('errors', sc.stddevs(data)),
         ]
     variables.extend(
         [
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1126, opened 2026-07-27
Remove reset for workflows where it is meaningless or harmful

SimonHeybrock:
Addresses finding 3 of #1037.

## Why

`JobAction.reset` clears only the workflow, not the shared `ToNXlog` preprocessor buffer. For a timeseries job this resets the publication index to zero, so the next finalize re-emits the entire history as a `'delta'` that every downstream buffer appends unconditionally — duplicating every point already displayed (and silently double-counting in correlation histograms using the stream as an axis). Clearing the preprocessor instead would be worse: it is shared by reference across all jobs consuming the stream (e.g. BIFROST's angle lookup table).

Timeseries outputs represent state, not accumulation since a start point, so reset has no meaningful semantics for them. Restarting the workflow already provides correct clear-and-refill behavior: the commit path's generation flip clears the dashboard buffers, and the freshly activated job is seeded with the full cached history. The wavelength LUT similarly recomputes from retained chopper history on the next trigger, so reset is a pointless affordance there too.

## What

- New `WorkflowSpec.supports_reset` flag, `False` for the timeseries and wavelength-LUT workflows.
- The dashboard disables the reset button for such workflows, with a tooltip explaining why (left-anchored so it is not clipped at the viewport edge).
- `JobManager.reset_job` rejects reset commands for such jobs — the UI no longer sends them, so any arrival is a programming error and surfaces as a logged `job_command_failed`.
- `TimeseriesOutputs` now declares its output view instead of relying on synthesis, which labelled the delta field `since_start` (run-cumulative — the opposite of its per-update semantics) and exposed the raw field name `delta` as the view name; the view is now titled "Timeseries".

## Test plan

- Unit tests for the backend guard and the disabled-button state.
- Verified in the browser (fake backend, dummy instrument): timeseries reset button renders disabled with the explanatory tooltip; other workflows' reset buttons unaffected.

### DIFF
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -589,6 +589,7 @@ class Instrument:
         outputs: type[Any],
         device_outputs: dict[str, str] | None = None,
         reset_on_run_transition: bool = True,
+        supports_reset: bool = True,
     ) -> SpecHandle:
         """
         Register workflow spec, return handle for later factory attachment.
@@ -648,6 +649,7 @@ class Instrument:
             outputs=outputs,
             device_outputs=device_outputs or {},
             reset_on_run_transition=reset_on_run_transition,
+            supports_reset=supports_reset,
         )
         return self.workflow_factory.register_spec(spec, service=service)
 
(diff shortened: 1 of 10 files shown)


SimonHeybrock (commented):
LGTM

PR #1128, opened 2026-07-27
Fix the flaky Kafka integration job

SimonHeybrock:
Fixes the two failure modes behind the flaky Kafka integration job. Both hit `clear_at_commit_test::test_recommit_clears_accumulated_data` — which is the *first* test collected, not the last: pytest prints errors at the end, so `8 passed, 1 error` reads misleadingly.

## Consumers failed closed during partition leader election

Three of the last four failures were `NOT_LEADER_FOR_PARTITION` while pinning start offsets, on the very first consumer created after the topics are made on a pristine broker.

Leadership is not established atomically with topic creation, only the leader answers `ListOffsets`, and librdkafka does not retry that query on our behalf — so we turned an error the broker considers transient into a hard startup failure. This is not CI-specific: the same window opens on any broker restart, so a routine election would take down every service that happens to start inside it.

The watermark fetch now waits out the errors that mean "no usable leader yet" under a single deadline shared by all topics, and still fails fast on everything else. `KafkaError.retriable()` is no help here — librdkafka reports `False` for all of these codes, so the set is explicit.

## The clear-at-commit assertion was decided by producer noise

The fourth failure was the test's own comparison, not a product race. The generation machinery is sound: `DataService` has a single writer, which re-checks the wire `job_number` per message under the same lock that performs the clear, so old-generation data cannot become readable after a flip.

The test compared two accumulation windows of the same producer while controlling only one of them. The dashboard exposes the latest ingested value rather than the new generation's first result, and its background thread ingests at 20 Hz while the test samples at 2 Hz. With a pre-commit bar built from a single update interval, the outcome came down to per-batch producer noise (sigma is roughly 2% of the bar; the observed discrepancy was 2.4%) and poll timing.

The bar is now a fixed accumulation window that dwarfs the batch or two the post-commit observation can span, the post-commit poll is tightened so the captured value stays near the new generation's start, and the assertion demands a clear drop rather than any drop.

Also restored: reading a key can race a generation flip from the ingestion thread, because iteration hides cleared keys but takes the lock separately from each read. Such a key now reads as absent instead of raising `KeyError`. Tightening the poll makes that window more likely to be hit.

## Test plan

- New unit tests cover the leader wait: transient errors are waited out and the assignment still pins the high watermark, the wait gives up at its deadline, and unrelated errors raise immediately.
- Full unit suite passes locally (4129 passed).
- The integration suite could not be run locally — this devcontainer has neither Docker nor Java, so no broker. The clear-at-commit change is unverified against a real pipeline and needs a run of the Kafka job to confirm.

### DIFF
--- a/src/ess/livedata/kafka/consumer.py
+++ b/src/ess/livedata/kafka/consumer.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+import time
 import uuid
 from collections.abc import Generator
 from contextlib import contextmanager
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1130, opened 2026-07-27
Make the rate-aware batcher the default

SimonHeybrock:
## Why

The rate-aware batcher prevents oscillation in outputs, steming from a variable number of source messages contributing to a given output message. This has been tested for a while already in PROD for BIFROST. We now switch this on for all instruments.

As a secondary reason, with the hardening and wall-clock liveness backstop from #1122 in place, the rate-aware batcher recovers from every timestamp pathology in the liveness suite, while the simple batcher retains two wedges pinned as strict xfails (lone far-future startup timestamp; parked future message across a silence gap). Flipping the default gives every instrument the backstop and builds production confidence toward removing the simple batcher entirely; `--batcher simple` stays available as a rollback until then.

## What changed

The `--batcher` CLI default flips to `rate-aware`; help text and docs updated to match. The Python-level `AdaptiveMessageBatcher` inner default remains `SimpleMessageBatcher` (`message_batcher.py` cannot import the rate-aware module without a cycle; production wiring goes through the CLI).

## Test plan

The service-level hostile-input liveness suite already runs against both inner batchers, so the new default is covered by the same invariants as before the flip. Deployment configs that pass `--batcher rate-aware` explicitly become redundant but stay valid.

- [ ] After merge and deploy: monitor `batcher_stall_recovery` warnings and per-instrument output cadence for a soak period before opening the simple-batcher removal PR.

### DIFF
--- a/docs/developer/design/message-flow-and-transformation.md
+++ b/docs/developer/design/message-flow-and-transformation.md
@@ -200,9 +200,13 @@ Groups messages into time-aligned windows for time-aligned processing, job sched
 
 `MessageBatch` = `(start_time, end_time, messages)`. Times in nanoseconds, `[start_time, end_time)` (inclusive, exclusive).
 
+### RateAwareMessageBatcher
+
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

Issue #1058, opened 2026-07-09
Drop redundant scalar 'time' coord; declare temporal role in output spec instead of magic coords


SimonHeybrock:
> ## Correction (2026-07-27)
>
> Implemented in #1132 and #1135. Three claims in the proposal below are wrong; the original text is kept for provenance.
>
> - **"Every result already carries `start_time`/`end_time`, so *all* outputs become bufferable."** Not so. `Job._add_time_coords` deliberately skips any DataArray carrying a `time` coord, timeseries log data being the stated reason, so the timeseries `delta` output has neither bound, by design.
> - **"`TemporalBuffer` keys on `start_time`."** Wrong bound. For a cumulative output `start_time` is pinned for the lifetime of a generation — NICOS depends on exactly that — so keying on it would stack every point of a growth curve on one x. `end_time` is the only bound that advances for both window and cumulative outputs, and it matches the existing lag definition (`now − end_time`). The "or the window midpoint" alternative is undefined for cumulative.
> - **"`temporal_role: window | cumulative`."** Two roles are not enough. The timeseries path is neither: its `time` coord is real per-sample data, not an annotation. The landed `Temporality` has three values — `window`, `cumulative`, `series`.
>
> The guardrail section is wrong in one direction and understated in the other:
>
> - **Rate normalization needs no gate.** On a cumulative field `_normalize_to_rate` divides by the run duration, giving the mean rate since the run started — a well-defined quantity, and BIFROST's `detector_region_counts` view description already tells users to ask for it. The case that *would* be wrong, rate over a `FullHistoryExtractor`'s whole-buffer bounds, is unreachable: the plotters that require that extractor (`timeseries`, both correlation histograms) expose no rate control.
> - **Window aggregation over a cumulative field is already unreachable from a fresh config**, since the window controls are hidden for views with no per-update field. It stays reachable from a *persisted* config, because `WorkflowSpec.field_for` falls back to the cumulative field — and re-keying the buffer turns that from a loud crash into a silently wrong number. That is what #1135 rejects.
>
> Also stale: the issue cites `core/stream_processor_workflow.py`; the module lives at `workflows/stream_processor_workflow.py`.

## Problem

Window outputs currently carry three scalar time coords, assigned together in `StreamProcessorWorkflow.finalize()` (`stream_processor_workflow.py:235-239`):

```python
results[name] = results[name].assign_coords(
    time=start_time_coord,      # <- always identical to start_time
    start_time=start_time_coord,
[...]

PR #1131, opened 2026-07-27
Stamp window bounds on AreaDetectorView 'current' output

SimonHeybrock:
`AreaDetectorView` discarded the window's `end_time` and stamped only a scalar `time` coord on its `current` output. Because a `time` coord makes `Job._add_time_coords` skip an output (it assumes such data carries its own timestamps), the `current` output reached the dashboard with no temporal bounds at all — neither the workflow's own nor the job-level fallback.

Two dashboard features depend on those bounds and were therefore dead for every area detector view: the cell titlebar freshness/lag pill (`_compute_time_bounds`) and counts-per-second normalization (`_normalize_to_rate`, which needs `end_time - start_time`). Both failed silently, so the plots looked fine while showing no lag and unnormalized counts.

`AreaDetectorView` now tracks the window end time and assigns `time`, `start_time` and `end_time` together, matching the contract every other window output follows (`StreamProcessorWorkflow.finalize`). The `cumulative` output deliberately keeps no time coords so it continues to be stamped with the job-level range.

Found while investigating #1058. The scalar `time` coord is redundant with `start_time` here, but #1058 will drop it from all window outputs uniformly — diverging from the shared contract in this PR would only make that change harder.

### DIFF
--- a/src/ess/livedata/workflows/area_detector_view.py
+++ b/src/ess/livedata/workflows/area_detector_view.py
@@ -37,6 +37,7 @@ class AreaDetectorView(Workflow):
         self._cumulative: sc.DataArray | None = None
         self._previous: sc.DataArray | None = None
         self._current_start_time: Timestamp | None = None
+        self._current_end_time: Timestamp | None = None
 
     def accumulate(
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1132, opened 2026-07-27
Declare output temporality instead of signalling it with a magic coord

SimonHeybrock:
Whether a workflow output can be plotted as a history — and whether summing successive messages of it is meaningful — was decided by a coord-name convention: a scalar `time` coord on the data (and on the output template) meant "per-window scalar". Nothing declared it, so the rule was enforced only by a runtime `ValueError` deep inside `TemporalBuffer`, and `find_timeseries_outputs` had to sniff output templates to rediscover it.

`Temporality` states the fact once, per output field: `window` (covers `[start_time, end_time)`, successive windows disjoint), `cumulative` (value as of `end_time`, with `start_time` pinned for the generation), or `series` (carries its own per-point `time` axis). Fields declare it by their annotation — `current: WindowOutput = pydantic.Field(...)` — which is `Annotated` metadata, the documented home for this kind of per-field fact; a bare `sc.DataArray` still means cumulative, so reduction outputs need no annotation.

The three-way split matters because the timeseries path is neither of the first two. Its `time` coord is real per-sample data, and it deliberately carries no `start_time`/`end_time` at all — `Job._add_time_coords` skips it. Issue #1058 proposes two roles and assumes every output already carries both bounds; that assumption does not hold, which is what motivated the third value here.

Three duplicated declarations collapse into it:

- The `window_outputs` lists passed to `StreamProcessorWorkflow` are derived from the outputs model rather than hand-maintained beside it.
- `find_timeseries_outputs` asks for the declaration instead of inspecting template coords.
- `OutputView.fields` no longer keys backing fields by `Windowing`. That key was never independent information — checked across all 157 output views of all instruments, it is a total function of `Temporality` (`window` and `series` back `per_update`, `cumulative` backs `since_start`). Views now list their member fields and the binding follows, so there is no second place to state it and nothing to keep in sync. Resolution needs the outputs model, so it lives on `WorkflowSpec` (`field_for`, `windowing_options`).

Selecting the same window mode from two fields of one view would be ambiguous, so a validator rejects it at registration — usually a missing annotation, since unannotated fields default to `cumulative`.

No behaviour change: the derived `window_outputs` reproduce the previous hand-written lists exactly, the derived windowing bindings reproduce the previously declared ones for every view, and correlation-axis discovery selects the same output views as before.

`WorkflowSpec.outputs` is tightened to `type[WorkflowOutputsBase]`, since every real outputs model already derives from it and the new accessors live there.

## Groundwork for #1058

This is the declaration half of #1058's third proposal. Follow-ups, in order:

- Re-key `TemporalBuffer` on `end_time` rather than the scalar `time`. Note #1058 proposes `start_time`, which does not work: for cumulative outputs `start_time` is pinned at generation start, so every point of a growth curve would land on the same x. `end_time` is the only bound that advances for both window and cumulative outputs.
- Drop the redundant scalar `time` coord and the `with_time_coord` templates, which is where #1058's da00 saving lands.
- Gate window aggregation and rate normalization on `Temporality.window` via `DataRequirements`, turning today's buried `ValueError` into a capability distinction at plot-configuration time.

Depends on #1131 for the second bullet: `AreaDetectorView`'s `current` output currently has no `start_time`/`end_time` at all.

## Test plan

- `pytest` full fast suite green.
- New unit tests for `temporality()` / `fields_with()`, including inheritance and override.
- New cross-check parametrized over all 47 registered workflow specs.

### DIFF
--- a/src/ess/livedata/config/instruments/bifrost/factories.py
+++ b/src/ess/livedata/config/instruments/bifrost/factories.py
@@ -9,6 +9,7 @@ from functools import cache
 import scipp as sc
 
 from ess.livedata.config import Instrument
+from ess.livedata.config.workflow_spec import Temporality
 
 from . import specs
 from .specs import (
@@ -114,7 +115,9 @@ def setup_factories(instrument: Instrument) -> None:
                 DetectorRegionCounts[Cumulative]: cumulative,
                 DetectorRegionCounts[Current]: window,
             },
-            window_outputs=['detector_region_counts'],
+            window_outputs=specs.DetectorRatemeterOutputs.fields_with(
+                Temporality.window
+            ),
         )
 
     # Q-map workflow factories
--- a/src/ess/livedata/config/instruments/bifrost/specs.py
+++ b/src/ess/livedata/config/instruments/bifrost/specs.py
@@ -12,7 +12,7 @@ for full instrument details.
 """
 
 from enum import StrEnum
-from typing import ClassVar, Literal
+from typing import Annotated, ClassVar, Literal
 
 import pydantic
 import scipp as sc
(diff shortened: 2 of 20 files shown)


SimonHeybrock (commented):
LGTM, tested with monitor workflow outputs

PR #1134, opened 2026-07-27
Exclude correlation points predating the axis history

### LINKED ISSUE #1075: Correlation histogram silently drops early data points when timestamp overlap begins
**Severity:** Medium

`_make_lookup` (`src/ess/livedata/dashboard/correlation_plotter.py:142-157`) picks the scipp bin-lookup mode based on whether the histogrammed data's timestamps currently precede the axis data's timestamps:

```python
mode = 'nearest' if data_max_time < axis_min_time else 'previous'
```

This mode is recomputed and reapplied to the *entire accumulated history* on every call to `compute()` (lines 198-254), because both the "primary" (histogram) and axis data roles use a full-history extractor, so `compute()` re-evaluates from scratch each time new data arrives (lines 234-252).

**Concrete sequence:** while all histogram-data timestamps are still earlier than the first axis reading, `_make_lookup` picks `'nearest'`, and those early points get correlated (incorrectly, but non-destructively) against the axis value nearest to them, i.e. the very first axis reading — measured entirely after the data. Once enough time passes that the two streams' timestamp ranges start to overlap, `data_max_time >= axis_min_time` flips the mode to `'previous'` for the *whole* re-evaluated history in one step. Under `'previous'` mode, any histogram point still timestamped before `axis_min_time` has no preceding axis value, so its assigned coordinate becomes `NaN`. `.hist()`/`.bin()` then silently drops rows with `NaN` coordinates. The result: a set of early points that were visible in the histogram (via the `'nearest'` fallback) abruptly and permanently disappears the moment the mode flips — the displayed histogram totals change discontinuously with no user action and no error.

**Fix hint:** the mode should be a per-point (or at least per-batch) decision, or the fallback for genuinely-pre-axis-history points should be preserved permanently rather than re-derived from a single global flag that flips once. Alternatively, points before `axis_min_time` could be explicitly and consistently excluded (documented data-availability behavior) rather than included-then-silently-dropped.

**Test hint:** build a `CorrelationHistogramPlotter`, call `compute()` first with histogram data entirely preceding the axis data's earliest timestamp (asserting those points appear in the output), then call `compute()` again with the same accumulated history plus new data whose timestamps now overlap the axis range, and assert the original early points either persist or are explicitly and knowingly excluded — not silently vanished.

Extracted from #1039.

SimonHeybrock:
Closes #1075.

The lookup mode for correlation histograms was picked from a global comparison of the histogrammed data's latest timestamp against the axis data's earliest one, and re-derived from scratch on every `compute()` over the full accumulated history. While the data still preceded the axis entirely, `'nearest'` correlated those points with the first axis reading — measured after them. Once the two streams' time ranges started to overlap, the whole history flipped to `'previous'` in one step and the early points, now NaN-coordinated, were silently dropped by `hist()`/`bin()`. Displayed totals changed discontinuously with no user action and no error.

This takes the option sketched at the end of the issue: there is no axis value to correlate a point with before the first axis reading, so such points are excluded explicitly and consistently, whatever the rest of the history looks like. They are trimmed up front rather than included and then dropped by the binning — the excluded set is identical, but it no longer depends on how the two streams happened to line up. A plot with nothing left to show renders nothing (the layer keeps its placeholder) rather than showing a guess, which is what the `'nearest'` fallback was originally added to avoid: `hist()` raising "Empty data range". With the mode no longer data-dependent, the mode switch and its scipp int-coord workaround go away, and the lookups are built once instead of per source.

Two consequences worth knowing:

- The cutoff is the first *available* axis reading, not the first ever. `TemporalBuffer` falls back to ring-buffer eviction when memory-capped even at `required_timespan=inf`, so on long runs the cutoff can advance and points can leave the histogram. This already happens today through the silent NaN path, so it is not a regression, and scalar logdata buffers should far outlast the primary's history in practice.
- Non-monotonic primary timestamps now raise instead of producing something. Label-based slicing requires a sorted time coord, where previously only the axis coord had to be sorted (`sc.lookup`'s requirement). Out-of-order results would be our own fault, so failing loudly seemed right, but it does turn a cosmetic anomaly into an ERROR-state plot.

## Test plan

Regression test follows the issue's hint: compute once with a history entirely preceding the axis data, then again with that history plus overlapping data, asserting the early points are consistently absent rather than shown and then silently removed. Tests for `_make_lookup` are replaced by behaviour tests through `compute()`, retaining the datetime64 and variance coverage.

- [ ] Correlation histogram in the live dashboard fills in as usual once the axis timeseries and the data overlap.

### DIFF
--- a/src/ess/livedata/dashboard/correlation_plotter.py
+++ b/src/ess/livedata/dashboard/correlation_plotter.py
@@ -147,24 +147,6 @@ CORRELATION_HISTOGRAM_PLOTTERS = frozenset(
 )
 
 
-def _make_lookup(axis_data: sc.DataArray, data_max_time: sc.Variable) -> sc.bins.Lookup:
-    """Create lookup table with appropriate mode based on time overlap.
-
-    Uses 'previous' mode normally, but falls back to 'nearest' if all data
-    timestamps are before the first axis timestamp (which would produce NaNs).
-    """
-    axis_values = sc.values(axis_data) if axis_data.variances is not None else axis_data
-    axis_min_time = axis_data.coords['time'].min()
-    mode = 'nearest' if data_max_time < axis_min_time else 'previous'
-    if mode == 'nearest':  # scipp>=25.11.0 rejects int coords in this mode
-        dim = axis_values.dim
-        coord = axis_values.coords[dim]
-        # Only convert to float64 if needed; datetime64 coords are accepted as-is
-        if coord.dtype != sc.DType.datetime64:
-            axis_values = axis_values.assign_coords({dim: coord.astype('float64')})
-    return sc.lookup(axis_values, mode=mode)
-
-
 @dataclass(frozen=True)
 class AxisSpec:
     """Specification for a correlation axis."""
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

SimonHeybrock (commented):
LGTM

PR #1135, opened 2026-07-27
Enable cumulative history plots: stamp every output with a 'time' coord

### LINKED ISSUE #1058: Drop redundant scalar 'time' coord; declare temporal role in output spec instead of magic coords

SimonHeybrock:
## User-facing changes

As a nice side-benefit, this enables plotting cumulative timeseries, showing how counts accumulate over time, and how/when resets affect this:

image

## Details

Closes #1058. Builds on #1132, which landed the `Temporality` declaration this relies on.

Four commits, best reviewed in order. The first makes a cumulative output plottable as history, the third lets a user ask for it, and the fourth is review follow-up.

## Stamp the instant a value refers to as `time`

A non-series output carried scalar `start_time` + `end_time`. That name is accurate for a window output, whose interval genuinely closed, but a lie for a cumulative one: the job is still running, and the bound really means "as of" — the observation watermark. The concept common to all three temporalities is *the instant the value refers to*, which is exactly what `time` already means for a series output. So the upper bound is now `time` everywhere, `start_time` keeps its name and its "since when" meaning, and `Temporality` says what the pair means per flavor rather than which coords each flavor happens to carry.

Every output now carries `time`, at unchanged wire cost — two timestamps either way, since window outputs no longer pay for a third that only duplicated `start_time`. A cumulative output becomes plottable as a history purely because it now carries the coord `TemporalBuffer` has always keyed on, so the buffer itself is untouched. `start_time` stays pinned for the lifetime of a generation, which NICOS depends on.

The dashboard keeps spelling the pair `(start_time, end_time)`: there it is provenance for the freshness/lag readout and rate normalization, not an axis. The extractors are the naming boundary, and all three now mint that pair uniformly — including `LatestValueExtractor`, which also fixes a pre-existing inconsistency where a window output reaching the plotters through a `SingleValueBuffer` was structurally different from the same output through a `TemporalBuffer`. `FullHistoryExtractor` captures the bounds before its timezone shift rewrites `time`, or they would misreport wall-clock time by the UTC offset. Nothing in `plots.py` changes.

`DeviceExtractor` adds `end_time` alongside `time` on the echoed device output. NICOS's consumer expects that name; adding rather than renaming decouples the published contract from our internal naming, so NICOS can migrate when it suits.

Three robustness fixes the rename makes reachable or worth pinning. `_add_time_coords` returned early when the job had no bounds, emitting outputs with no time coords at all — harmless while cumulative outputs were unbufferable, a `TemporalBuffer` rejection once they are not. The JobManager only finalizes jobs that accumulated primary data, so this is a scheduling fault, and it now raises rather than publishing untimestamped data for the dashboard to trip over far from the cause. A generation flip must clear an accumulated cumulative buffer, because a recommit may carry a changed config and the curve either side of it need not be the same quantity. A reset is the opposite case — the same accumulation restarting, whose history is worth keeping, since NICOS resets once per scan point and the resulting sawtooth is how that scan reads against wall-clock time. Nothing on the data path tells the two apart: `DataKey` strips the job number so the buffer is reused, and `TemporalBuffer` drops the accumulated coords before comparing metadata, so the changed `start_time` triggers nothing either. `clear_keys` on the generation flip is the only thing separating them. Both halves are pinned: the recommit clears, the reset does not. And the wall-clock x of a window output is now its `time`, the interval's right edge, where it used to be its left, so every existing window timeseries and correlation join shifts by one update interval — pinned as a convention rather than left incidental.

## Refuse to aggregate a time window over a cumulative field

Every message of a cumulative field holds the total accumulated since the run started, so summing consecutive ones counts the same events repeatedly. Fresh configs cannot reach this — the window controls are hidden for views without a per-update field — but `WorkflowSpec.field_for` falls back to the cumulative field, so a persisted config carrying a nonzero window duration, from a view that has since lost its per-update field, resolves to one. While a cumulative field carried no `time` coord this combination died in `TemporalBuffer.add`; now it buffers happily and returns a wrong number, which is worse, hence the guard.

Aggregating a `series` field is deliberately untouched: `nanmean` over non-uniformly spaced samples is #843's problem, not a temporality question, and blocking it would remove something users have today.

## Offer the cumulative flavor to the timeseries plotter

Until here, the growth curve was only reachable where a view has no per-update field for `field_for` to prefer. For the 53 zero-dim views backing both flavors — every `total_counts` and `total_in_range` across the instruments — a timeseries plot stayed pinned to the per-update field, with no control to say otherwise. The general plots' window controls do not transfer: a `FullHistoryExtractor` collapses the window bounds to the whole buffer, so duration and aggregation mean nothing on this plotter. It gets a single `Cumulative` checkbox instead.

Two spellings of one choice, so `WindowModeMixin` names what consumers actually want — the resulting `Windowing` — and the field-resolving, mode-validating and title-formatting sites read it through that one method rather than reaching for `time_window.mode`. What a view must back for a control to apply differs per spelling and is declared per params class: the duration control needs only a per-update field, the checkbox needs both, since with one missing one of its two states would either lie about the data shown or be rejected on submit. `PlotConfig.supports_windowing` is removed — written in three places, read in none, and unable to express that per-class rule anyway.

## Leave a series output's time axis alone at the extractor boundary

Review follow-up, correcting the first commit. Minting the pair from whatever coords the data happened to carry gave a series output an `end_time` taken from its last sample, and with it a freshness pill and a lag readout it did not have before. Both would then track the device's own sampling rate rather than the pipeline's latency, so a temperature logged every few minutes reads as permanently stale, and the "Lag" label would no longer mean what its docstring promises. Its age is already legible on the plot's x axis, which *is* that coord.

The pair is therefore minted whole or not at all, keyed on `start_time` — the coord only a bounded output carries. Every assertion the first commit had added to `extractors_test.py` was a series-shaped fixture, so that file is byte-identical to `main` again; the rule is pinned for both extractors instead.

Two smaller points from the same review. `LatestValueExtractor` drops its non-`DataArray` escape hatch and takes the `DataArray` the `DataService` is declared to hold — the branch was reachable only from tests publishing bare `Variable`s, i.e. it was absorbing a type violation rather than a real wire shape, and those tests now publish `DataArray`s. And `_build_subscription_keys` mints the `DataKey`s and their temporality in one pass, so `setup_pipeline` takes a single per-key mapping rather than a role-keyed one it had to re-zip against the keys.

## Deviations from the issue

#1058 is wrong in three places; a correction block is on the issue. The two that change the design:

- It proposes keying the buffer on `start_time` (or the window midpoint). Both are wrong for cumulative outputs: `start_time` stays pinned for the lifetime of a generation, which would stack every point of the growth curve on one x. The rename makes the question moot — the buffer keys on `time` as it always did.
- It proposes gating rate normalization on temporality. That would remove a working feature: on a cumulative field `_normalize_to_rate` divides by the run duration, giving the mean rate since the run started, which BIFROST's `detector_region_counts` view description already tells users to ask for. The genuinely wrong case — rate over a `FullHistoryExtractor`'s whole-buffer bounds — is unreachable, since the plotters requiring that extractor expose no rate control.

## Verification

Compatible plotters per output view are byte-identical before and after the template changes, dumped across all 157 views of all instruments.

Every output of every `dummy` workflow was driven through the real fake backend into a real `TemporalBuffer` and out through the extractors: 33 window, 48 cumulative and 3 series messages into 28 buffers, no rejections, every window and cumulative buffer yielding time bounds after extraction (series buffers deliberately yield none — see the fourth commit), and all 23 counts outputs normalizing to counts/s.

Driven in a fake-backend dashboard with a config covering all three temporalities through `TemporalBuffer`: a cumulative-only 0-D view (`dummy/total_counts/1`, the case that used to raise), the same plotter over a view that does have a window field, a 5 s window aggregation, and a series output.

The `Accumulation` group is present on a both-flavors view (`monitor_histogram/total_counts`) and absent on a cumulative-only one (`dummy/total_counts`), checked in the same fake-backend dashboard. That check cannot show the two subscriptions differing, since the fake backend synthesizes every 0-D field from the same wiggle regardless of temporality; the resolved field name is asserted in `plot_orchestrator_test.py` instead.

## Test plan

- [x] History plot of a cumulative 0-D output renders a growth curve against wall-clock time
- [x] The timeseries `Cumulative` checkbox switches a `total_counts` plot between the fluctuating per-update value and the monotonic total
- [x] Existing window-output plots (detector image, monitor histogram, totals) still show lag/freshness and rate normalization — worth re-running, since the bounds these read are now minted at the extractor boundary rather than carried on the wire
- [x] Timeseries plots of device readings are unaffected
- [x] Correlation plots still align sensibly — their wall-clock axis for a window output is now its `time` (the old `end_time`) rather than its `start_time`

### DIFF
--- a/docs/developer/adr/0006-nicos-derived-devices.md
+++ b/docs/developer/adr/0006-nicos-derived-devices.md
@@ -1,6 +1,6 @@
 # ADR 0006: Expose designated workflow outputs to NICOS as derived devices
 
-- Status: accepted
+- Status: accepted (amended 2026-07-29)
 - Deciders: Simon
 - Date: 2026-06-29
 
@@ -45,7 +45,8 @@ device.
   is not part of the external identity.
 - **Wire schema**: `da00`. It carries the signal, its error bar, and the
   `start_time` coordinate as named variables. `f144` is a lone scalar and cannot
-  carry the marker; it is not used.
+  carry the marker; it is not used. (See the 2026-07-29 amendment for the upper
+  time bound this also carries.)
 - **Eligibility**: decided by the workflow registry (`WorkflowSpec.device_outputs`).
   The extractor trusts the contract and emits whatever it designates; there is no
   runtime scalar/cumulative check. Devices are scalar cumulative outputs by
@@ -240,3 +241,20 @@ job-level generation marker is a reliable signal. That notification is deferred.
   heartbeat; `ep01` is not used) and a separate validity channel (the `da00` error
   bar carries the uncertainty; NICOS scripts its own statistic). The internal
   `ResultKey` / `job_number` refactor is off this critical path.
+
+## Amendment 2026-07-29: the upper time bound is echoed as `end_time`
+
+The wire schema above names only `start_time`, but the device topic has always
+also carried the upper bound of the accumulation: `Job._add_time_coords` stamps
+it next to `start_time`, and `DeviceExtractor` republishes the output verbatim.
+
+That bound is now called `time` internally. It is the instant the value refers
+to, which for a cumulative device is when the total was observed, not when
+anything ended — `end_time` was a poor name for a still-running accumulation, and
+that is why it was renamed. It is, however, the name NICOS's consumer reads.
+
+`DeviceExtractor` therefore *adds* `end_time` alongside `time` rather than
+renaming it, so the device topic publishes `signal`, `start_time`, `end_time`,
+and now `time` as well. NICOS's contract is unchanged and it can adopt `time` on
+its own schedule; the decision above — `start_time` as the sole generation
+marker — is untouched.
--- a/src/ess/livedata/config/instruments/bifrost/factories.py
+++ b/src/ess/livedata/config/instruments/bifrost/factories.py
@@ -347,8 +347,8 @@ def _create_base_reduction_workflow():
         """Sum detector counts in the selected arc and pixel range for one update.
 
         Emits bare per-update counts with no time coords: ``StreamProcessorWorkflow``
-        stamps ``start_time``/``end_time`` (needed for the dashboard's rate
-        normalization) on the accumulated outputs, per the window/cumulative
+        stamps ``start_time``/``time`` (the bounds the dashboard's rate
+        normalization needs) on the accumulated outputs, per the window/cumulative
         convention. Stamping here instead would suppress that (see
         ``_add_time_coords``) and break coord matching across the accumulator sum.
         """
--- a/src/ess/livedata/config/instruments/bifrost/specs.py
+++ b/src/ess/livedata/config/instruments/bifrost/specs.py
@@ -117,10 +117,7 @@ class DetectorRatemeterOutputs(WorkflowOutputsBase):
     )
 
     detector_region_counts: WindowOutput = pydantic.Field(
-        default_factory=lambda: sc.DataArray(
-            sc.scalar(0, unit='counts'),
-            coords={'time': sc.scalar(0, unit='ns')},
-        ),
+        default_factory=lambda: sc.DataArray(sc.scalar(0, unit='counts')),
         title='Detector Region Counts',
         description=(
             'Counts for the selected arc and pixel range, for the latest update '
--- a/src/ess/livedata/config/workflow_spec.py
+++ b/src/ess/livedata/config/workflow_spec.py
@@ -55,22 +55,30 @@ class Temporality(enum.Enum):
     operations over successive messages are valid. The two are not
     interchangeable — ``since_start`` is the default binding for fields that are
     not accumulations at all (ROI readbacks, static views).
+
+    Every emitted DataArray carries a ``time`` coord: the instant its value
+    refers to. What differs per temporality is what that instant means and what
+    accompanies it, and this is the one place that says so. Code holding data
+    but not the declaration sniffs for those coords; it is answering this
+    question.
     """
 
     window = 'window'
-    """Covers ``[start_time, end_time)``; successive windows are disjoint and
-    advance. Summing or averaging over consecutive messages is meaningful, as is
-    normalizing counts by the window duration."""
+    """Covers ``[start_time, time]``, where ``time`` is when the window closed.
+    Successive windows are disjoint and advance. Summing or averaging over
+    consecutive messages is meaningful, as is normalizing counts by the window
+    duration."""
 
     cumulative = 'cumulative'
-    """Value as of ``end_time``, accumulated since ``start_time``, which stays
-    pinned for the lifetime of a job generation. Plotting the growth curve is
-    meaningful; aggregating over successive messages double-counts."""
+    """Value as observed at ``time``, accumulated since ``start_time``, which
+    stays pinned for the lifetime of a job generation. Plotting the growth curve
+    against ``time`` is meaningful; aggregating over successive messages
+    double-counts."""
 
     series = 'series'
-    """Carries its own per-point ``time`` axis. Messages are disjoint chunks of
-    samples concatenated along that axis, so there is no single window duration
-    to normalize by."""
+    """``time`` is the per-sample axis rather than a bound, and there is no
+    ``start_time``. Messages are disjoint chunks of samples concatenated along
+    that axis, so there is no single window duration to normalize by."""
 
 
 WindowOutput = Annotated[sc.DataArray, Temporality.window]
--- a/src/ess/livedata/core/job.py
+++ b/src/ess/livedata/core/job.py
@@ -213,39 +213,45 @@ def _add_time_coords(
     data: sc.DataGroup, start_time: Timestamp | None, end_time: Timestamp | None
 ) -> sc.DataGroup:
     """
-    Add start_time and end_time as 0-D coordinates to all DataArrays in a DataGroup.
+    Add start_time and time as 0-D coordinates to all DataArrays in a DataGroup.
 
-    These coordinates provide temporal provenance for each output, enabling lag
-    calculation in the dashboard (lag = current_time - end_time).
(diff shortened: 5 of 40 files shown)


SimonHeybrock:
LGTM

PR #1136, opened 2026-07-27
Fix plot axes that collapse on a degenerate data range

SimonHeybrock:
A 1D correlation histogram whose axis values are all identical renders with its x axis collapsed to a point — every tick label stacked on the same pixel — and no visible bars. Found while manually testing #1134; it reproduces unchanged on `main`. The trigger is not exotic: a device parked at a setpoint, or an axis timeseries that has produced a single reading so far while the data updates faster, both correlate every point onto one value.

Two faults meet there, and only one of them is specific to correlation histograms.

`hist` derives bin edges from the value range and does not guard a degenerate one: for a constant axis it emits N bins of zero width, bumping only the top edge by one ULP to keep the range technically valid (reported upstream as scipp/scipp#3935). Zero-width bars stay invisible however wide the axis around them, so this has to be fixed where the edges are made, and the correlation plotter is the only place in the dashboard that derives edges from data rather than from workflow config. Patching the edges after binning would not work: scipp puts the counts in bin index 1 whatever the bin count, so relabelling the edges over a wider interval would leave the bar 1% from the left of an axis centred on the value.

The zero-width guard behind every plotter's autoscale range then has two holes of its own, neither of which needs a correlation histogram to reach. It tests for an exactly zero span, which the ULP bump defeats. And it widens a degenerate range by 5% of its magnitude, which presumes the axis is measured from its zero — a datetime axis is measured from the epoch, so a timeseries plot showing a single sample was framed by an x axis spanning roughly 2000 days. A single sample is a normal state at startup, and it is also what any set of samples within ~256 ns collapses to, that being the float64 spacing at a 2020s epoch. Degeneracy is now measured against the float64 spacing at the range's own magnitude, which is the right resolution both for a temperature in kelvin and for an epoch in nanoseconds, and datetime axes are widened by a fixed interval of one second.

The correlation plotter takes both the test and the widening from that shared guard rather than restating them.

One deliberate narrowing: the constant-axis test is a few ULP rather than a relative tolerance, so an axis jittering in its last few digits now gets genuine narrow bins instead of a widened range. Whether jitter that small is worth binning over is a display policy, and it belongs in the range layer if we want it, not in the binning.

## Test plan

Regression tests assert that a constant correlation axis produces strictly increasing edges and retains all counts, that an axis constant at zero (no magnitude to widen relative to) also spans a visible range, and that a varying axis still takes its edges from the data. On the shared guard, a line plot gets a visible x range for a constant axis and for one spanning a single ULP, and a single-sample datetime axis is widened by seconds rather than years. The constant correlation axis, the ULP case and the datetime case all fail on `main`.

- [x] A correlation histogram against a parked device shows a bar and a readable x axis.
- [x] A timeseries plot showing its first sample has a readable x axis.

### DIFF
--- a/src/ess/livedata/dashboard/correlation_plotter.py
+++ b/src/ess/livedata/dashboard/correlation_plotter.py
@@ -29,7 +29,15 @@ from .plot_params import (
     PlotDisplayParams1d,
     PlotDisplayParams2d,
 )
-from .plots import ImagePlotter, LinePlotter, PresenterBase, TimeBounds, TitleResolver
+from .plots import (
+    ImagePlotter,
+    LinePlotter,
+    PresenterBase,
+    TimeBounds,
+    TitleResolver,
+    ensure_span,
+    is_degenerate_span,
+)
 from .range_hook import Axis, RangeTargets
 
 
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1137, opened 2026-07-27
Add --collapsed-sidebar, used by the automation launcher

SimonHeybrock:
Adds a `--collapsed-sidebar` flag to the dashboard and passes it from the Playwright driving kit's `--launch`.

The sidebar carries announcements and the version label. Automation runs disable announcements, so the drawer is static there — it only narrows the plots under test and the screenshots taken of them, and eats a chunk of every image handed to a reviewer or a model. The flag sets `MaterialTemplate.collapsed_sidebar`, so the page arrives collapsed on the first frame: no hamburger click, no resize while Bokeh models are still settling, and it survives a page reload. Plots go from 1230 px to the full 1600 px of the automation viewport.

Not automation-only: a small screen is as good a reason to want the extra plot width, hence a plain flag rather than something coupled to `--transport fake`.

The alternative was collapsing browser-side by clicking the hamburger, which also covers dashboards the kit did not launch itself. Rejected: it hangs the automation off Panel's Material markup (`#sidebar.mdc-drawer--open`, `button.mdc-top-app-bar__navigation-icon`), which is exactly the kind of selector that breaks on a Panel upgrade, and there is no current need to drive a hand-run server this way.

## Test plan

- `pytest -m browser tests/dashboard/` — 5 passed.
- `pytest tests/dashboard` — 1784 passed.
- `python scripts/drive_dashboard.py --launch --tab Detectors --screenshot out.png` — sidebar absent, grid spans the full width.
- `--check --collapsed-sidebar` constructs the app.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -87,7 +87,7 @@ it, on a port other than 5009 (interactive dev uses 5009):
 ```sh
 cp -r tests/dashboard/ui_config_fixtures/dummy "$TMP/cfg/dummy"
 python -m ess.livedata.dashboard.reduction --instrument dummy --transport fake \
-    --port 5011 --config-dir "$TMP/cfg" --no-fetch-announcements
+    --port 5011 --config-dir "$TMP/cfg" --no-fetch-announcements --collapsed-sidebar
 ```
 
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1138, opened 2026-07-28
Add integration test for the stop path halting backend publishing

SimonHeybrock:
## Summary

Adds an integration test for the "stop path" checkbox in issue #1116: a commanded stop must actually halt backend work, not just flip the status shown in the dashboard.

`JobOrchestrator.stop_workflow` also deactivates the workflow on the dashboard side, so `DataService` would filter out late results even if the backend worker kept churning after the stop. A dashboard-only assertion (e.g. on job status) would therefore pass vacuously whether or not the backend actually stopped. The test instead asserts directly on Kafka: after starting a monitor workflow it confirms the results topic's (`livedata_data`) high watermark is genuinely advancing, issues the stop via `job_orchestrator.stop_workflow` (the same call the dashboard's stop button makes), and then asserts the watermark stalls and stays stalled. The raw monitor topic keeps advancing throughout since `fake_monitors` publishes independently of any workflow, so the test is careful to only assert on the results topic.

Refs #1116.

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -114,6 +114,41 @@ def wait_for_watermark_stall(
         previous = current
 
 
+def wait_for_watermark_advance(topic: str, *, since: int, timeout: float = 30.0) -> int:
+    """Wait until a topic's high watermark exceeds ``since``, then return it.
+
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1139, opened 2026-07-28
Integration test: rapid recommit under a live producer

SimonHeybrock:
Covers the "Rapid recommit under a live producer" integration checkbox of #1116, with one deliberate reduction in scope explained below.

A commit flips the dashboard's generation and clears the workflow's buffers before its commands reach the backend, so results computed for a superseded generation can still arrive afterwards. The test recommits a running workflow several times back-to-back against a live producer and asserts that the data reaching `DataService` carries the final commit's `job_number` alone.

## Why drop counting is not asserted here

The checkbox also asks that stale-generation drops be counted. That half is not assertable at this layer, and the investigation is the more useful outcome of this PR.

Message forwarding is driven by a background thread pumping every 50 ms, independently of the explicit `update()` calls the integration wait helpers make. A message therefore cannot be held unforwarded until a chosen moment, which rules out making the assertion deterministic by controlling when forwarding happens.

The drops are also systematically absent in this scenario rather than merely hard to catch. A stop reaches the backend in roughly 100 ms while a monitor workflow publishes about once a second, so a flip usually finds nothing of the old generation in flight. A generation superseded before the backend ever starts it publishes nothing at all — which is precisely what a rapid burst produces. And every flip resets the counter, so only the final commit's window could ever be asserted. Two CI runs counted zero drops, matching `begin_generation`'s own independently logged tally.

Worth recording against #1116: the checkbox's wording assumes in-flight superseded results that this pipeline does not normally produce. Counting is covered deterministically by the `message_pump` unit test instead, which constructs the superseded result directly and now asserts the count rather than only the discard.

To head off a plausible-sounding wrong explanation: data admission consults `is_current`, a single-entry check, so the immediately preceding generation is dropped like any other. The two-entry window governs heartbeat adoption only, and is not what absorbs these messages.

## Production change

`ActiveJobRegistry.stale_count` is added as a public read of the drop counter, which previously had no read path outside its log line. That left the counting behaviour unassertable — the existing unit test could only check that recording a drop has no side effects. It is now covered directly.

## Test plan

- [x] Unit coverage for the accessor (per-workflow scoping, reset at generation flip, unobserved workflow) and for the drop being counted.
- [ ] Integration test passes against real Kafka.

### DIFF
--- a/src/ess/livedata/dashboard/active_job_registry.py
+++ b/src/ess/livedata/dashboard/active_job_registry.py
@@ -194,6 +194,18 @@ class ActiveJobRegistry:
                     return gen.config
             return None
 
+    def stale_count(self, workflow_id: WorkflowId) -> int:
+        """Number of data messages dropped since the current generation began.
+
+        Every :py:meth:`begin_generation` resets the counter, so the value is
+        scoped to the workflow's current generation: it measures how much
+        in-flight data the latest commit superseded. Zero for a workflow with
+        no generation record.
+        """
+        with self._lock:
+            record = self._generations.get(workflow_id)
+            return 0 if record is None else record.stale_count
+
     def record_stale(self, workflow_id: WorkflowId, job_number: JobNumber) -> None:
         """Count a dropped data message from a non-current generation.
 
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1140, opened 2026-07-28
Poison-message liveness over real Kafka

SimonHeybrock:
Adds the end-to-end poison-message liveness test from #1116 (third integration checkbox: "publish a malformed `ev44` into the detector/monitor topic mid-job; the service keeps processing subsequent good data").

The claim that a malformed payload is contained per-message is currently tested at the adapter level and at the service level in-process, where payloads are handed straight to the service loop. Over real Kafka the path also contains a real consumer, schema routing off a topic shared by many sources, and process-level containment — none of which the in-process harness can exercise.

Liveness is asserted by ordering rather than by timing. The fake producer is restricted to `monitor1`, which leaves `monitor2` — a configured monitor source of the dummy instrument — with no producer other than the test. Poison and probe messages reach the same single-partition topic in that order, so a workflow result for `monitor2` can only exist if the service consumed past the poison. Watching `monitor1` keep producing instead would not distinguish "survived the poison" from "has not reached the poison yet", which is the failure mode that makes this kind of test rot into a vacuous pass.

Injected are the malformed entries of the shared corpus in `tests/helpers/hostile_wire.py` (garbage bytes, empty payload, truncated `ev44`, `ev44` without event vectors, wrong schema on the monitor topic, unmapped source name). The well-formed-but-insane timestamps stay out: they are a batcher concern, covered in-process where the clock is controllable, and over real Kafka they would only add wall-clock-dependent silence windows to wait out.

The service combination needed here (fake producer restricted to one monitor) does not exist among the shared fixtures, so the group-creation helper in the integration `conftest.py` is now public and the `services` marker's extension point is documented.

## Test plan

- [ ] `tox -e integration` against a local Kafka

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -6,7 +6,7 @@ Integration tests verify end-to-end behavior of ESSlivedata by running actual se
 
 - **`@pytest.mark.integration`**: Marks a test as an integration test
 - **`@pytest.mark.instrument('name')`**: Runs test with specified instrument (default: 'dummy')
-- **`@pytest.mark.services('name')`**: **Required** when using `integration_env` fixture. Valid values: `'monitor'`, `'detector'`, or `'reduction'` (specifies which services to run)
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

Issue #1133, opened 2026-07-27
Device clock skew: a creeping device silently becomes the service's data clock


SimonHeybrock:
## Context

Production devices creep by O(1 s/day) against the shared pulse clock. The batchers survive this — liveness under creep is pinned by `TestClockCreep` (#1122): delivery flows through every threshold crossing, and a months-accumulated offset with intermittent traffic does not ping-pong. But probing the equilibria showed quality degradations that are operationally invisible:

- **The window tracks the fastest clock.** The timeout path follows the high-water mark, a bare max over all streams, so once a device runs ahead by more than the timeout margin (~0.2 s, i.e. after ~half a day of creep) it owns window placement. In a 900 s simulation at accelerated creep, batch `end_time` ended 4.1 s ahead of true time; with a static 240 s offset the window settled permanently on the skewed clock. Restart does not help — bootstrap re-anchors on the majority, then the window walks back within seconds.
- **Slot-gated closure silently degenerates to the timeout path** at the same threshold, losing the pulse-slot alignment the rate-aware batcher was built for (#875): the healthy stream's messages landed inside their batch's time range only 16.7 % of the time (97.8 % in the control). Batch contents straddle boundaries again.
- **The data clock skews with the device.** Batch end times drive job activation and run resets (`JobManager._advance_to_time`), so an ahead-creeping device fires them early by its accumulated offset — minutes, once creep has run for months.

Behind-creep is the benign mirror: the window stays with the healthy stream, only the laggard's data lands outside batch ranges, and at >1000 batches behind it is demoted from gating.

## Directions

Window placement deliberately has no wall reference (clock policy in `rate_aware_batcher.py`), so the batcher cannot tell a creeping clock from a true one. In rough order of value:

- **Detection at the adapter boundary** — arrival time is where skew is visible. The #1047 validation work is the natural home: track per-source offset between data timestamps and arrival wall time, surface creep as telemetry/warnings long before it exceeds any bound.
- **Cheap batcher-side observability** — warn (rate-limited) when a gating stream is persistently unsatisfied at close, or when per-stream frontiers disagree by more than the plausibility horizon. Catches "gating has silently stopped mattering" in ops regardless of cause.
[...]

PR #1141, opened 2026-07-28
Re-latch a job start time contradicted by its own data

SimonHeybrock:
## Why

`Job._start_time` is sticky by design: it records when the job began accumulating and is never lowered. That makes it a permanent amplifier for a transient bad timestamp. A batch built from a single far-future message anchors on that message (`plausible_anchor` has nothing to weigh a lone timestamp against), so a job whose first active batch is such a batch keeps that start time for its lifetime.

Every later result then carries `start_time > end_time`, with two consequences that are silent in operation: the dashboard's rate normalization treats a non-positive duration as "cannot normalize" and returns raw counts instead of counts/s with no warning, and job adoption ranks record-less observed jobs by start time, so the poisoned job outranks every legitimately newer one.

## What changed

A start time later than the newest accumulated data is contradicted by the data itself — a job cannot begin after its accumulation ends. That is sufficient evidence to correct the latch in band: no wall clock, no plausibility bound, no new layer. The job re-latches on the current batch and logs it.

The condition is conservative on purpose. Batch end times legitimately move backwards within a job whenever a batcher recovers a misplaced window, but they do not drop below the job's own start unless that start is wrong.

Bad timestamps still produce wrong times on the batches that carry them. What changes is that the damage no longer outlives them.

## Relation to #1059

#1059 attacked the same class of failure from the other end, by clamping far-future data-derived timestamps at the Kafka adapter boundary. It was written against an earlier revision of #1122; re-examined against `main` now that #1122 has landed with a wall-clock liveness backstop and the rate-aware batcher as the default, most of its motivation is gone:

- Liveness no longer depends on it. The backstop bounds a poisoned window at roughly ten batch lengths and costs one stall per poison message; the remaining hole is specific to `SimpleMessageBatcher`, which is now opt-in.
- The data-clock damage it was sized against is currently unreachable. `_fire_pending_resets` only ever sees an empty list, because `with_run_control_route` is disabled for #795, and no `JobSchedule` carrying an `end_time` is constructed anywhere in `src/`.
- Clamping in the dashboard-side adapters cannot do anything: `MessagePump` forwards only `stream` and `value` and discards the message timestamp.

What remained live was this latch, which is in-band and needs no arrival-time evidence. #1059 is closed in favour of it. The parts of #1059 that stand on their own merits are separable and unaffected: per-source clock-offset telemetry belongs with #1133 (and should extend the existing lag reporting rather than add a parallel counter), and the ev44 absent-event-vector handling is #1124.

Note this does not address `ToNXlog._last_time`, which latches on the *payload* timestamp and rejects every later sample for the process lifetime. That is a separate defect on the payload side and no envelope-level fix reaches it.

Also here: the six adapter-boundary far-future cases were strict xfails asserting that such a timestamp must be clamped, dropped or replaced before crossing the boundary — #1059's acceptance criterion. That remedy is now rejected rather than merely unimplemented, and a strict xfail encoding a rejected design is a standing invitation to implement it. `_record_lag` measures a producer by the gap between the payload's claimed time and the broker's create time, so a clamp upstream of it would report a device with a badly wrong clock as perfectly punctual — erasing exactly the evidence #1133 needs. The assertion is inverted to pin the pass-through instead, keeping all six timestamp entry points covered.

## Test plan

Unit tests for both directions of the re-latch condition, and a service-level test on the hostile-input harness asserting that published result timestamps return to the real timeline after a lone far-future message. `SimpleMessageBatcher` wedges outright on that input, so no later batch can contradict the poisoned start; that case is pinned as a strict xfail beside the existing one for the same hole.

### DIFF
--- a/src/ess/livedata/core/job.py
+++ b/src/ess/livedata/core/job.py
@@ -9,11 +9,14 @@ from enum import StrEnum, auto
 from typing import Any, Protocol, runtime_checkable
 
 import scipp as sc
+import structlog
 
 from ess.livedata.workflows.workflow_factory import Workflow
 
 from ..config.workflow_spec import JobId, ResultKey, WorkflowId
 from .timestamp import Timestamp
+
+logger = structlog.get_logger(__name__)
 
 
 @dataclass(slots=True, kw_only=True)
@@ -377,6 +380,22 @@ class Job:
             )
             if data.is_active():
                 if self._start_time is None:
+                    self._start_time = data.start_time
+                elif data.end_time < self._start_time:
+                    # A job cannot begin after its accumulation ends, so the
+                    # latched start is provably wrong: it came from a batch
+                    # anchored on a timestamp outside this stream's timeline
+                    # (plausible_anchor has nothing to weigh a lone message
+                    # against). The start is sticky, so without this it would
+                    # pin that value -- and an inconsistent time pair on every
+                    # result -- for the job's lifetime.
+                    logger.warning(
+                        'job_start_time_relatched',
+                        job_id=str(self._job_id),
+                        previous_start_time=str(self._start_time),
+                        start_time=str(data.start_time),
+                        end_time=str(data.end_time),
+                    )
                     self._start_time = data.start_time
                 self._end_time = data.end_time
             return JobReply(job_id=self._job_id)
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1142, opened 2026-07-28
Loop ROI requests back into the fake backend

SimonHeybrock:
Implements the infrastructure enabler from #1116.

`FakeBackendTransport` wired its `roi_sink` to a null sink, so ROIs drawn in the UI were discarded. Nothing downstream could respond: `roi_spectra` outputs were synthesized by the generic template expansion, which treats the empty `roi` dimension as unsized and stretched it to a fixed 64 rows of data unrelated to any ROI the user had drawn. The geometry readbacks were worse off — the same expansion stretched their `bounds`/`vertex` dimension and dropped the coordinates the readback plotter requires, so the ROI overlay could not render at all. Both are the same bug, and both are fixed here: ROI requests now loop back into the fake backend, which stores the latest request per job and geometry, echoes it as the readback, and synthesizes one spectrum per drawn ROI whose values depend on that ROI's own coordinates.

The synthesized row count follows the published ROI set, down to zero rows while nothing is drawn. That matches production rather than special-casing it: ROI requests are context keys seeded to `None`, the providers tolerate `None`, and the real `roi_spectra` returns an empty result from the first batch onward. Detection of ROI-shaped outputs is structural rather than name-based, so it does not need a list of output names kept in sync.

Tests drive the real `ROIPublisher` through the real transport and command service rather than stubbing the path, covering the ROI dimension tracking additions, removals and clears, index preservation, rectangles and polygons sharing the dimension, geometry-dependent values, readback round-tripping, and requests addressed to a job that is not running.

No browser-level ROI test is included. Driving the box-edit tool would be the first canvas interaction in the suite, and it is not committable as non-flaky: there is no automation hook on the figure or toolbar, tool activation is client-side so there is nothing server-side to await, and pixel targets depend on axis ranges that autoscale can move mid-gesture. A deterministic widget-driven alternative exists — the ROI-request plotter publishes its initial set from a text parameter — but it needs an ROI layer added to the shared, drift-guarded dummy UI fixture, which is regression risk disproportionate to this change. The loopback is what unblocks that work either way.

## Test plan

- [x] Fast suite
- [x] Existing browser suite, to confirm the changed synthesis does not disturb rendering

### DIFF
--- a/src/ess/livedata/dashboard/fake_backend.py
+++ b/src/ess/livedata/dashboard/fake_backend.py
@@ -13,6 +13,10 @@ Result data is synthesized from each workflow's output templates (the
 the dashboard selects plotters from these same templates, generated data is
 guaranteed to match what the plotter expects, so plots render correctly.
 
+ROI requests the dashboard publishes are looped back in the same way: the
+backend stores the latest request per job and geometry, echoes it as the
+geometry readback, and synthesizes one spectrum per drawn ROI.
+
 The whole flow is driven by the normal UI: start a workflow and plots come to
 life. No fixtures, no external injection, no Kafka.
 """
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1143, opened 2026-07-28
Guard grid-edit Save against a concurrently vanished grid

SimonHeybrock:
Implements the concurrent-edits browser check from #1116, and fixes the crash it uncovered.

Probing the scenario before writing the test showed it did not behave as the issue assumed. Two sessions editing the same grid's properties and saving back-to-back raised an unhandled `KeyError` server-side, reproducibly: the Save Changes click is queued client-side before the losing session's next topology poll, so its handler read a grid id that had already been invalidated and passed it into an unguarded orchestrator lookup. The losing edit vanished with a traceback in the server log and no feedback in the browser. The fix re-checks that the grid still exists immediately before applying the change — the same check the topology poll already performs — and surfaces a notification instead of crashing, matching the wording this codebase already uses for other "removed mid-action" cases.

Worth recording for the issue: the resulting behaviour is not last-writer-wins as #1116 describes it. Whichever save commits while the grid still exists wins, and in a tight race that is closer to first-commit-wins, because the first save landing is precisely what makes the second session's target disappear. The test therefore asserts the invariants that always hold — exactly one of the two submitted titles survives, no duplicate or lost tab, the losing session is told why, no server-side exception, plots still updating — rather than pinning which session wins.

Coverage is deliberately split across two layers. The browser test drives the real race through two sessions, which is the only place the client-side click ordering exists at all; two fast unit tests pin the guard directly so the stale and `None` grid-id paths do not depend on a slow browser run to stay fixed. Both were confirmed to fail with the exact `KeyError` when the guard is reverted.

Asserting that no exception reached the server required the browser harness to expose its log alongside its URL, so its callers now unpack the URL from the returned value. The interactive `--map` and `--screenshot` paths were verified unaffected.

## Test plan

- [x] New browser test run repeatedly to check for flake
- [x] Full browser suite
- [x] Fast suite

### DIFF
--- a/scripts/drive_dashboard.py
+++ b/scripts/drive_dashboard.py
@@ -48,6 +48,7 @@ import tempfile
 import time
 import urllib.request
 from contextlib import contextmanager
+from dataclasses import dataclass
 from pathlib import Path
 from urllib.error import URLError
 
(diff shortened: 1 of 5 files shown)


SimonHeybrock (commented):
LGTM

PR #1144, opened 2026-07-28
Cover session churn and reaper teardown with a browser test

SimonHeybrock:
Implements the session-churn browser check from #1116: churn sessions through a real browser and check that the server neither leaks them nor stops working. Two teardown paths release per-session state — the clean websocket close, and the reaper that catches a browser which vanished without one (#955, #1095) — and neither had ever run end to end through a browser.

The observable is the server's own account of session lifecycle. `SessionRegistry` names every session it registers, unregisters and reaps, so the test follows session ids through the server log: ids registered while browsers churn must come back as *unregistered*, ids registered by a browser that then went offline must come back as *reaped*, and neither set may appear under the other's teardown line.

Naming the teardown path is what keeps the two apart. A session count can only distinguish them by *when* it drops, which pins the test to Bokeh's unused-session defaults (15 s lifetime, swept every 17 s) that this app does not set — an upstream default change would silently collapse the two paths into one with the test still green. A count is also less literal than it looks: the harness's own readiness probe renders the app over plain HTTP, registering a browser-less session that inflates any baseline.

What the log cannot show is whether the deeper teardown ran: a leaked viewer token keeps a hidden layer computing, and nothing logs it. That needs the app in-process, filed as #1147; until then it stays manual on #1097.

Two notes from building this, neither addressed here:

- Every session open logs an unhandled `ValueError: A callback of the same type has already been added with this ID` traceback, raised inside Bokeh's `with_document_locked` after `pn.state.add_periodic_callback`. Sessions work regardless, but it is error-level noise on every connect, and it rules out a blanket "no traceback in the log" assertion.
- The `Dropping a patch ... previously known reference` line that #1097 asks to watch for is `logging.debug` in Bokeh 3.9.1, and upstream calls it usually harmless. It cannot appear in a journal at info level, so that part of the manual checklist cannot be satisfied as written.

## Test plan

- [x] New browser test passes
- [x] Fails with `SessionRegistry.unregister` disabled (cleanly closed sessions reported unaccounted for, by id)
- [x] Fails with `cleanup_stale_sessions` disabled (abandoned sessions reported unaccounted for, by id)
- [x] Full browser suite (11 passed, with #1149 merged in)

### DIFF
--- a/tests/dashboard/session_churn_test.py
+++ b/tests/dashboard/session_churn_test.py
@@ -153,7 +153,7 @@ def assert_server_still_serves(browser: Browser, url: str) -> None:
 
 @pytest.mark.browser
 def test_session_churn_returns_to_baseline_and_server_stays_usable() -> None:
-    with fake_dashboard("dummy", 5070) as fake, open_browser() as browser:
+    with fake_dashboard("dummy") as fake, open_browser() as browser:
         churned_at = fake.log.stat().st_size
         for _ in range(CHURN_CYCLES):
             contexts = [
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM, but could be improved once we have #1147.

PR #1145, opened 2026-07-28
Address every cell titlebar tool per cell, not just the pencil

SimonHeybrock:
A cell titlebar holds three tools: the gear (a layer-picker menu when the cell has several layers), the pencil, and the layer-details toggle. All three already carried `lt-tool` plus a semantic `lt-tool-*` class, but only the pencil received the caller's `lt-cell-r{row}c{col}` context, so the other two were addressable only by ordinal across the whole grid. The repeated-instance rule in `.claude/rules/dashboard-widgets.md` asks for a context class precisely so that cannot happen, and the gear is the tool automation most wants, since it is what opens the layer config.

Ordinals are not an acceptable substitute here. Cell DOM order is deliberately unstable — a rebuilt cell moves to the end of the document — so `nth(i)` addresses a different cell after any rename, and driving several cells in one session needs a page reload between them just to keep the indices meaningful. I hit exactly that while verifying #1135: reaching each cell's gear took ordinals plus a reload per modal, and one modal still failed to open.

`create_cell_titlebar` now passes its `css_classes` to all three tools instead of to the pencil alone. Each stays distinguishable by its own `lt-tool-*` class, so `.lt-cell-r0c1.lt-tool-settings` names exactly one button. The two hand-rolled widgets (the toggling-icon button and the dropdown) append the context to the hooks they already tag themselves with.

These classes carry no style rules, so the change is visually inert.

**Second commit: a multi-layer cell in the dummy fixture.** Every fixture cell was single-layer, so the gear's layer-picker variant — and cell overlays generally — never appeared in a driven dashboard, leaving the `len(layers) > 1` branch verifiable only by unit test. The Detectors grid gains a third row holding two `lines` layers over the monitor histogram, one per monitor: a config users actually build, and the cheapest overlay the fake backend renders correctly. Existing cells keep their geometry, so `lt-cell-r0c0`/`lt-cell-r0c1` still address what they did and the fixture's tab list is unchanged — which the grid-reordering test asserts against.

## Verification

The existing pencil guard generalizes over the three tools and over both gear shapes (single-layer button and layer menu). It fails on all three tools before this change.

In a fake-backend dashboard, all nine compound selectors across the fixture's three cells resolve to exactly one element each, and `.lt-cell-r0c1.lt-tool-settings` opens the layer config modal — the click that previously had no per-cell selector at all.

A new browser test walks the picker on the multi-layer cell: both entries are listed, and choosing the second opens the config modal for *that* layer rather than the cell's first. It asserts on the source-selector chip, since the dialog's own `inner_text` is empty — Panel renders each widget into its own shadow root. The full browser suite passes (6 tests).

## Test plan

- [x] Cell titlebar renders unchanged (gear, pencil, toggle; classes carry no styles)
- [x] Gear opens layer config for the addressed cell, on a single-layer cell
- [x] Multi-layer cell: the gear is a dropdown, lists the layer titles, and opens the chosen one

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -60,10 +60,11 @@ no text, `title`, or `aria-label` — leaving nothing semantic for browser autom
 *name* identity to slug (`workflow_status_widget.py`'s `_tool_css_class`), but a grid has
 none, so `lt-grid-*` slugs the grid *title* instead (`plot_grid_manager.py`) — the same
 title that also drives the grid's download filename. Plot cells have neither a name nor
-a stable title, so their titlebar pencil carries the grid position as
-`lt-cell-r{row}c{col}` (`cell.py`) — unique per grid, and only the active tab's grid is
-rendered. Do not rely on DOM order for cells: a rebuilt cell (e.g. after a rename) moves
-to the end of the document.
+a stable title, so every button in their titlebar — gear (or layer menu), pencil, and
+layer-details toggle — carries the grid position as `lt-cell-r{row}c{col}` (`cell.py`),
+unique per grid, and only the active tab's grid is rendered. Address one with a compound
+selector: `.lt-cell-r0c1.lt-tool-settings`. Do not rely on DOM order for cells: a rebuilt
+cell (e.g. after a rename) moves to the end of the document.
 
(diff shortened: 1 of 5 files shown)


SimonHeybrock (commented):
LGTM

PR #1148, opened 2026-07-28
Size table columns from their content

### LINKED ISSUE #1146: Table plotter columns wider than necessary, resulting in horizontal scrollbars
In this example the columns are wider than they need be, and we get a horizontal scrollbar:
Image

The scrollbars disappear at browser zoom 75% or lower. Is our default to wide? Can we derive width from data better, or will that lead to unpredictable results?

SimonHeybrock:
Closes #1146.

Table columns were split 50/50 regardless of content: Bokeh's `DataTable` defaults to `autosize_mode='force_fit'`, which stretches every column to an equal share of the widget (HoloViews leaves all column widths at Bokeh's 300px default, so the ratio is always 1:1). Short source names and values ended up stranded in half-empty cells, with the right-aligned numbers pinned to the far edge of the widget instead of sitting next to the row they belong to.

Widths are now computed from the content: string columns from their longest entry, value columns from a character budget derived from the configured notation and precision. Scientific, auto and compact notation are bounded by their format; only decimal notation follows the data, and only in steps of an order of magnitude, so widths do not jitter as values update.

This also removes the reported horizontal scrollbar. It came from force-fit sizing the grid canvas to the widget width and then leaving it a few pixels stale after a resize or zoom change (reproduced in a browser: a zoom change alone flips the scrollbar on). With the canvas sized from the columns it no longer tracks the widget width, so the slack absorbs any rounding.

Bokeh's own content-fitting modes are not usable: `fit_columns` and `fit_viewport` raise `ReferenceError: row is not defined` in the SlickGrid bundled with bokeh 3.9.1 (`getColContentSize`, the `GetLongestText` branch) as soon as a column holds strings, which aborts the render and collapses the pane.

Because the plot is only ever driven from Python, the pixel-per-character estimate is deliberately generous, so a column is a little wider than it needs rather than truncating a cell to an ellipsis.

## Test plan

- Unit tests cover the width behaviour: content-derived widths, header floor, bounded vs. data-dependent notation, and `autosize_mode`.
- Verified in a browser (Playwright, classic scrollbars) against the real plotter: no truncated cells and no horizontal scrollbar across container widths 300-700px, fractional widths, and zoom 0.67-1.5, where the old code showed the scrollbar.
- [x] Check a real table plot in the dashboard (e.g. a total-counts view with several detector banks) looks right.

### DIFF
--- a/src/ess/livedata/dashboard/table_plotter.py
+++ b/src/ess/livedata/dashboard/table_plotter.py
@@ -19,30 +19,54 @@ from .plot_params import PlotParamsTable, TableNotation
 from .plots import Plotter, TitleResolver
 from .range_hook import Axis
 
+# Bokeh renders the table in 12px Helvetica/Arial. Seven pixels per character is
+# a deliberate overestimate of the average advance width there, so columns are
+# never so tight that Bokeh truncates a cell to an ellipsis.
+_CHAR_PX = 7
+_CELL_PADDING_PX = 12
+
 
 class _FormattedTablePlot(TablePlot):
-    """HoloViews bokeh ``Table`` plot whose value columns carry a custom formatter.
+    """HoloViews bokeh ``Table`` plot with custom column formatters and widths.
 
     HoloViews hardcodes the numeric column formatter and rebuilds every column on
     each streaming update (``update_frame``) without re-running ``hooks``, so a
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1149, opened 2026-07-28
Allocate browser-test dashboard ports from the OS

SimonHeybrock:
Browser tests each hand-picked a dashboard port literal (5032-5038). Nothing enforced that the set stayed disjoint, and the failure mode is silent: two checkouts running the suite at once, or a suite running alongside an interactive dashboard, land on the same port and one of the servers dies. Keeping the literals disjoint was also manual work every time a test landed on a parallel branch — #1143 picked 5037 by reading off what #1145 had already taken.

`fake_dashboard` now takes a port from the OS when it is not given one, which is how all the tests call it. `--launch` does the same, so parallel driving sessions no longer fight over 5011 either. Passing an explicit port keeps the existing in-use guard, for the case where the URL has to be known up front.

A port chosen this way is only reserved until our server binds it, so a racing process can still take it in between. That race is not closed — closing it would mean handing the listening socket to the Panel server — but it no longer hides: `_wait_until_ready` reports a server that exited before serving as soon as it exits, so this and any other startup failure surface immediately with the server's log tail instead of after a 60s wait.

The three new tests are marked `browser` despite driving no browser: `drive_dashboard` imports Playwright at import time, so that marker is what gets them run rather than skipped everywhere.

## Test plan

- [x] Full browser suite (10 passed)
- [x] Fast suite
- [x] `--launch --map` with an allocated port, and the in-use guard still firing on an explicit taken port
- [x] `tox -e static`

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -97,6 +97,11 @@ UI, then copying the persisted `workflow_configs.yaml` (strip the runtime
 `current_job` key, keep `jobs`) and `plot_configs.yaml` from the config
 dir back into the fixture; `ui_config_fixtures_test.py` guards against drift.
 
+**Ports.** `fake_dashboard(...)` without a port takes one the OS reports free — how the
+browser tests launch, so two checkouts (or a suite next to an interactive dashboard) can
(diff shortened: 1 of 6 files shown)


SimonHeybrock (commented):
LGTM

PR #1150, opened 2026-07-29
tests: guard against collapsed/inconsistently-sized plots

SimonHeybrock:
Replaces #1030, which was written before the browser-test framework existed and no longer applies. That PR carried four things; only two survive, and both needed rework.

**What was dropped.** The `browser` marker and the `not browser` filters in `tox.ini` landed on main verbatim via #1053, and main now has a `tox -e browser` env with a CI job running it. The plumbing half of #1030 is redundant.

**What changed.** The browser test was written as local-only, on the explicit premise that Playwright-in-CI was a separate, undecided question. That decision has since been made, so this test is now CI-gating rather than a thing to remember to run. It was rewritten against `tests/helpers/browser.py`: no `sys.path` seam, no port literal, no module-scoped fixture. It also now filters to figures with a displayed view — figures of tabs visited earlier stay in the document with their last dimensions, so without that filter a stale figure is reported against whatever tab is active now. Deriving the plot-grid tabs from the last static tab replaces a hardcoded list of the three static tab titles.

The headless invariant test was silently broken by the `ResultKey` → `DataKey` migration of the dashboard data plane: its data no longer reached the plotter, `compute()` turned the failure into a "no data" `Text` placeholder, and the assertions passed against a single default-sized figure. Fixing the keys exposed that this failure mode is invisible, so the placeholder is now rejected explicitly. The matrix gained the missing `data_aspect` type and the second stretch mode for the constrained aspects — the stretch mode decides which axis pane and figure must agree on, so covering only one left half the surface untested.

**Why the source change.** The agreement the invariant asserts — pane sizing mode vs. figure sizing mode — was split across two modules, which is precisely what let them diverge in #1029. Rather than #1030's approach of reaching into `cell._get_sizing_mode` from a test, the mapping moves next to the opts it must agree with, as `frame_aspect.pane_sizing_mode`. Tests then assert against a public function, and the two halves of the invariant are readable side by side.

## Test plan

Both guards were verified to have teeth by mutation: dropping the cross dimension from `make_frame_aspect_opts` (reproducing the #1029 sizing split) fails 48 of 56 headless cases, and the browser test fails with the original symptom, `ih: 0`.

- [x] Headless invariant: 56 cases pass; 48 fail under the mutation.
- [x] Browser test: passes in ~18 s; fails under the mutation with `figures with a data area below 50px: {'Detectors': [{'iw': 631, 'ih': 0, ...}]}`.
- [x] Full default suite green (4236 passed).

### DIFF
--- a/src/ess/livedata/dashboard/frame_aspect.py
+++ b/src/ess/livedata/dashboard/frame_aspect.py
@@ -246,3 +246,28 @@ def make_frame_aspect_opts(aspect: PlotAspect) -> dict[str, Any]:
             return {'responsive': True}
     cross = 'height' if aspect.stretch_mode == StretchMode.width else 'width'
     return {'responsive': True, cross: _INITIAL_CROSS_SIZE_PX, 'hooks': [hook]}
+
+
+def pane_sizing_mode(aspect: PlotAspect) -> str:
+    """Panel ``sizing_mode`` for the pane wrapping a figure sized by this aspect.
+
+    Must agree with the sizing mode HoloViews derives from
+    :func:`make_frame_aspect_opts`: a pane that stretches along only one axis
+    around a ``stretch_both`` figure leaves the other axis unconstrained, and
+    the figure collapses to zero size.
+
+    Parameters
+    ----------
+    aspect:
+        Plot aspect configuration.
+
+    Returns
+    -------
+    :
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1151, opened 2026-07-29
Fix duplicate session-callback registration on every session open

SimonHeybrock:
Every browser session opening the dashboard logged an unhandled `ValueError: A callback of the same type has already been added with this ID`, raised inside Bokeh's `with_document_locked`.

Panel's `pn.io.hold()` cannot release a hold taken while the app callable runs: it classifies the caller as off-thread whenever `state._thread_id` is unset, and Panel only assigns that once it initializes the document — after `create_layout` returns. So the first widget to use `pn.io.hold()` during the build (`DerivedDevicesOverview`) left the document held for the remainder of the build, with its unhold deferred to a next-tick callback. Filed upstream as holoviz/panel#8691.

The per-session periodic callback started at the end of `create_layout` was therefore registered under that hold, so its `SessionCallbackAdded` event was queued rather than dispatched. Bokeh's `ServerSession` constructor registered the callback while sweeping the callbacks already on the document, and the deferred unhold replayed the queued event afterwards — a second registration of the same callback id.

Beyond the log noise, `unhold()` dispatches queued events in a plain loop, so the exception discarded the 38 document events queued behind the offending one. That is benign only because no websocket is attached that early.

`create_layout` now takes the hold explicitly for the whole build and releases it deterministically. Nested `pn.io.hold()` calls see a hold already active, so they neither re-hold nor defer an unhold, and the flush happens while no session exists — each callback is then picked up exactly once by the session's initial sweep. This closes the whole class rather than the one call site: the announcements pane and the dev-mode log producer register callbacks during the build too, and today escape only by running before the first hold.

Once holoviz/panel#8691 is fixed and released the workaround becomes redundant, but it stays correct either way (verified against an emulated upstream fix).

## Test plan

- New unit test asserts the build leaves no hold on the document; it fails on the previous code with `assert 'combine' is None`.
- Full fast suite passes; `-m browser` suite passes.
- Manually verified against a running dashboard: repeated session opens produce zero occurrences of the error (previously one per open), and the UI renders and receives data.

### DIFF
--- a/src/ess/livedata/dashboard/dashboard.py
+++ b/src/ess/livedata/dashboard/dashboard.py
@@ -226,6 +226,31 @@ class DashboardBase(ServiceBase, ABC):
 
     def create_layout(self) -> pn.template.MaterialTemplate:
         """Create the basic dashboard layout."""
+        # Own the document hold for the whole build so that widget code using
+        # ``pn.io.hold()`` nests inside it instead of taking its own.
+        #
+        # A nested ``pn.io.hold()`` would not release here (holoviz/panel#8691):
+        # it decides whether it was called off the document's thread by comparing
+        # against ``state._thread_id``, which Panel only assigns once it
+        # initializes the document -- after this function returns. Every hold
+        # taken here is therefore misclassified as off-thread and defers its
+        # unhold to a next-tick callback. Bokeh constructs the ServerSession in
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1152, opened 2026-07-29
Pop a cell's plot out into a floating window

SimonHeybrock:
## What

A tool in the cell titlebar opens that cell's plot in a draggable, resizable floating window. The cell is untouched: it keeps its plot, its toolbars and its place in the grid. The window renders the cell's *own* composition a second time rather than composing its own, so both views repaint from one layer `Pipe` and share one autoscale controller — the window is live rather than a snapshot, and turning an autoscale axis off in either view turns it off in both.

## Why

A grid cell is often too small to read detail from, and the alternatives (resize the grid, open a modal) either disturb the layout or block the dashboard underneath. A *floating* window is also what makes a plot watchable while working elsewhere in the dashboard: a popped-out cell stays live when its grid is not the visible tab. That is the capability that justifies the machinery — an expand-in-place or a modal would be simpler but could not do it.

## Cost, and why there is no cap on the number of pop-outs

Liveness follows what a window *shows*, not that it exists — the same rule a hidden tab already obeys. Minimizing or smallifying a pop-out puts its cell back to sleep, so popping out many cells and minimizing the windows costs nothing: what is not rendered is not computed. A cell rebuild (a rename, a job restart) carries the window along *with its status*, so a parked window stays parked and its cell asleep, rather than every rebuild popping the parked windows open. A cap would punish the case the feature exists for (comparing two detectors side by side) while leaving the actual cost unbounded, since minimized windows would still count against it. Liveness is also per cell rather than per grid, so the rest of a hidden grid still sleeps.

## Liveness under the gated wake ticks (#1105)

Session ticks are gated on a `has_work` predicate over shared state, and this branch widens it. Every term used to be scoped to the *visible* grid, so a session parked on another tab would evaluate the predicate, find no work, and never run the tick body -- which is where cross-tab pop-out liveness lives. The gate now tracks frame generations per grid: the visible one unioned with the grids behind live pop-outs. Predicate and pass read one `_live_generations`, so they cannot disagree about what counts as live; a minimized window drops out of it by the same rule that lets it sleep.

Flushes are per grid for the same reason the predicate is. A presenter holds pending data as soon as the ingestion thread builds its layer, *before* the grid's frame commits, so a single session-wide flush flag would let a frame for the pop-out's grid push the visible grid's half-built burst (and vice versa), staggering one burst's layers across repaints -- the artifact the frame clock exists to prevent. Each grid's pipes flush only when that grid's own generation moved or it just became visible.

Working a window's controls moves no shared state, only what this session shows, so no predicate can see it. The manager reports those changes and the tab widget requests a full tick, exactly as it already does for tab switches and modal closes -- otherwise a restored window sits frozen at whatever it showed when it was minimized. Opening needs no such call: the pop-out button is in a cell titlebar, so the cell is on the visible tab and already live.

A window must not outlive what it shows: removing the cell closes it, and disabling its grid closes it too, since the grid's layers lose their viewers and the window would float on frozen.

The pop-out tests drive the gated helper rather than the pass directly, so a predicate that misses a pop-out fails there instead of passing while the window freezes in the browser.

## Notes for review

Three findings are worth the reviewer's attention because they cost real time and are now recorded in the automation rules. Panel's `FloatPanel` does not carry a window's size down to its content: the wrappers between jsPanel's content element and ours have no height, which both collapses a stretching child to a sliver and prevents a fixed-height child from ever following the window. Panel re-lays out only on `jspanelresizestop`, which a drag fires but maximize and normalize do not. And an invisible `ReactiveHTML` helper needs a public class name — a leading underscore makes Bokeh fail to resolve the model and takes the whole session down with it.

Browser-test viewports default to 1000 px tall, which hid a geometry bug that only appears below roughly 820 px; the geometry test parametrizes over 700 px for that reason.

## Known gaps

A live window shows the plot alone: the freshness/lag pill stays in the cell titlebar, which a hidden tab does not render, so a stalled stream freezes a pop-out with no staleness cue. Recorded in the module docstring; a window-side freshness cue is future work.

ROI-request layers share one edit stream per presenter, so popping out an ROI cell drives a single `BoxEdit`/`PolyDraw` from two figures. This is untested and likely needs handling before the feature is relied on for ROI work.

## Test plan

Automated coverage added with the feature:

- unit tests for the window lifecycle, the shared composition and autoscale controller, cross-tab liveness, sleeping when minimized, and the wake gate (a frame for a popped-out grid wakes a session on another tab; one for a hidden grid alone, or behind a minimized window, does not);
- unit tests for per-grid flushing (a frame for the pop-out's grid leaves the visible grid's mid-burst data pending, and each grid's frame flushes its own), a rebuild preserving a minimized window and its cell's sleep, closing the window when its grid is disabled, and cascade slots not being reused after a close;
- browser tests that the window opens inside the viewport at both 700 px and 1000 px heights, stays live while another tab is shown, sleeps when minimized and wakes on restore, and that the plot tracks the window across maximize and normalize.

Manual checks for the tester:

- [ ] Pop out several cells at once; each window's controls stay reachable as they cascade.
- [ ] Drag and resize a window; the plot follows without leaving whitespace.
- [ ] Open a pop-out on a laptop-height screen; the title bar and its buttons are reachable.
- [ ] The popped-out toolbar's hover and save tools behave as they do in the cell.
- [ ] Rename a cell that has a pop-out open; the window follows the rebuilt cell.
- [ ] Minimize a pop-out, then rename its cell; the rebuilt window arrives minimized in the taskbar strip, not reopened.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -60,11 +60,14 @@ no text, `title`, or `aria-label` — leaving nothing semantic for browser autom
 *name* identity to slug (`workflow_status_widget.py`'s `_tool_css_class`), but a grid has
 none, so `lt-grid-*` slugs the grid *title* instead (`plot_grid_manager.py`) — the same
 title that also drives the grid's download filename. Plot cells have neither a name nor
-a stable title, so every button in their titlebar — gear (or layer menu), pencil, and
-layer-details toggle — carries the grid position as `lt-cell-r{row}c{col}` (`cell.py`),
-unique per grid, and only the active tab's grid is rendered. Address one with a compound
-selector: `.lt-cell-r0c1.lt-tool-settings`. Do not rely on DOM order for cells: a rebuilt
-cell (e.g. after a rename) moves to the end of the document.
+a stable title, so every button in their titlebar — pop-out, gear (or layer menu),
+pencil, and layer-details toggle — carries the grid position as `lt-cell-r{row}c{col}`
+(`cell.py`), unique per grid, and only the active tab's grid is rendered. Address one
+with a compound selector: `.lt-cell-r0c1.lt-tool-settings`. Do not rely on DOM order for
+cells: a rebuilt cell (e.g. after a rename) moves to the end of the document.
+
+A cell's pop-out window (`plot_popout.py`) is slugged by the same grid position:
+`lt-popout` + `lt-popout-r{row}c{col}`, one per cell at most.
 
 These classes have no associated style rules — adding/removing them is visually inert.
 Treat them as a stable contract: do not drop them in refactors (a test in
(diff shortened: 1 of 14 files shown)


SimonHeybrock:
LGTM, works like a charm so far.

Issue #1155, opened 2026-07-30
Client model graph grows on every plot-cell rebuild: two upstream BokehJS leaks


SimonHeybrock:
Every rebuild of a plot cell leaves the previous cell's entire Bokeh figure subtree alive in the browser's model graph. The server-side document is cleaned up correctly; the client is never told to drop the models.

## Measurement

Fake backend, committed dummy fixture, one browser session on the **Detectors** tab. Six stop/start cycles of `dummy/area_panel_xy` (each generation change replaces the plotter, bumps the layer version and rebuilds the cell), counting the client's model graph after each cycle via `Bokeh.documents[0]._all_models`:

```
cycle 0  total  594   Figures 4
cycle 1  total  733   Figures 5
cycle 2  total  872   Figures 6
cycle 3  total 1011   Figures 7
cycle 4  total 1150   Figures 8
cycle 5  total 1289   Figures 9
```

+139 models and **+1 `Figure` per rebuild**, monotonic. Expected steady state for that fixture is 4 figures (3 cells, one with two layers).

The same probe also shows stale `stopped` freshness pills lingering: after starting the jobs, the client still holds the 7 pill models from the pre-start widgets alongside the 3 new ones. A page reload leaves only the 3.

## The server side is fine

`PlotGrid._insert_widget_into_grid` deletes the target region before assigning, and Panel's `GridSpec._get_objects` calls `old._cleanup(root)` for replaced children. Forcing `doc.models.recompute()` right after each `_build_cell` and reading `len(doc.models._models)`, the server document does not grow by a cell subtree per rebuild — it creeps by roughly one model per stop/start cycle, which is a much smaller, separate question.

So the server knows the old models are unreachable and the client does not.

## Why this matters beyond memory

A detached figure is not inert. The cell's HoloViews hooks are attached to the session's `DynamicMap`, which the stale figure still belongs to, so the stale figure keeps being rendered and keeps invoking those hooks. That is what broke the autoscale toolbar: `CellAutoscaleController` installed its `CustomAction` tools once per controller, and whichever figure rendered first consumed the installation, leaving the figure the user actually sees with an empty toolbar (flaky, and the reason autoscale toggles disappeared after tab switching). Fixed separately by keying installation on the toolbar, but the leak is the underlying cause and will keep producing this class of symptom.

It is also a plausible contributor to long-session slowdown: every job restart adds a full figure to the graph the client recomputes.

## Not yet diagnosed

[...]

PR #1157, opened 2026-07-30
Install autoscale tools per toolbar, not once per controller

SimonHeybrock:
A 2D image plot's autoscale toggles (and Fit) sometimes never appeared, or disappeared after switching tabs, leaving the color range frozen at whatever the first frame produced.

A cell's autoscale hook is attached to the session's `DynamicMap`, which HoloViews can render into more than one Bokeh figure: a kdim/Layout figure swap, or — far more often — a rebuilt cell whose previous figure is still alive in the client model graph (#1155). `CellAutoscaleController` installed its `CustomAction` tools behind a one-shot latch, so whichever figure rendered first consumed the installation and the figure the user sees was left with an empty toolbar. Which figure wins is a race, hence the intermittency.

Installation is now keyed on the toolbar. The tool models are still created once and shared across figures, so a toggle the user turned off stays off across a figure swap.

Note this fixes the symptom, not the cause: rebuilt cells should not be leaving detached figures behind at all. That is #1155.

## Test plan

- New regression test: a controller whose hook renders into a second figure installs the tools there too, sharing the same tool models. Fails on `main` (second figure gets zero tools).
- `tests/dashboard`: 1957 passed.
- Manual, fake backend + dummy fixture: 2D image cells keep Color/X/Y autoscale + Fit across four tab-switch cycles, with no duplicated buttons.

### DIFF
--- a/src/ess/livedata/dashboard/cell_autoscale.py
+++ b/src/ess/livedata/dashboard/cell_autoscale.py
@@ -107,7 +107,6 @@ class CellAutoscaleController:
         # session's own Bokeh document (see dashboard-widgets rules).
         self._toggles: dict[Axis, Any] = {}
         self._fit_tool: Any | None = None
-        self._tools_installed = False
         # Last target written per axis. Read back on subsequent off-state
         # renders so the c-axis freeze has a stable value to apply.
         self._last_targets: dict[Axis, tuple[float, float]] = {}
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
I doubt upstream fixes will be quick, so I think this symptom fix seems necessary.

PR #1164, opened 2026-08-03
Add BEER instrument configuration

### LINKED ISSUE #1158: BEER requirements for ess live data
This is a list of requirements to start the implementation of ess live data for BEER

_Raw data_

- accelerator: 1D plot as function of time
- monitor: 1D intensity plot as a function of time for the 3 monitors
- detectors:
    - 2D plots of each panel (CC and HC), 2D plots of banks (First Science and later)
    - 1D intensity plot for each bank
- sample environment: for example temporal evolution of temperature...
- Scanning: 3D plot of CAD of sample (to be also stored in the NeXus file) + cloud of points to be collected with highlight of progress of data collection (data collection time per point: a few seconds)

_Simple reduction_
1D plot as function of d-spacing for each bank

SimonHeybrock:
Adds BEER as a registered instrument, covering the generic device setup, the three beam monitors and the two detector banks from #1158. The 3D scanning plot is explicitly out of scope.

Everything is derived from `coda_beer_999999_00015484.hdf`, which has **not** been verified against the real instrument. Several things in it look wrong or provisional; they are listed below so they can be taken up with the instrument team.

## What is wired up

Detector banks `beer_detector_s2` / `beer_detector_n2` (Kafka sources `detector_a` / `detector_b`) get two logical views: a bank image summed over the 12 panels, and a per-panel 3D view. Both are spatially downsampled — the bank image to 250x250 (4 mm pixels), the panel view to 12 x 125x125 (8 mm pixels). At the file's 1 mm pitch a single bank is 12M pixels, which no live view can accumulate over 100 TOA bins. The standard detector-view scalar outputs already provide the requested per-bank intensity-versus-time trace, so no extra workflow is needed for that.

Detector numbers are supplied explicitly rather than loaded from a geometry artifact, because the file carries no detector placement (see below) and logical views do not need one. No geometry artifact is generated or registered.

The three monitors use the standard TOA monitor workflow. Their source names do not follow the `cbm{i}` convention (`cbm1`, `cbm2`, `hereon`), which is why `make_common_stream_mapping_inputs` gained an explicit-LUT argument. `cbm1`/`cbm2` publish da00 histograms and `hereon` publishes ev44 events; both schemas already route to the same workflow.

All 78 f144 streams on `beer_motion`, `beer_choppers` and `tn_data_general` are declared, which covers the accelerator and sample-environment timeseries requirements. The hexapod axes are renamed to a `hexapod/<axis>/<leaf>` prefix, since the auto-derived names would otherwise be bare `position_x/value` and friends.

Choppers are deliberately not declared: the wavelength-LUT auto-wiring requires a registered geometry artifact, and the chopper source names in the file are placeholders.

## Example

image

## Findings on the baseline NeXus file

- **The detector banks have no placement.** Both carry `depends_on = '.'` with no `NXtransformations` group, and their pixel offsets are byte-identical, so the file puts both banks at the origin. The +/-90 degrees at 2 m from the requirements is absent. Consequence: no geometric projection and no geometry artifact are possible today.
- **The file mixes two coordinate origins.** The sample is at the origin, yet the choppers (6.4 to 80 m), collimators and slits (152 to 158 m) carry positive z at what look like source-referenced distances, placing them far downstream of the sample. The monitors, at z = -6, -11.5 and -157 m, are the only components consistent with a sample-at-origin frame.
- **Most f144 source names are `TODOnn` placeholders.** Every stream on `beer_motion` and `beer_choppers` except the two slit-gap PVs. Two consequences: those timeseries are subscribed but will receive nothing until upstream fills them in, and no motorised device is auto-detected — device detection keys off the EPICS `.RBV`/`.VAL`/`.DMOV` source suffixes, so the hexapod and slit axes appear as ~60 separate value/target/idle streams instead of ~21 merged devices.
- **The detector resolution is probably provisional.** 12 x 1000 x 1000 pixels of 1 mm per bank, i.e. 24M pixels total, would make BEER by far the largest detector in the system. The *structure* (12 panels summed to give the bank image) matches what the instrument scientist described, so it is taken at face value, but the resolution is what forces the downsampling above.
- The two spatial axes are equally sized, so the file does not reveal which one `x_pixel_offset` indexes. A swap would transpose the images and nothing else.

## Not done

`docker.yml` is left untouched — that matrix is the set of instruments we publish images for, and BEER is not deployed yet.

## Test plan

- [x] Full test suite passes (four pre-existing `to_nxlog_test.py` failures from a numpy deprecation, unrelated and also failing on `main`)
- [x] Backend service `--check` for all four services and dashboard `--check`, for `beer`
- [x] End-to-end detector service test extended to cover `beer`

Closes #1158 (3D requirement will be discussed offline and tracked elsewhere).


SimonHeybrock (commented):
LGTM

PR #1165, opened 2026-08-03
Migrate generics to PEP 695 syntax (UP046/UP047)

### LINKED ISSUE #1161: Migrate generics to PEP 695 syntax (UP046/UP047)
Raising the minimum Python to 3.12 (#1160) makes ruff infer `target-version = py312`, which activates the PEP 695 rules `UP046` (`non-pep695-generic-class`) and `UP047` (`non-pep695-generic-function`). They are currently suppressed in `pyproject.toml` so that the version bump stays a version bump.

Migrating means rewriting 24 classes and 3 functions from `Generic[T]` / module-level `TypeVar` to `class Foo[T]` and `def f[T]()`. Affected: `Message`, `MessageSource`, `MessageSink`, `MessagePreprocessor`, `OrchestratingProcessor`, `Accumulator`, `PreprocessorFactory`, `IdentityProcessor`, `ConfigurationAdapter`, `DataServiceSubscriber`, `PlotterFactory`, `BaseROIRequestPlotter`, `TemporalBufferManager`, `BufferProtocol`, `Transport`, `WizardStep`, `SingleRunReport`, and others.

Worth doing on its own because it is a typing-semantics change, not a formatting one:

- PEP 695 type parameters are implicitly invariant; any `TypeVar` currently declared with `covariant=`/`contravariant=` or with bounds changes meaning if converted mechanically.
- A module-level `TypeVar` shared by several classes does not convert one-to-one — each class gets its own parameter.
- Several of the affected types are pydantic models or `Protocol`s, where runtime introspection differs from the `Generic[T]` form and needs verifying rather than assuming.

Ruff can autofix much of it, but the result needs review and a full test run, including the pydantic-generic and protocol paths.

SimonHeybrock:
## Summary

Closes #1161.

Raising the minimum Python version to 3.12 made ruff infer `target-version = py312`, activating `UP046`/`UP047` (non-PEP-695 generic class/function), suppressed in a follow-up since the rewrite changes typing semantics rather than just formatting.

Converts all flagged classes/functions to `class Foo[T]` / `def f[T]()`, including:
- Implicitly-generic classes that inherited a subscripted generic base without their own `Generic[T]` (e.g. `KafkaAdapter`, `ChainedAdapter`, `DataService`).
- Method-scoped generics distinct from their class's own parameters (e.g. `ChainedAdapter.__init__[U]`, `PlotterRegistry.register_plotter[P]`).
- Module-level `TypeVar`s previously shared across several classes/files, now split into a parameter per class (e.g. `message.py`'s `Tin`/`Tout` used by `processor.py`/`preprocessor.py`; `Events` shared between `to_nxevent_data.py` and `group_by_pixel.py`).
- Bound and constrained type parameters, and a dropped explicit `covariant=True` since PEP 695 infers variance.

`AccumulationMode` and `to_nxevent_data.Events`'s use as a sciline `Scope` specialization key across independent `NewType`/`Scope` declarations in several unrelated files are left as plain `TypeVar`s -- not expressible as a class- or function-scoped PEP 695 parameter.

A before/after mypy diff confirms no new errors and resolves eight pre-existing "invariant type variable used in protocol where covariant/contravariant expected" warnings, since mypy now infers variance for PEP 695 parameters instead of requiring an explicit `covariant=`/`contravariant=` TypeVar.

## Test plan
- [x] `ruff check` / `ruff format --check` clean on all touched files
- [x] Full test suite: 4321 passed, 4 pre-existing failures unrelated to this change (confirmed present on `main`)
- [x] mypy before/after diff: zero new errors, 8 resolved variance warnings

### DIFF
--- a/docs/about/loader.py
+++ b/docs/about/loader.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 from dataclasses import dataclass
-from typing import NewType, TypeVar
+from typing import NewType
 
 import pandas as pd
 from environments import BenchmarkResultFilePath, BenchmarkRootDir
@@ -9,10 +9,8 @@ from runner import BenchmarkReport
 
 ResultPath = BenchmarkResultFilePath
 
-T = TypeVar("T")
 
-
-def _extract_only_if_one(values: list[T], target_name: str) -> T | None:
+def _extract_only_if_one[T](values: list[T], target_name: str) -> T | None:
     if len(cands := set(values)) > 1:
         raise ValueError(f"More than 1 {target_name} found in the list.")
     elif len(cands) == 0:
@@ -25,7 +23,7 @@ def _extract_unit(units: list[str | None]) -> str | None:
     return _extract_only_if_one([unit for unit in units if unit is not None], "unit")
 
 
-def _extract_dtype(values: list[T]) -> type[T] | str | None:
+def _extract_dtype[T](values: list[T]) -> type[T] | str | None:
     none_type = type(None)
     cands = [valt for val in values if (valt := type(val)) is not none_type]
     if (extracted := _extract_only_if_one(cands, "type")) is str:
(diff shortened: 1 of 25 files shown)


SimonHeybrock (commented):
Finally, I have been waiting for this :))

LGTM

PR #1166, opened 2026-08-03
Give the dashboard's static tabs icons

SimonHeybrock:
The three fixed tabs sit next to a variable number of plot-grid tabs, and were distinguishable from them only by bold text. Each now carries a leading icon; grid tabs deliberately carry none, so a bare label is what identifies a user-created grid.

Dashboard tab bar with icons on the Workflows, System Status and Manage Plots tabs

Bokeh renders a tab label as plain text, so the icon cannot be part of the title. It is painted on a `::before` pseudo-element selected by tab position — the same positional mechanism the bold rule already used. The icon is a CSS **mask** rather than a background image, so it inherits the tab's `currentColor` and follows the active-tab accent (blue above) without a second, recolored copy.

The Workflows icon is Tabler's `sitemap` rotated 90°, so its branches converge left-to-right: several input streams reduced into one result. It is registered as `workflow` rather than `sitemap` because it is no longer Tabler's glyph in Tabler's orientation — the same reason `autoscale-y` is not called `arrow-autofit-content`.

The screenshot above lives on the throwaway branch `pr-assets-tab-icons`, which can be deleted once this is merged.

## Test plan

Existing dashboard tests cover the wiring: a bad icon name raises `KeyError` from `get_icon` while `PlotGridTabs` is constructed.

- [x] Icons render and are legible at the default zoom
- [x] The active tab's icon takes the accent color, inactive ones stay dark
- [x] Plot-grid tabs remain unstyled, including after adding/renaming a grid

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -43,6 +43,12 @@ def refresh(self):
 
 Do not use Unicode characters for button icons. Use embedded SVG icons from `dashboard/widgets/icons.py` via `get_icon()`. Use the `create_tool_button()` helper from `dashboard/widgets/buttons.py` for consistent styling.
 
+Where the icon cannot be a widget — a Bokeh tab label is plain text, for instance —
+paint it on a `::before` pseudo-element with `mask-image: url(get_icon_data_uri(name,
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
LGTM

PR #1167, opened 2026-08-03
Fix slicer plots failing to render inside a grid cell

SimonHeybrock:
Any 3D output plotted with the slicer raised an `AssertionError` deep inside HoloViews instead of rendering. This is not data-specific — it reproduces with the `(z, y, x)` fixture the existing slicer unit tests already use, and it affects every 3D logical view (ESTIA and TBL multiblade, and BEER's forthcoming panel view).

## Cause

`SlicerPresenter.present()` declared the static element options on its DynamicMap. HoloViews cannot apply element options to a DynamicMap statically, so it wraps it in a `Dynamic` operation whose callback takes `*args`; `validate_dynamic_argspec` therefore leaves that wrapper without `_posarg_keys`. The grid cell then attaches its own hooks (SaveTool filename, hover suspend, autoscale) with a second `.opts()` call, and the resulting outer wrapper passes the kdim values as keywords. Rebuilding the positional key from those keywords requires the inner wrapper's `_posarg_keys`, so the inner map ends up indexed with an empty key and `_validate_key` asserts.

Put differently: a DynamicMap that carries kdims tolerates exactly one `.opts()` call. Either half alone renders fine; both together fail. Only the slicer is affected, because every other presenter builds a DynamicMap with no kdims, where indexing with an empty key is valid.

Introduced when the slicer's styling was hoisted onto the DynamicMap to stop it being reassembled on every render.

## Fix

Style each rendered slice instead, folded into the `.opts()` call that already applied the pre-computed clim. The options dict is assembled once in the presenter, and the number of `.opts()` calls per render is unchanged, so the measured extra cost is ~40 us per slice.

Cell-side workarounds were tried first and none help: `clone=False`, element-typed hooks, and collating through a single-layer `hv.Overlay` all still produce the second wrapper.

## Tests

Red/green, both verified to fail before the fix and pass after:

- `test_presenter_dmap_renders_with_cell_hooks` — renders the slicer the way a grid cell does (present, then attach hooks) and would have caught this directly. A sibling test pins the same invariant for the 2D image presenter.
- `test_slicer_figure_adopts_the_panes_sizing_mode` — extends the render-geometry invariant to the slicer. Moving styling onto the render path makes it easy to drop silently; deleting the styling makes this fail for every aspect type.

## Test plan

- [x] Full test suite passes (four pre-existing `to_nxlog_test.py` failures from a numpy deprecation, also failing on `main`)
- [x] A 3D detector view renders in a grid cell (confirmed against BEER's per-panel view on #1164)
- [ ] Slice/flatten controls work and the plot resizes with the cell

### DIFF
--- a/src/ess/livedata/dashboard/slicer_plotter.py
+++ b/src/ess/livedata/dashboard/slicer_plotter.py
@@ -16,13 +16,11 @@ from ess.livedata.config.workflow_spec import DataKey
 from .data_roles import PRIMARY
 from .plot_params import PlotParams3d, PlotScale, PlotScaleParams2d, TickParams
 from .plots import (
-    _HOVER_ELEMENTS_2D,
     Plotter,
     PresenterBase,
     _hv_axis_padding,
     _normalize_to_rate,
     _pad_range,
-    _typed_opts,
 )
 from .range_hook import Axis
 from .scipp_to_holoviews import to_holoviews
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1168, opened 2026-08-03
Bound ToNXlog history by size and age

### LINKED ISSUE #1127: ToNXlog buffer grows unbounded for the process lifetime
`ToNXlog` accumulates every f144 sample for the process lifetime; `clear()` is never called in production, and the capacity-doubling buffer grows without bound.

Constraints for a bounding/trimming design, established while resolving finding 3 of #1037:

- `TimeseriesStreamProcessor` locates new samples by absolute positional index into the shared buffer (`_last_returned_index`). Any trimming breaks positional indexing, so publication bookkeeping must become timestamp-based first.
- BIFROST's `group_by_rotation` does a per-event `sc.lookup(..., mode='previous')` against the raw buffer; queries before the first retained entry return NaN. Retention must keep at least one sample at or before the oldest `event_time_zero` in the current detector batch.
- Restart-refill fidelity: on a workflow commit the dashboard clears its buffers (generation flip) and the freshly activated job republishes the retained history. Correlation plots (`FullHistoryExtractor` consumers of timeseries streams) rebuild from that refill, so once the backend trims, every restart silently truncates their axis history to the retained window. The widest frontend consumer — not just BIFROST's lookup window — is a stakeholder in the retention policy.

Related: the frontend `TemporalBuffer` for timeseries keys has the same unbounded-growth property, with correlation plots as a legitimate full-history consumer.

SimonHeybrock:
Closes #1127.

`ToNXlog` accumulated every f144 sample for the process lifetime. Measured on this branch, the cost of that grew without limit: the capacity-doubling `sc.concat([ts, ts])` took 7.5 ms at 1M samples, 44 ms at 16M and 86 ms at 33M, and transiently needed **3x** the allocated capacity — 1.4 GB for a single 16M-sample device log.

History is now bounded by `max_size` samples and `max_age` of timespan, whichever binds first. One pair of defaults (1M samples, 30 days) covers both regimes without per-stream config, because the two knobs bind on different logs:

| log | binding knob | retained |
|---|---|---|
| slow f144, one sample / 2 s | `max_size` | 23 days |
| medium, 1 Hz | `max_size` | 12 days |
| chopper, 14 Hz | `max_size` | 20 hours |
| sample environment, 1/min | `max_age` | 30 days |

That asymmetry is the point: week-long experiments depend on the slow logs, and fast logs are read for recent behaviour rather than history.

## Three properties worth reviewing

**Trimming copies nothing.** It moves a start offset, so it is affordable on every append. Without that, `max_age` can only be enforced when the buffer happens to fill — which is what makes the dashboard's equivalent a retention *floor* rather than a bound.

**Compaction reallocates instead of shifting in place.** `get()` hands out views aliasing the buffer, and `sc.lookup` holds a reference to them rather than copying. I verified an in-place shift rewrites data behind live consumers: BIFROST's rotation lookup then fails with `Coordinate of lookup function must be sorted`. So the pattern `VariableBuffer.drop` uses on the dashboard side is not transplantable here — it is safe there only because `DataService` copies extractor output before handing it out. Reallocating leaves outstanding views owning the old buffer.

The pause is then bounded by `max_size` alone — 0.24 ms at 50k samples, 0.90 ms at 200k, 2.9 ms at 1M — so no incremental or background compaction scheme is warranted.

**Relocations are phase-shifted per stream.** The relocation period is a fixed number of *samples*, so equal-rate logs that start together stay in phase for the process lifetime and relocate in the same batch every time. Measured at 109 ms for 100 logs relocating together. Seeding the slack from a stable hash of the stream name (not `hash`, so it survives restarts) spreads them: with 100 equal-rate logs, the worst batch relocates 4.

Age is measured against the newest buffered sample rather than wall clock, so a device that stops reporting keeps its last known value instead of ageing out to an empty buffer — consumers rely on the buffer being non-empty, and retention always keeps the newest sample at or before the age cutoff so `sc.lookup(mode='previous')` has an anchor for every query inside the window.

## The timeseries service gets a shallower bound

Bounding retention exposed what retention is actually *for* in each service. Detector and reduction read the buffer as a lookup window and need the depth. The timeseries service does not: its only reader of history is the backfill a job receives on activation, published as its first `delta`. The wavelength-LUT workflow hosted by the same service reads only the newest sample, and reset was removed in #1126. Retention there is a backfill budget.

That distinction matters because the budget is paid per stream, all at once. Instruments declare hundreds of log streams — ODIN close to 300 — and the dashboard can start a job for each at one click. At 1M samples that is 16 MB per stream and over a gigabyte published in a single cycle, and it is not recoverable traffic: the sink drops what it cannot produce or queue while the workflow has already advanced its publication cursor, so an oversized backfill is lost rather than retried.

`BACKFILL_MAX_SIZE = 10_000` keeps a week of history for a log ticking once a minute — the regime long experiments depend on — while capping a 14 Hz chopper's backfill at ~160 kB. Starting every job at once then moves tens of megabytes. No evidence exists that anyone needs deeper backfill here; the number is trivially raised once a requirement appears.

Its floor is the number of samples one stream can deliver between two service cycles. Below that, retention drops samples before the workflow publishes them and the delta stream has a silent gap, which `TimeseriesStreamProcessor` now warns about — conservatively, since the retained window alone cannot prove a sample existed in the gap. Without that warning, any future tightening of the bound would be unfalsifiable.

## Not addressed here

The frontend `TemporalBuffer` still has the same unbounded-growth property, bounded only by a hardcoded 20 MB per key. Its cap and the backend's are really the two ends of one policy: `FullHistoryExtractor` asks for `float('inf')`, so the dashboard keeps whatever the backend sends and backend retention silently determines correlation-plot history after any workflow restart (generation flip → refill). The two should eventually become one finite window rather than independently drifting policies. A hundred plots at 20 MB each is also unbudgeted in aggregate on the dashboard side.

Per-log caps still do not bound the aggregate across streams in the detector and reduction services. In the timeseries service the aggregate is now bounded in practice, which is where the one-click fan-out lives.

## Test plan

- [x] `tests/preprocessors/to_nxlog_test.py`: 17 new cases covering both bounds, the lookup anchor, relocation integrity (values/coords/variances), view aliasing, and phase spreading
- [x] `tests/workflows/timeseries_test.py` (new, 12 cases) — the timestamp-cursor prerequisite, plus the silent-gap warning firing and staying quiet while trimming stays behind the cursor
- [x] `tests/preprocessors/timeseries_test.py` (new) — the backfill budget applies to all three preprocessor-creation paths (f144, device, synthetic `chopper_cascade`)
- [x] Full non-integration suite: 4428 passed. The 4 `to_nxlog_test.py` failures are pre-existing numpy-timedelta deprecations in the test file, confirmed identical on `main`.

### DIFF
--- a/src/ess/livedata/preprocessors/data_reduction.py
+++ b/src/ess/livedata/preprocessors/data_reduction.py
@@ -23,7 +23,9 @@ class ReductionPreprocessorFactory(JobBasedPreprocessorFactoryBase):
             case StreamKind.MONITOR_COUNTS:
                 return Cumulative(clear_on_get=True)
             case StreamKind.LOG | StreamKind.DEVICE:
-                return nxlog_for_stream(self._instrument.streams.get(key.name))
+                return nxlog_for_stream(
+                    self._instrument.streams.get(key.name), name=key.name
+                )
             case StreamKind.MONITOR_EVENTS:
                 return ToNXevent_data()
             case StreamKind.DETECTOR_EVENTS:
--- a/src/ess/livedata/preprocessors/detector_data.py
+++ b/src/ess/livedata/preprocessors/detector_data.py
@@ -53,7 +53,9 @@ class DetectorPreprocessorFactory(JobBasedPreprocessorFactoryBase):
             case StreamKind.LIVEDATA_ROI:
                 return LatestValueAccumulator()
             case StreamKind.LOG | StreamKind.DEVICE:
-                return nxlog_for_stream(self._instrument.streams.get(key.name))
+                return nxlog_for_stream(
+                    self._instrument.streams.get(key.name), name=key.name
+                )
             case _:
                 return None
 
--- a/src/ess/livedata/preprocessors/timeseries.py
+++ b/src/ess/livedata/preprocessors/timeseries.py
@@ -26,6 +26,21 @@ logger = structlog.get_logger(__name__)
 #: would have).
 _SYNTHETIC_LOG_SOURCES: frozenset[str] = frozenset({CHOPPER_CASCADE_SOURCE})
 
+#: Retained samples per log in this service, shallower than
+#: :data:`DEFAULT_MAX_SIZE`, which detector and reduction workflows need for their
+#: lookups. Here the buffer has a single consumer: the backfill a job receives on
+#: activation, published as its first ``delta`` (the wavelength-LUT workflow reads
+#: only the newest sample). Retention is therefore a backfill budget, capping both
+#: that message and the RAM held for every log stream, plotted or not. Instruments
+#: declare hundreds of log streams and the dashboard can start a job for each at
+#: one click, which publishes every buffer at once. This keeps a week for a log
+#: ticking once a minute -- the regime long experiments depend on -- and ~160 kB
+#: of backfill for a 14 Hz chopper. Raise it when a concrete requirement for
+#: deeper backfill appears. It must stay well above the samples one stream can
+#: deliver between two service cycles, or they are dropped before publication --
+#: the gap :class:`TimeseriesStreamProcessor` warns about.
+BACKFILL_MAX_SIZE = 10_000
+
 
 class LogdataPreprocessorFactory(
     JobBasedPreprocessorFactoryBase[LogData, sc.DataArray]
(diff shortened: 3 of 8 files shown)


SimonHeybrock:
LGTM

PR #1171, opened 2026-08-03
Keep cell titlebar panes visible and current after a rebuild

### LINKED ISSUE #1154: Freshness pill stays blank when jobs are started in an already-open session
A session that is already open when a job is started often ends up with a permanently blank freshness pill on the affected cells, while the plots themselves update live. Reloading the page fixes it.

## Reproduction (fake backend, no Kafka)

1. Launch with the committed dummy fixture but *without* `--auto-start`, so the layers start out `STOPPED`:
   `python -m ess.livedata.dashboard.reduction --instrument dummy --transport fake --config-dir <copy of tests/dashboard/ui_config_fixtures/dummy's parent> --collapsed-sidebar --no-fetch-announcements`
2. Open the session and switch to the **Detectors** tab. All three cells show the gray `stopped` pill.
3. Go to **Workflows**, start the workflows, come back to **Detectors**.
4. Plots render and keep updating. The titlebar pills are empty and stay empty (observed for 15+ s).
5. `page.reload()` → pills appear immediately (`0.1s`, `0.7s`, …).

Reproduced 2/2 in this state. It is a race, not a fixed ordering: adding log statements inside `_poll_for_plot_updates` slowed the pass enough that it stopped reproducing (3/3 clean), which also matches the "happens sometimes" reports from live use.

## What is ruled out

Instrumented in the failing runs, on the cell widget that was actually inserted into the grid:

- `CellWidget.update_freshness` **is** called every poll for the affected cells, so the cell is on the active grid and `is_active` holds.
- `_pill_frozen` is `False` (it is only `True` for the `stopped`/`error` pills, which render non-empty text).
- The merged `TimeBounds` handed to `update_freshness` is **not** `None` — real bounds, roughly one data period old.
- The `_stopped_layers` snapshot is empty.

So the widget receiving the update is the one on screen and its inputs are good, yet the pill stays blank. The remaining gap is between `self._freshness_pane.object = html` and what the browser renders.

## Structural contributors found while looking

These are all in `PlotGridTabs._poll_for_plot_updates`, and are worth fixing regardless of which one is decisive:

- **Bounds are sampled before the first `compute()`.** `bounds = state.plotter.time_bounds` is read at `plot_grid_tabs.py:782`, but `activate_layer` at `:807` is what synchronously produces the layer's first computed frame (0→1 viewer transition → `_refresh_layer` → `_pull_and_build` → `plotter.compute`, which sets `_time_bounds` in `plots.py`). On the pass where a cell first becomes visible, the pill is therefore always computed from a pre-compute `None`.
- **Every rebuild re-blanks the pill.** `_build_cell` mints a fresh `CellWidget`, hence a fresh `create_freshness_pane()` with `object=''`. Refilling needs a *later* freshness-due poll.
- **`_last_freshness_update` advances even when nothing was updated.** It is set whenever `freshness_due` is true (`:882`), including passes where `active_cell_bounds` is empty (hidden grid), so a rebuild landing just after such a pass waits for the next flush or the 2 s stall.
- **A `STOPPED`-at-setup layer cannot be un-frozen by data alone.** `_setup_layer` drives a layer to `STOPPED` when the workflow's active job number is not yet observed. `data_arrived()` is an explicit no-op in `STOPPED` and does *not* bump the version (`plot_data_service.py:147`), so no cell rebuild follows; only `sync_job_states` observing the job-number transition unfreezes it. This is the pre-start state in the reproduction above.

## Suggested direction

Sample `state.plotter.time_bounds` *after* `activate_layer` (or re-read it in the freshness loop at `:884`), and force `freshness_due` for cells rebuilt in the current pass, so a newly built/revealed cell fills its pill on the same pass that renders its plot instead of depending on a later one.

## Context

Found while investigating a separate autoscale-toolbar bug in the same code path. Related and possibly interacting: rebuilt cells leak their whole Bokeh figure subtree into the browser's model graph (~139 models and one extra `Figure` per stop/start cycle), so a stale detached figure for the same cell survives in the client — that is what breaks the autoscale toolbar's one-shot tool installation.

SimonHeybrock:
Cells rebuilt while their tab is open lost their whole titlebar — cell title and freshness pill both — until the page was reloaded. The plots themselves kept updating, which is what made this look like a freshness-data bug.

## Root cause

Panel keeps a markup pane's content behind `visibility: hidden` from the moment it renders and reveals it only once every `<link>` stylesheet in the pane has fired `load` (`PanelMarkupView.watch_stylesheets` -> `style_redraw`). That reveal is armed exactly once, during `render()`.

A pane built after page load — every cell the plot poll loop rebuilds — first renders with the design's stylesheet URLs pointing at cdn.holoviz.org, because its model is not attached to a document yet and Panel falls back to `CDN_DIST`. Panel then patches those URLs to the locally served copies, which swaps the pane's `<link>` elements. The load events the reveal is waiting on belong to the discarded elements and never arrive, so the pane stays invisible for the rest of the session with a correct, live-updating Bokeh model behind it.

CDN reachability is not the trigger, only who wins the race: stubbing every CDN request with a `200` still reproduced it, because the links are detached before they finish. That is why it looked intermittent. It reproduces in ~20 lines of plain Panel 1.9.3 (`MaterialTemplate` plus a pane appended from a periodic callback) and does not reproduce without the template, so it is an upstream bug — reported as holoviz/panel#8696. A `Design` modifier overrides the reveal for every markup pane in the app until a Panel release re-arms it.

## Also fixed

The structural contributors listed in the issue, which delayed a rebuilt cell's pill and time range by up to a poll each: bounds were sampled before `activate_layer` (i.e. before the 0->1 transition that computes a newly revealed layer's first frame), a rebuild did not make the panes due for a refresh, and the stall clock restarted on passes that updated nothing.

The fourth item — a `STOPPED`-at-setup layer not being un-frozen by data alone — is left alone. `data_arrived()` is a deliberate no-op in `STOPPED` ("a layer bound to a stopped workflow's retained data renders it but stays STOPPED"), so the gap is in how `_setup_layer` classifies "job number not observed yet" versus "genuinely stopped". That is a state-machine design question, not a drive-by.

## Tests

A browser regression test rebuilds a cell in an open session and asserts the title and the pill are visible; it blocks cdn.holoviz.org, as the deployment network does, so the discarded links fail for certain rather than racing the swap. It fails on `main` and passes here. Two unit tests cover the refresh timing, and one covers the design modifier. The reordering of the bounds sample has no unit coverage: the poll-test harness bypasses data subscriptions, so `activate_layer` never computes there.

## Test plan

- [x] `pytest tests/dashboard` (1972 passed)
- [x] `pytest -m browser` (12 passed)
- [x] Manual: open a session, start a stopped workflow from the Workflows tab, return to the plot tab — pills and titles show without a reload

Closes #1154. Workaround-removal reminder: #1175.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -207,8 +207,8 @@ against `cdn.holoviz.org` URLs — its model is not in a document yet, so Panel 
 to the CDN — and Panel then swaps those links for the locally served copies. The load
 events the reveal waits on belong to the discarded links, so the pane can stay invisible
 for the rest of the session while its model, text and layout are all correct and
-live-updating (#1154). `dashboard/design.py` overrides the latch for the whole app; a
-template built without `LivedataDesign` brings the bug back.
+live-updating (#1154, holoviz/panel#8696). `dashboard/design.py` overrides the latch for
+the whole app; a template built without `LivedataDesign` brings the bug back.
 
 ## Colors and styling
 
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -198,6 +198,18 @@ Note that `dynamic=True` only gates Bokeh model creation. Python-side periodic c
 (e.g., `SessionUpdater` custom handlers) still run for all registered widgets regardless
 of which tab is visible. Use an `is_visible` predicate to skip refresh work for hidden tabs.
 
+### Markup panes built after page load
+
+Panel keeps a markup pane (`pn.pane.HTML`, `Markdown`, `Str`) behind
+`visibility: hidden` until every `<link>` stylesheet in it has fired `load`, and arms
+that reveal exactly once, while rendering. A pane built after page load first renders
+against `cdn.holoviz.org` URLs — its model is not in a document yet, so Panel falls back
+to the CDN — and Panel then swaps those links for the locally served copies. The load
+events the reveal waits on belong to the discarded links, so the pane can stay invisible
+for the rest of the session while its model, text and layout are all correct and
+live-updating (#1154). `dashboard/design.py` overrides the latch for the whole app; a
+template built without `LivedataDesign` brings the bug back.
+
 ## Colors and styling
 
 All colors must come from `dashboard/widgets/styles.py`. Do not hard-code hex color values
--- a/src/ess/livedata/dashboard/dashboard.py
+++ b/src/ess/livedata/dashboard/dashboard.py
@@ -14,6 +14,7 @@ from ess.livedata import ServiceBase, __version__, format_version
 
 from .config_store import ConfigStoreManager
 from .dashboard_services import DashboardServices
+from .design import LivedataDesign
 from .fake_backend import FakeBackendTransport
 from .kafka_transport import DashboardKafkaTransport
 from .session_registry import SessionId
@@ -294,6 +295,7 @@ class DashboardBase(ServiceBase, ABC):
 
         template = pn.template.MaterialTemplate(
             title=self.get_dashboard_title(),
+            design=LivedataDesign,
             sidebar=sidebar_with_heartbeat,
             collapsed_sidebar=self._collapsed_sidebar,
             main=main_content,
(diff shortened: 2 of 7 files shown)


SimonHeybrock:
Works, LGTM

PR #1172, opened 2026-08-03
Let a plot layer recover from a transient error

### LINKED ISSUE #1169: A plot layer that errors once never recovers: ERROR is only cleared by job_started
## Summary

`LayerState.ERROR` is only ever left via `job_started` — that is the whole exit in the transition table (`dashboard/plot_data_service.py`). Nothing else clears it.

Meanwhile the build path keeps running for an errored layer: the data subscription marks it dirty regardless of state, and `flush_frames` gates only on `has_viewers`. So after a single transient extraction or compute failure:

- extraction and `plotter.compute` recover and keep succeeding, silently;
- `data_arrived()` is rejected from ERROR, logging `Invalid transition: data_arrived() called in state ERROR (expected WAITING_FOR_DATA)` **once per flush**, i.e. at data rate;
- `has_displayable_plot()` stays False (it requires READY or STOPPED), so the cell renders the error placeholder;
- the layer stays stuck until the workflow restarts (new generation → `_reset_layer_presentation` → `job_started`) or the layer is re-added.

Net effect: one transient fault permanently blanks a plot whose data is flowing and computing fine, and emits a warning per frame while doing so.

## Reproduction

Appending this to `tests/dashboard/plot_orchestrator_test.py` reproduces it with real components — an extractor that fails once, then works:

```python
def test_transient_extraction_error_recovery(
    workflow_id, workflow_spec, job_orchestrator, fake_data_service, fake_stream_manager
):
    from ess.livedata.dashboard.extractors import LatestValueExtractor

    fail = [True]

    class FlakyExtractor(LatestValueExtractor):
        def extract(self, data):
            if fail[0]:
                raise RuntimeError("extractor boom")
            return super().extract(data)

    controller = FakePlottingController(
        stream_manager=fake_stream_manager, extractor_factory=FlakyExtractor
    )
    plot_data_service = PlotDataService()
    orchestrator = PlotOrchestrator(
        plotting_controller=controller,
        job_orchestrator=job_orchestrator,
        data_service=fake_data_service,
        instrument='dummy',
        plot_data_service=plot_data_service,
    )
    grid_id = orchestrator.add_grid(title='G', nrows=3, ncols=3)
    config = make_plot_config(workflow_id)
    cell_id = add_cell_with_layer(orchestrator, grid_id, DEFAULT_GEOMETRY, config)
    layer_id = orchestrator.get_cell(cell_id).layers[0].layer_id
    commit_workflow_for_test(job_orchestrator, workflow_id, workflow_spec)

    class _Viewer:
        pass

    viewer = _Viewer()
    orchestrator.activate_layer(layer_id, viewer, True)

    feed_result_data(fake_data_service, config, workflow_id)
    orchestrator.flush_frames()

    fail[0] = False
    for _ in range(3):
        feed_result_data(fake_data_service, config, workflow_id)
        orchestrator.flush_frames()

    assert plot_data_service.get(layer_id).state is LayerState.READY
```

Observed: state stays `LayerState.ERROR`, `has_displayable_plot()` stays False, and three `Invalid transition: data_arrived() called in state ERROR` warnings are emitted — one per flush.

## Not observed in production

Seven days of dashboard journals across the bifrost, dream, loki and tbl hosts contain no `Invalid transition` lines, so this is not currently firing in the field. Filing on the strength of the reproduction rather than a field report.

## Suggested direction

Make ERROR self-healing: accept `data_arrived()` from ERROR, on the grounds that a successful compute is evidence the fault cleared. The trade-off is that genuinely intermittent failures would make a cell alternate between plot and error rather than settling — noisier, but more truthful than being permanently stuck on a stale fault.

If flapping turns out to matter, the fallback is to clear ERROR only after several consecutive successes, at the cost of extra machinery that is probably not warranted yet.

Either way, a transition the system routinely and legitimately attempts should stop being logged as invalid.

## Provenance

Found while evaluating #1060. The two are unrelated: #1060's race is silent by construction and cannot produce this log line. This one needs no concurrency at all.

SimonHeybrock:
Fixes #1169.

`LayerState.ERROR` was only ever left via `job_started`, yet the build path keeps running for an errored layer: the data subscription marks it dirty regardless of state and `flush_frames` gates only on `has_viewers`. So a single transient extraction or compute failure left a layer permanently blank — extraction and `plotter.compute` recovered and kept succeeding silently, `has_displayable_plot()` stayed False so the cell rendered the error placeholder, and `Invalid transition: data_arrived() called in state ERROR` was logged once per flush, i.e. at data rate. Recovery required a workflow restart or re-adding the layer.

A completed build is evidence the fault cleared, so `data_arrived()` now recovers the layer to READY and clears the error message. The trade-off, as noted in the issue: a genuinely intermittent fault makes the cell alternate between plot and error rather than settling — noisier, but not stale. Counting consecutive successes before clearing is the fallback if flapping turns out to matter; it does not look warranted yet.

One further consequence worth flagging: a layer whose plotter *re-creation* fails on a generation change (`_reset_layer_presentation`) also self-clears on the next frame, and keeps rendering with the previous generation's plotter — so its internal accumulation, such as autoscale ranges, is not reset. I think that is right, since `create_plotter` succeeded earlier with the same config and such a failure is therefore environmental rather than deterministic, but it does turn that error into a transient one in the UI.

A stop is not evidence the fault cleared, so `job_stopped()` while in ERROR is now an explicit no-op rather than a logged "invalid transition". This is reachable for layers whose plotter creation or pipeline setup failed — no data can flow, so they cannot heal, and the workflow stopping later emitted a spurious warning. Keeping the error also matches the cell titlebar, where an errored layer already outranks a stopped one.

Not observed in production: seven days of dashboard journals across the bifrost, dream, loki and tbl hosts contain no `Invalid transition` lines.

## Test plan

- New orchestrator test with real components (extractor that fails once, then works) reproduces the stuck layer on `main` and passes here.
- New state-machine tests cover ERROR → READY, the resulting `has_displayable_plot()`, and ERROR surviving a stop.
- Full dashboard suite passes. The four `tests/preprocessors/to_nxlog_test.py` failures in the full run are a pre-existing numpy timedelta deprecation, unrelated to this change.

### DIFF
--- a/src/ess/livedata/dashboard/plot_data_service.py
+++ b/src/ess/livedata/dashboard/plot_data_service.py
@@ -46,11 +46,17 @@ class LayerState(Enum):
     STOPPED → WAITING_FOR_DATA           [job_started(plotter)]
     STOPPED → ERROR                      [error_occurred(msg)]
 
+    ERROR → READY                        [data_arrived()]
     ERROR → WAITING_FOR_DATA             [job_started(plotter)]
 
     A layer in STOPPED may still display retained data: ``data_arrived``
     while STOPPED is a no-op and ``has_displayable_plot`` reports True once
     the plotter holds a computed frame.
+
+    ERROR is left by evidence that the layer works again — a completed build
+    — or by a fresh job. A stop is not such evidence, so ``job_stopped`` while
+    in ERROR is a no-op: an error needs attention and outranks the frozen-
+    snapshot presentation of STOPPED.
     """
 
     WAITING_FOR_DATA = auto()
@@ -139,7 +145,13 @@ class LayerStateMachine:
         """
         Transition to READY when data arrives.
 
-        Valid from: WAITING_FOR_DATA.
+        Valid from: WAITING_FOR_DATA and ERROR. Recovering from ERROR is
+        deliberate: extraction and compute keep running for an errored layer,
+        so a completed build proves the fault cleared and a transient failure
+        must not blank the plot until the workflow restarts. An intermittent
+        fault therefore makes the cell alternate between plot and error —
+        noisier than settling, but not stale.
+
         No-op from READY (data continues to arrive after first update) and
         from STOPPED (a layer bound to a stopped workflow's retained data
         renders it but stays STOPPED).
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1173, opened 2026-08-03
Print browser-test failure diagnostics, and reap killed dashboards

SimonHeybrock:
## Why

The Playwright suite has been failing on roughly a third of CI runs (9 of the last 25 browser jobs; before Aug 2 the preceding 13 were green). It is genuine flakiness, not a regression: runs 30783040638 and 30783040550 are the same commit, one failed and one passed.

There are two distinct failure modes, and only one is fixed here.

## Fixed: killed dashboards were never reaped

The launcher terminates the fake-backend dashboard and, if it does not exit within 10 s, kills it — but never waits on it. An unreaped child is still running when its `Popen` is garbage-collected, so `__del__` raises a `ResourceWarning`, `filterwarnings = ["error"]` turns it into a failure, and it lands on whichever test happened to be running when the collection fired. That is why it always showed up against the session-churn test and never against the test that actually leaked.

## Not fixed: the modal timeouts (#1174)

The other mode is a click that Playwright reports as landed on the cell pencil or the layer-picker entry, followed by a `[role=dialog]` that never becomes visible. It has never reproduced locally — 11 runs here, 6 of them under 8x CPU throttling, all passed — and CI gave nothing but the bare locator timeout, so it has been re-investigated from scratch every time and diagnosed none of the times.

## So: record the failure, on stdout

A failing session now prints what it knew at the moment it failed: the browser console tail, the page state, a base64 PNG screenshot, and the dashboard's server log tail. Capturing unconditionally is the point — by the time a test is known to have failed, the run that failed is over, and a rerun passes.

All of it goes to stdout rather than a build artifact. pytest folds captured output into the failure report, and the run-level log zip is downloadable while the run is still going, whereas the artifact endpoint is firewalled in the devcontainer where the debugging actually happens. An upload that silently matches no files is also indistinguishable from a run that captured nothing, which is not hypothetical: the first cut of this branch wrote its trace to a dotted directory, `actions/upload-artifact` skips hidden paths unless told otherwise, and it uploaded nothing for the branch's entire life without ever saying so.

The page state is aimed squarely at #1174: which tab is up, how many `[role=dialog]` nodes exist and how many of those are visible, and the `lt-*` hooks present to be clicked. No dialog in the DOM means the click never reached its handler; a dialog present but hidden means it rendered and was hidden again. That is the split the bare timeout never gave us.

One thing to know when reading the console tail: the constant `pageerror: Cannot read properties of undefined (reading 'parent_style')` is *not* the flake. It is bokeh/bokeh#15274 — our plot grid is a `pn.GridSpec`, and Panel emits a `children` change plus the recomputed sizing props in one patch, which `GridBoxView._update_layout` indexes inconsistently. It fires on passing runs too. Real bug, since it drops the rest of the patch, but unrelated to any intermittent failure. The rule file records this so it is not chased twice.

Dropping tracing removes a confound worth naming, too: recording DOM snapshots throughout is not free, and it may itself have shifted the timing of the race being chased. If the modal failures change character rather than staying put, that is the first thing to suspect.

## Test plan

- [x] `pytest -m browser tests/dashboard/` — 14 passed.
- [x] The failure path was exercised with a deliberately raised error: page state, console tail, screenshot and server log tail all appear, and the screenshot round-trips byte-exact out of the captured log via the `base64 -d` pipeline the log prints beside it.
- [x] The reaping fix has a unit test that drives `_terminate` against a process which ignores SIGTERM.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -103,6 +103,26 @@ UI, then copying the persisted `workflow_configs.yaml` (strip the runtime
 `current_job` key, keep `jobs`) and `plot_configs.yaml` from the config
 dir back into the fixture; `ui_config_fixtures_test.py` guards against drift.
 
+**Diagnosing a failure.** When a block driving a session raises, everything is printed
+to **stdout** — browser console tail, page state (active tab, dialog count and how many
+of them are visible, `lt-*` hook inventory), a base64 JPEG screenshot, and the
+dashboard's server log tail. Nothing is uploaded as a CI artifact, deliberately: the
+artifact endpoint is firewalled in the devcontainer, whereas stdout lands in the pytest
+failure report and in the run-level log zip, which is allowlisted and readable mid-run
+(`gh api repos/scipp/esslivedata/actions/runs/<id>/logs > run.zip`). Recover the
+screenshot with the `base64 -d` pipeline the log prints next to it.
+
+`dialogs` vs `dialogs_visible` is the first thing to read on a modal timeout: a dialog
+present but not visible means it rendered and was hidden, none at all means the click
+never reached its handler.
+
+One console line is a red herring: `pageerror: Cannot read properties of undefined
+(reading 'parent_style')` fires on **every** run, passing or failing. It is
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1176, opened 2026-08-04
Add single-command launcher for the local development stack

SimonHeybrock:
Running the dev stack meant one terminal per service, retyping the instrument (and the NeXus file to replay) everywhere on a switch, and restarting every process by hand after a code change.

`esslivedata-dev <instrument>` (or `python -m ess.livedata.scripts.dev`) now creates the topics, starts the fakes, the backend services and the dashboard as subprocesses, merges their logs into one terminal with a colour-coded per-service prefix, and stops them all together on Ctrl-C or when any of them dies.

```
esslivedata-dev dream                      # everything
esslivedata-dev dream monitors timeseries  # groups as positionals, no flags
esslivedata-dev dream --reload             # restart the stack when sources change
```

Groups are `detectors`, `monitors`, `reduction` and `timeseries`; the dashboard runs unless `--no-dashboard` is given. The `reduction` group starts the fakes plus `data_reduction` only, since reduction consumes the raw streams itself.

The remaining per-machine detail, the NeXus file to replay, comes from a git-ignored `dev.toml` found by walking up from the working directory, so that switching instrument stays a one-token change:

```toml
[nexus_files]
dream = "~/data/268227_00024779_Si_BC_offset_240_deg_wlgth.hdf"
```

Pairing the fakes with `--dev` consumers is now done by construction. Hand-written command lists tend to get this wrong: the fakes publish to the dev topic names unconditionally, so a consumer started without `--dev` subscribes to the production topics and silently sees nothing.

Topic creation moves out of the integration-test conftest into the launcher, which needs the same thing against a pristine broker.

## Test plan

Unit tests cover command construction, group resolution, `dev.toml` lookup, the source watcher and the subprocess supervisor (including killing a job that ignores SIGINT). The full stack against a live broker was not run, as the dev container this was written in has neither docker nor Kafka. Every generated backend command was instead checked against the real service parsers via `--check`.

- [x] `esslivedata-dev dummy` brings up the whole stack and the dashboard shows data
- [ ] `esslivedata-dev dream detectors` with a `dev.toml` entry replays the configured file
- [x] Ctrl-C shuts everything down, leaving no stray processes
- [ ] `--reload` restarts the stack on a source edit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ docs/reports/
 *.zip
 *.sqw
 *.nxspe
+
+# Machine-local settings for the dev stack launcher (ess.livedata.scripts.dev)
+dev.toml
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,8 +56,17 @@ Uses [Copier](https://copier.readthedocs.io/) with [Scipp template](https://gith
 
 Services require Kafka (`docker-compose up kafka`). Use `--dev` for simplified topic structure.
 
+The whole stack runs in one terminal via `esslivedata-dev <instrument>` (or
+`python -m ess.livedata.scripts.dev`): it creates the topics, starts the fakes, the
+backend services and the dashboard, prefixes their logs per service, and stops them
+together on Ctrl-C. Name groups (`detectors`, `monitors`, `reduction`, `timeseries`) to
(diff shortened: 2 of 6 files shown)


SimonHeybrock (commented):
Works, will see how useful it turns out to be or whether we need to build sth. different.

PR #1177, opened 2026-08-04
Drop the cosmetic projection enum from detector view specs

SimonHeybrock:
`register_detector_view_spec` took a `projection` argument typed against its own `ProjectionType` literal, but never passed it to a workflow — it selected only the spec's name, title and description. The projection actually performed comes from `GeometricViewConfig.projection_type` in the instrument's factories, and nothing checked that the two agreed, so a spec could advertise "Detector XY Projection" while projecting onto a cylinder mantle.

Two enums for one concept also meant they could drift, and they had. The cosmetic one never gained `cylinder_mantle_y`, so MAGIC had to bypass the helper to register its spec; DREAM kept a bank-to-projection dict whose values nothing read; and the docstrings in `projectors.py` still described the dispatch as handling only `xy_plane` and `cylinder_mantle_z` although it had handled the Y case for a while.

The helper had one caller left, LOKI, which now registers its spec directly the way DREAM, MAGIC and dummy already did. `GeometricViewConfig.projection_type` is now the only place a projection is named.

## No user-visible change

All 56 registered workflow specs across every instrument compare byte-identical before and after (title, description, source names, params, aux sources, group), so no persisted dashboard config is invalidated. LOKI's `detector_xy_projection` keeps its WorkflowId in particular.

## Docs

The instrument guide taught the helper as *the* way to declare a geometric projection. It now shows the two-phase pattern instruments actually use — spec in `specs.py`, one `GeometricViewConfig` per bank in `factories.py` — and lists the projections from the surviving enum.

## Drive-by

`test_no_heavy_imports` asserted nothing. It imported `ess.livedata.preprocessors.detector_view_specs`, which does not exist — the module is under `workflows/` — and only ever passed because its own `ess.reduce in sys.modules` guard skipped it in a full run; in isolation it raised ImportError. It now runs the import in a subprocess, where the check is meaningful. Verified it fails if `ess.reduce` is pulled in.

## Test plan

- [x] Full suite passes. The four `to_nxlog_test.py` failures are pre-existing on `main` (numpy timedelta deprecation) and reproduce with this branch's changes stashed.
- [x] Registered specs diffed before/after across all instruments: identical.
- [x] Import-isolation test verified against a negative control.

### DIFF
--- a/docs/user-guide/adding-new-instruments.md
+++ b/docs/user-guide/adding-new-instruments.md
@@ -474,37 +474,64 @@ instrument.add_logical_view(
 
 #### Geometric Projections (complex geometries)
 
-For detectors with complex 3D geometries, use helper functions to register projections:
+Detectors with complex 3D geometries are projected using their calibrated pixel
+positions. This follows the two-phase pattern: `specs.py` registers *that* there is a
+projection workflow, `factories.py` declares *which* projection each bank uses.
+
+In `specs.py`:
 
 ```python
-from ess.livedata.workflows.detector_view_specs import register_detector_view_spec
-
-# Single projection for all detectors
-xy_handle = register_detector_view_spec(
-    instrument=instrument,
-    projection='xy_plane',
-    source_names=['detector_0', 'detector_1'],
+from ess.livedata.config.workflow_spec import DETECTORS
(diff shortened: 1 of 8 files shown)


SimonHeybrock:
LGTM

PR #1178, opened 2026-08-04
Fix numpy DeprecationWarning failures in to_nxlog tests

SimonHeybrock:
numpy deprecated adding bare integers to datetime64 (implicit 'generic' timedelta unit); with `filterwarnings = error` this fails four tests in `to_nxlog_test.py` under current numpy. Use explicit `np.timedelta64(..., 'ns')`.


SimonHeybrock (commented):
LGTM

Issue #1153, opened 2026-07-29
Timeseries downsampling destroys the peaks of a cumulative sawtooth


SimonHeybrock:
`downsample_timeseries` keeps the last sample of each bucket. For a monotonically increasing curve that sample is the bucket maximum, so the shape survives. For a curve that resets, it is an arbitrary point on the tooth: whenever a reset falls inside a bucket, the pre-reset peak is dropped and the tooth is rendered at whatever value the accumulation happened to have reached at the bucket boundary.

This matters now that #1058 (PR #1135) makes cumulative outputs plottable as history. NICOS resets a counting job once per scan point (ADR 0006: move motor, reset, count, repeat), so a cumulative counts output on a wall-clock axis *is* a sawtooth — one tooth per scan point, its height the counts collected at that point. Reading the tooth heights is the point of the plot.

With the default downsampling params (fine period 1 s over a 1 h recent window, coarse period 300 s beyond it) a scan resetting every ~30 s reads correctly while it is live and has its tooth heights silently corrupted once it ages out of the recent band. Nothing on the plot indicates that the older part of the curve is no longer quantitative.

Not a regression: the reducer predates cumulative history plots, and for the per-update and series data it was written for, last-per-bucket is a reasonable choice. This is the first plot for which the choice of reducer changes what the data says.

Two directions, both cheap:

- Keep min *and* max per bucket. Doubles the point count in the coarse band but preserves the envelope for any signal, without needing to know anything about the field.
- Keep the bucket maximum for cumulative fields specifically. Exact for a monotone segment, and it preserves the peak across a reset rather than an arbitrary point after it. `Temporality` is already declared per output field and already reaches the plotter's data source resolution, so the plotter can know which rule to apply.

The second is more faithful where it applies; the first needs no plumbing. Worth deciding before the sawtooth becomes something users read numbers off.

Issue #870, opened 2026-04-14
Overlay silently mixes elements with different coordinate or value units


SimonHeybrock:
## Problem

When multiple data sources are overlaid via `Plotter.compute()`, there is no validation that they share the same coordinate or value units. HoloViews silently accepts an `hv.Overlay` of elements with inconsistent kdim/vdim units — Bokeh renders the axis using the first element's unit metadata while plotting all other elements' raw values unscaled.

This was exposed by #868: fixing the monitor histogram output to preserve the user-configured edge unit meant event-mode monitors (ev44) now correctly output `ms` coordinates, while some other monitors (e.g., da00 histogram-mode) may still produce `ns`. The overlay then mixes `ms`-scale values (0–71) with `ns`-scale values (0–71,000,000) on a single axis, making some curves appear collapsed near zero.

### Affected scope

The bug is **not** limited to `LinePlotter` or to coordinate dimensions:

- **Coordinate dimensions (X axis):** Affected for `LinePlotter` (the originally reported case). `ImagePlotter` and `Overlay1DPlotter` are technically affected too, though overlaying 2D images is unusual in practice. `Overlay1DPlotter`'s internal overlay (slices of the same array) is safe, but the outer overlay in `compute()` is not.
- **Value dimensions (Y axis for curves, colorbar for images):** Same issue. If sources produce different value units (e.g., `counts` vs `counts/s`), the axis label shows the first element's unit while all others are plotted unscaled. `BarsPlotter` is also affected here (categorical kdim, but value units can mismatch).

The unit information flows through `create_value_dimension` and `coord_to_dimension` in `scipp_to_holoviews.py` — each element carries its own units, but the overlay never reconciles them.

## Expected behavior

The plotter should either:
- Convert all data to a common unit before overlaying, or
- Raise a clear error on unit mismatch

## Location

`Plotter.compute()` in `src/ess/livedata/dashboard/plots.py` — the `hv.Overlay(plots)` call at line 468 has no unit consistency check. The fix should go here, before the overlay is created, covering all plotter types.

SimonHeybrock:
# Overlay unit validation (issue #870)

## Problem

When multiple data sources are overlaid, there is no validation that they share
the same coordinate or value units. HoloViews silently renders all elements
against a single shared axis using the first element's unit metadata, plotting
all other elements' raw values unscaled.

This affects both coordinate dimensions (X axis) and value dimensions (Y axis
for curves, colorbar for images), and all plotter types that go through the
[...]

Issue #757, opened 2026-03-05
Correlation plot: Histogram Bins and X-axis does not show output name/title


SimonHeybrock:
In the third step of the plot creation wizard, only the source title is shown (such as "Incident Monitor"), but the output title (which shows correctly on the dropdown in the second step) is omitted.

We seem to have the same problem for the X-axis of the resulting correlation plot - it only displayed the source name, not the output name. I presume for a 2D correlation plot the same also goes for the Y-axis.

PR #1180, opened 2026-08-04
Pin numeric values, units, and labels of rendered plot elements

SimonHeybrock:
## Numeric correctness tests for the display pipeline

Nothing in the test suite asserted what the dashboard actually shows. Plotter tests checked that an element of the right *type* came out; the browser suite checks only that a checksum changed. The bug class that keeps escaping us is the other one: a plot that renders fine and shows the wrong numbers. Recent examples are #1153, #933, #844, #870, #703 and #757 — every one of them produced a plausible-looking figure.

This adds unit-level tests that pin exact values, coordinates, units and axis labels through the real plotters, for lines, images, overlays, correlation histograms and the scipp-to-HoloViews conversion. Known input, exact expected output. Among the things now pinned: bin edges surviving as edges while line mode moves contents to the centres, error elements carrying standard deviations rather than variances, image rows and columns keeping their place with the extent taken from the outer edges, log color scaling masking non-positive pixels without touching the rest, and the correlation join's bin contents and per-second normalization factor.

Rate normalization (#844) previously had three integration tests that asserted `result is not None`, which the toggle being a silent no-op would have passed. They now pin the factor, and a new test drives counts through a real `TemporalBuffer` and `WindowAggregatingExtractor` into the plotter to pin the counts/s the user sees end to end.

Three open issues reproduce at unit level and get strict xfails, so fixing them flips a test loudly: #1153 (the coarse downsampling band drops the peak of each tooth of a cumulative sawtooth), #870 (an overlay mixes layers with different coordinate or value units, twice — once per axis), and #933 (a window aggregation over a source that has gone quiet keeps reporting its last frame as the current window). No production code changes.

### Test plan

- [x] `pytest -n auto` green, with the four strict xfails above. Four pre-existing failures in `tests/preprocessors/to_nxlog_test.py` are unrelated and reproduce on `main`.


SimonHeybrock (commented):
LGTM

PR #1181, opened 2026-08-04
Close wire-coverage gaps: data_reduction, f144 timeseries, no-replay contract

SimonHeybrock:
## What

Adds three integration tests covering pipelines that had no broker-level coverage at all, and one fix to the service-process harness that restarting a service requires.

Stacked on the `integration-test-hygiene` PR, which rescoped `tests/integration/` to wire-only assertions; this PR fills the holes that rescoping made visible.

## Why

**data_reduction** had a `reduction_services` fixture that no test used, so the reduction pipeline had never crossed a broker. The wire residue over the existing in-process service test is addressing: `data_reduction` and `detector_data` share both the commands topic and the raw detector topic, so a start command has to reach the one service hosting the workflow, and its results have to come back keyed to the job_number the dashboard minted. Job status is where that round trip closes, since the data plane's keys carry only the source name.

**The f144 log path** had nothing between broker and service loop under test. An f144 flatbuffer has to resolve through the stream mapping — topic plus in-payload source name — to a configured source and pick up its unit from the instrument's stream record before the service test's starting point is reached. The test produces the payloads itself rather than running `fake_logdata`, which hardcodes a BiFrost source name and therefore emits nothing the dummy instrument can map; see the findings below.

**The no-replay contract** is what `assign_all_partitions` exists for, and it was asserted only against a fake consumer. The consequence — durable commands sitting behind the assigned offsets are never acted on — is a fact about records a service does not read, which only a broker can show. A start command issued while a worker is down is therefore lost rather than delayed, which is the mechanism behind the deliberate asymmetry documented on `JobOrchestrator.reconcile_observed_jobs`. The proof is ordering rather than timing: both commands go to the same single-partition topic, and results for the later one establish that the worker consumed past the earlier one's offset, so the stale job's absence cannot be read as "not reached yet".

Restarting a service also needed a harness fix: readiness is detected by matching captured subprocess output, which a restarted process inherited from the previous run and so reported ready before it had started.

## Findings, no code change here

`with_run_control_route()` has been a no-op since the filewriter topic's prod auth issue (#795), so no service subscribes to that topic and a pl72/6s4t run transition has no observable effect end-to-end. Run control was the fourth intended gap; both sides of the disabled seam are already pinned in unit tests (`tests/kafka/run_control_adapter_test.py`, `tests/core/run_transition_test.py`), and an integration test would only re-assert a `return self`. The integration conftest still creates the filewriter topic, which is harmless and keeps the path ready for re-enabling.

`fake_logdata` ignores its `--instrument` argument and always publishes source name `detector_tank_angle_r0`, a BiFrost name. For every other instrument its messages are dropped as unmapped before adaptation, so the service produces nothing usable.

`pydantic.BaseModel()` is instantiated as an empty params model in two places in `src/ess/livedata/dashboard/`; pydantic v2 raises on that, so both paths are broken.

## Test plan

There is no local broker in this environment, so these tests are validated by CI. Locally verified: they collect under `-m integration`, and the default suite is unaffected.


SimonHeybrock (commented):
LGTM

PR #1182, opened 2026-08-04
Fix silent drop of multi-ROI edits (numpy-array truthiness)

SimonHeybrock:
The browser syncs BoxEdit/PolyDraw columns back as numpy arrays, whose truth value raises for every length but one. The truthiness guards in `RectangleConverter`/`PolygonConverter` therefore raised for 0 or 2+ ROIs, and since the edit handler logs and swallows, drawing more than one ROI — or a single ROI touching coordinate 0 — silently dropped the entire edit. Guard on length instead.

Surfaced by adding an ROI-request cell to the browser-test fixture (follow-up PR), which logged the error on every dashboard start. Parametrized tests (0/1/2 ROIs with array-typed columns) fail on the old guard.

### DIFF
--- a/src/ess/livedata/dashboard/roi_request_plots.py
+++ b/src/ess/livedata/dashboard/roi_request_plots.py
@@ -107,7 +107,9 @@ class RectangleConverter:
         :
             Dictionary mapping box index to RectangleROI. Empty boxes are skipped.
         """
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1183, opened 2026-08-04
Fake backend: stopped/error job states; dummy fixture: ROI, timeseries, table, slicer cells

SimonHeybrock:
## What

Makes the in-process fake backend (`--transport fake`) reproduce two backend states it previously never emitted, and enriches the committed dummy UI fixture so the browser tests cover the plotter families it did not reach.

The fake backend now emits a terminal `stopped` job status when a job is stopped, and gains `FakeBackend.fail_job`, which faults a running job into the `error` state: it keeps heartbeating that state and stops producing results, exactly as the real backend does when a workflow raises. Only `active` was ever emitted before, so the dashboard's stopped and errored presentations (frozen freshness pill, error badge) were not reachable without Kafka. Synthesized values are also generalised from special-cased 1-D/2-D bumps to an N-D drifting Gaussian, so 3-D outputs carry structure along every axis instead of being flat.

The dummy fixture gains an ROI-request cell (image plus the readback and interactive-rectangle overlays the UI stacks on it), exercising the ROI loopback the fake backend already implemented but nothing drove, and a second grid tab holding one cell each for the `timeseries`, `table` and `slicer` plotters. `render_geometry_test.py` visits every grid tab, so those plotters are now measured in a real browser on every browser-test run. The slicer needs 3-D data, which the dummy instrument had none of; it gains a logical detector view folding panel_0's rows into stacked layers, mirroring the 3-D views TBL, ESTIA and BEER already register.

Adding the ROI-request cell surfaced a real production bug (multi-ROI edits silently dropped), split out as #1182, which this PR stacks on.

One deliberate divergence from production: a stop command in the real backend removes the job without a terminal status (`stopped` is only emitted at service shutdown). The fake emits it on stop so the stopped presentation is drivable; the dashboard has already deactivated the workflow locally at that point, so the extra status is inert there.

The fixture drift guard is extended to check that each layer's plotter accepts the view it is pointed at, and that every plotted workflow is staged — mismatches previously survived config load and only showed up as a red cell at compute time.

## Test plan

- [ ] `Detectors` tab: the panel_0 image carries the ROI toolbar; drawing two rectangles shows both as readback outlines and yields two ROI spectra
- [ ] `Diagnostics` tab: the timeseries cell draws two device curves that keep extending
- [ ] `Diagnostics` tab: the table cell shows one row per source with a live total
- [ ] `Diagnostics` tab: the slicer renders a slice and its dimension/index controls step through the volume
- [ ] Stopping a workflow from the Workflows tab freezes its cells' freshness pills instead of letting them age


SimonHeybrock (commented):
LGTM (superficial review only, not production code)

PR #1184, opened 2026-08-04
Test the plot-creation flow: wizard unit tests and a click-to-place browser test

SimonHeybrock:
# Test the plot-creation flow

Creating a plot — clicking out a region on an empty grid, then walking the configuration wizard — was the least-tested part of the product. `plot_config_modal.py` sat at 25% statement coverage, and no test at any level drove the click-to-place gesture: every plot in every browser test came pre-seeded from the fixture, so a regression anywhere along that path would have shipped silently.

The wizard is now driven end to end in unit tests through its rendered widgets — the same handlers a browser goes through, rather than the steps' internals: step transitions and Back, the group/workflow/output cascades, plotter options derived from the output's dimensionality, parameter forms for two plotter types, source pre-selection, the two rejection paths (an empty source selection, and 'since run start' on a view with no cumulative stream), cancel, edit mode, and the resulting `PlotConfig`. The two-click region gesture is covered at the grid seam, including that the region normalizes regardless of which corner is clicked first. Coverage of the modal goes 25% → 79%. One browser test now takes the whole path for real: two clicks on a bare grid, through the wizard, to a plot receiving live data.

Making that testable needed two small production changes. Empty grid cells and download buttons carried no `lt-*` automation hook, so there was no stable way to address the click target; they now follow the existing convention (`lt-empty-cell-r{row}c{col}`, `lt-tool-download`), with the hook inventory in the widget rules updated. The wizard's radio groups passed Panel's deprecated `button_style`, whose `PendingDeprecationWarning` turned every headless construction of the modal into a test error under `filterwarnings = ["error"]`; `variant` takes the same values and renders identically.

One behaviour is pinned as it stands today rather than changed here: stepping Back to the plotter step resets a non-default plotter choice, because the step rebuilds its radio group and preselects the first option, while the workflow/output step keeps its selection. Worth a follow-up decision.


SimonHeybrock (commented):
LGTM

Issue #1174, opened 2026-08-03
Browser tests: cell modal intermittently never opens after a click that landed


SimonHeybrock:
Two browser tests fail together in CI, several times a week, on branches with nothing to do with either of them:

- `tests/dashboard/browser_ui_test.py::test_cell_title_survives_noop_save_of_cell_properties_modal`
- `tests/dashboard/browser_ui_test.py::test_multi_layer_cell_gear_picks_the_layer_to_configure`

Both fail the same way: a click that Playwright reports as landed (on the cell pencil, or on a layer-picker menu entry), followed by `[role=dialog]` never becoming visible within the 10 s wait.

```
E  playwright._impl._errors.TimeoutError: Locator.wait_for: Timeout 10000ms exceeded.
E  Call log:
E    - waiting for locator("[role=dialog]").first to be visible
```

## Frequency

Seven of the last 25 browser jobs, spread across unrelated branches and `main` itself. The 13 jobs before Aug 2 were all green, so it either arrived recently or was always latent and is now being sampled more often.

It is flakiness, not a regression: runs 30783040638 and 30783040550 are the same commit, one failed and one passed.

## What is ruled out

Not general runner slowness. Failing jobs run only 40-60 s longer than passing ones (9m05-9m25 vs 8m25-8m35), and that delta is exactly the timeout budget the failures spend — the rest of the suite is not running slower. Raising the 10 s timeout would be treating the symptom.

Not reproducible locally. The full suite passes repeatedly in the devcontainer at roughly the same wall time CI takes.

## Suspicions

Both failing tests are the only ones that open a modal on the **Detectors** tab, i.e. while the fake backend streams at 1 Hz and cells rebuild. `test_concurrent_grid_property_edits_resolve_to_one_title_without_crash` opens its editor on Manage Plots, where nothing is streaming, and has never flaked. A cell rebuild between locating the trigger and its JS acting on the click would produce exactly this signature: the element was there, the click landed, nothing happened. Same rebuild churn as #1155.

A related observation, unconfirmed against the pinned dependency set: driving the Detectors tab locally logs a page error, `Cannot read properties of undefined (reading 'parent_style')`. A JS exception mid-render is another way to end up with a rendered-but-unwired widget. Seen with bokeh 3.9.2; the test env pins 3.9.1, so this may be a red herring.

[...]

PR #1187, opened 2026-08-04
Pause the plot poll while the cell-properties modal is open

SimonHeybrock:
While investigating #1174 I found that the poll-pause gate only ever engaged for one of the two modals.

`_get_active_grid_id` returns `None` while a modal is open, so the poll skips rendering plots the overlay hides anyway. `_show_config_modal` records the modal it opens; `_show_cell_properties_modal` did not. So renaming a cell or editing its layers left the poll rendering — and rebuilding cells — behind the open dialog.

That gate is what keeps the work off the session event loop, which is also where the click that opens the next modal has to be processed. This is not a fix for #1174, but it removes work from the window where that flake happens.

Every close path routes through the modal `open` watcher, so the existing `on_close` teardown clears the gate on Save, Cancel, X and Escape alike — the new test covers the close path as well as the open one, since holding the gate open forever would be the damaging way to get this wrong.


SimonHeybrock (commented):
LGTM

Issue #1185, opened 2026-08-04
Streaming plot grid blocks the session's IOLoop for seconds


SimonHeybrock:
A dashboard session showing a streaming plot grid blocks its own Tornado IOLoop for seconds at a time. This is the root cause behind #1174, but it is a live-dashboard problem in its own right: while the loop is blocked the session serves nothing, and every client interaction queues behind it.

Important caveat: I have not observed such large delays in the deployed dashboards. Evidence here is from CI and runs in a devcontainer.

## Evidence

From the server log captured during a CI browser-test failure (run 30884968215), while the Detectors tab was streaming:

```
GET /static/extensions/panel/bundled/theme/material_variables.css (::1) 4778.14ms
GET /static/extensions/panel/css/button.css                       (::1) 4778.58ms
```

Those are small static files served from localhost. A 4.8 s response time is the event loop not getting a turn for 4.8 s. Neighbouring requests in the same session show the ramp: 0.6 ms, then 20-29 ms, then the two 4.8 s outliers.

Locally the same effect is visible in miniature. Polling a small static asset from the page via `fetch` while the grid streams, with the process pinned to 2 cores:

```
ManagePlots  n=80 median=6.0ms p95=  9.1ms max= 89.2ms  over_1s=0
Detectors    n=80 median=6.0ms p95= 72.3ms max=170.0ms  over_1s=0
```

Manage Plots does no per-tick plot work and stays flat; Detectors has an 8x heavier tail. The shape matches, the magnitude does not — a loaded CI runner is roughly 28x worse than my worst local sample.

## Why it matters beyond CI

Everything a session does — serving assets, receiving widget events, pushing patches — runs on the same loop as the periodic pass. A multi-second pass means a multi-second unresponsive UI for whoever is looking at that grid, and the stall scales with how much work the pass has to do.

## Where to look

`SessionUpdater`'s pass runs all widget work inside `pn.io.hold()` + `doc.models.freeze()`, with a full unconditional pass every 5 s (`session_updater.py`). Two image plots at 64x64 plus titlebars is not obviously seconds of work, so the first job is to measure rather than guess which part dominates — serialization, the freeze/hold batch flush, or the plot model updates.

Related prior work on render cost: #1055.

## Reproducing

[...]

PR #1188, opened 2026-08-04
tests: Re-click rather than wait once for a modal to open

SimonHeybrock:
A click can land and still open nothing, without raising — so `open_modal` spending its entire budget inside a single `wait_for` could only ever recover one of the two ways that happens.

The diagnostics from #1173 show both, and the server log tail tells them apart. Static assets are served off the same IOLoop as everything else, so their timings measure how long that loop was blocked; in the failing runs two of the modal own stylesheets took 4778 ms to serve from localhost. That is a queued round-trip, and waiting longer does fix it. The other signature — sub-millisecond serves and no post-click server activity at all — is a dropped click, and no amount of waiting helps.

So each attempt now re-clicks, giving a larger total budget than before *and* covering the dropped case. `click`/`open_modal` also accept a ready-made locator, which lets the layer-picker menu entry (no `lt-*` hook of its own, previously waiting on the dialog inline) go through the same path.

This is a harness workaround and labelled as one in the docstring. The stalls it papers over are #1185; the full diagnosis is in #1174.

## Test plan

- [x] The three flaky modal tests pass locally
- [ ] Browser job green across a few CI runs (it has been failing ~44% of the time)

### DIFF
--- a/scripts/drive_dashboard.py
+++ b/scripts/drive_dashboard.py
@@ -170,17 +170,15 @@ class Dashboard:
         """
         for attempt in range(retries):
             try:
-                locator = (
-                    self.page.locator(target).first
-                    if isinstance(target, str)
-                    else target
-                )
-                locator.click(timeout=4000)
+                self._locate(target).click(timeout=4000)
                 return
             except PlaywrightTimeoutError:
                 if attempt == retries - 1:
                     raise
                 self.page.wait_for_timeout(1000)
+
+    def _locate(self, target: str | Locator) -> Locator:
+        return self.page.locator(target).first if isinstance(target, str) else target
 
     def open_modal(self, trigger: str | Locator, *, retries: int = 3):
         """Click a trigger and wait for its modal (``[role=dialog]``) to show.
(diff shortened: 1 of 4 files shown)


scripts/drive_dashboard.py
```diff
@@ -154,32 +159,62 @@ def goto_tab(self, name: str) -> None:
...
+            except PlaywrightTimeoutError:
+                if attempt == retries - 1:
+                    raise
+                try:
+                    if self._locate(trigger).is_visible():
+                        self._locate(trigger).click(timeout=4000)
+                except PlaywrightTimeoutError:
+                    # A re-click that cannot land is not the failure worth
+                    # reporting; let the next wait raise on the missing dialog.
+                    pass
```
SimonHeybrock: Make recoveries loud: once retries absorb the failures the browser job goes green and CI stops reporting the ~44% rate — the only metric for judging whether #1185 work helps. A `print` when an attempt beyond the first succeeds (and which recovery it was: later wait vs re-click) keeps the frequency readable from run logs and distinguishes the two mechanisms in the field.

SimonHeybrock (commented):
Solid diagnosis — the queued-vs-dropped split matches what the CI logs show, and the second commit's survived-its-own-click guard already covers the race I was worried about (a queued click opening the dialog after the trigger is consumed). One suggestion inline.

Side note: #1184's `_wizard_click` hand-rolls a similar wait-then-click and can be simplified onto the `str | Locator` click path once this lands.

SimonHeybrock:
Good point, and it would have bitten us — a green job that got there by retrying looks identical to a fixed one, and the retry rate is the only field measurement #1185 has.

Done in 3c47c4b07. A recovered modal now warns, naming which recovery it was (waiting out a queued click vs re-clicking a dropped one) and on which attempt.

One wrinkle worth flagging: a `print` would not have worked. pytest captures stdout and only replays it for *failing* tests, so a recovery inside a passing test would have gone nowhere — the exact case we want to see. Hence a warning, which lands in the run warnings summary. That needed `filterwarnings` to downgrade this one message, since the blanket `error` setting would otherwise have turned every recovery into a failure and defeated the purpose from the other direction. Verified both halves: the recovery message is reported and passes, other `UserWarning`s still fail.

Agreed on #1184 — `_wizard_click` should collapse onto the `str | Locator` path once this lands.

SimonHeybrock (commented):
LGTM

PR #1189, opened 2026-08-04
Keep the plotter choice when stepping back into the plotter step

SimonHeybrock:
Follow-up to the behaviour #1184 pinned rather than changed: stepping **Back** into the plotter step reset a non-default plotter to the first offered one, because the step rebuilt its radio group on every entry. For correlation histograms the same rebuild also cleared the axis-source selectors, so an axis pair chosen before visiting the configuration step was silently lost too.

The step now rebuilds only when the output its plotters were derived from actually changed — the guard the configuration step already applies to its parameter panel, so the three wizard steps finally agree on what surviving a Back means. A plotter picked for one output still does not outlive switching to another output.

Stacked on #1184 (`plot-config-modal-tests`), whose test suite pins the old behaviour; that test is updated here rather than left to fail after a merge. Review/merge #1184 first.

## Test plan

- Unit tests in `plot_config_modal_test.py` cover both directions: Back keeps the choice, changing the output drops it. The updated test fails on the parent branch and passes here.
- Full dashboard suite passes.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/plot_config_modal.py
+++ b/src/ess/livedata/dashboard/widgets/plot_config_modal.py
@@ -704,6 +704,7 @@ class PlotterSelectionStep(WizardStep[OutputSelection | None, PlotterSelection])
         self._initial_config = initial_config
         self._instrument_config = instrument_config
         self._output_selection: OutputSelection | None = None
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1190, opened 2026-08-04
Remove unreachable empty-BaseModel fallbacks for plotter params

SimonHeybrock:
`pydantic.BaseModel()` was instantiated as an empty params model in two places in the dashboard. Pydantic v2 raises `PydanticUserError` ("Pydantic models should inherit from BaseModel, BaseModel cannot be instantiated directly") on that, so both would have blown up — but neither is reachable: `PlotterSpec.params` is a required, non-optional `type[BaseModel]` field. Pydantic rejects `None` at construction, and `register_plotter` always derives the value from the factory's `params` type hint, so the `spec.params is None` / `if spec.params` guards standing in front of those calls can never be taken.

Rather than replace the fallback with an empty model subclass, this removes the dead guards: the field's type already says params is always present. A third site in `cell_properties_modal.py` had the same `if spec.params else None` check — dead for the same reason, harmless but misleading about the contract — and goes too. Removing the guard in `PlottingController` also exposed a duplicate `get_spec` call it was hiding.

For contrast, `WorkflowSpec.params` *is* genuinely `type[BaseModel] | None`, and 13 of 56 registered workflows rely on it (`timeseries_data` on every instrument, plus three detector/counts workflows). Those sites are untouched; whether that optionality earns its keep is a separate question. The distinction that matters here is that `PlotterSpec` enforces presence structurally rather than by convention, so these branches are unreachable by construction and not merely unused today. A plotter that genuinely takes no parameters is already expressible as an empty model — `spec.params(**{})` works — so the `None` branch was never an escape hatch anyone needed.

## Test plan

No behaviour change — the removed code was unreachable, so no test can distinguish before from after. Existing suite passes (4600 passed, 54 skipped, 8 xfailed).

### DIFF
--- a/src/ess/livedata/dashboard/plot_orchestrator.py
+++ b/src/ess/livedata/dashboard/plot_orchestrator.py
@@ -1400,9 +1400,6 @@ class PlotOrchestrator:
             self._logger.warning('Skipping cell with unknown plotter %s', plot_name)
             return None
 
-        if spec.params is None:
-            return pydantic.BaseModel()
-
         try:
             return spec.params(**params)
         except Exception:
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1194, opened 2026-08-04
Make WorkflowSpec.params required, defaulting to NoParams

SimonHeybrock:
`WorkflowSpec.params` was `type[BaseModel] | None`, which forced every consumer to branch on absence: the spec↔factory validator, workflow instantiation, dashboard job staging, monitor factory attachment, and the whole configuration-adapter chain down to the widget that renders the form. Workflows taking no configuration now use `NoParams`, an empty model, which is also the field default — so their registrations simply omit `params`.

The optionality turned out not to be load-bearing anywhere. The clearest evidence: `configuration_widget` already read `if model_class is None or not model_class.model_fields`, collapsing "no model" and "model with no fields" into the same `NoParamsWidget`. Dropping `| None` at the source removes it through the adapter chain, so `model_class()` and `set_aux_sources()` stop returning `Optional` too.

This is not dead-code removal — 13 of 56 registered workflows take no params (`timeseries_data` on every instrument, plus three detector/counts workflows), so the case for the change is schema clarity, not disuse.

Two things worth a reviewer's attention:

- **Extra-params rejection changed mechanism.** Sending params to a workflow that takes none was previously a hand-rolled `ValueError`; it is now `extra='forbid'` on `NoParams`, so it raises `ValidationError`. Without `extra='forbid'` pydantic would silently ignore the extra fields, which would be a quiet regression. That path had no test at all — this PR adds one, verified to fail if `extra='forbid'` is removed.
- **Factories are untouched.** They already opt into `params` by declaring it in their signature (`if 'params' in sig.parameters`), so param-less factories do not grow an unused argument.

Skipped tests drop from 54 to 41: the params serialization-roundtrip test previously skipped the 13 param-less workflows and now exercises them.

## Test plan

Full suite including slow and browser tests passes: 4697 passed, 67 skipped, 8 xfailed.

### DIFF
--- a/src/ess/livedata/config/instrument.py
+++ b/src/ess/livedata/config/instrument.py
@@ -25,6 +25,7 @@ from .workflow_spec import (
     DETECTORS,
     REDUCTION,
     AuxSources,
+    NoParams,
     WorkflowGroup,
     WorkflowId,
     WorkflowSpec,
@@ -600,7 +601,7 @@ class Instrument:
         title: str,
         description: str = '',
         source_names: Sequence[str] | None = None,
-        params: type[Any] | None = None,
+        params: type[pydantic.BaseModel] = NoParams,
         aux_sources: AuxSources | None = None,
         outputs: type[Any],
         device_outputs: dict[str, str] | None = None,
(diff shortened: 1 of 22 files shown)


SimonHeybrock (commented):
LGTM

PR #1195, opened 2026-08-04
Keep cells alive through transient render errors

### LINKED ISSUE #1192: Concurrent lazy pandas import inside holoviews can fail the first plot render
During a plot render, holoviews lazily imports pandas (`holoviews/core/util/types.py:137`, `masked_types`, reached via `isfinite` in range computation). If this first pandas import of the process races another thread's import, the render fails with

```
File ".../holoviews/core/util/types.py", line 137, in masked_types
  from pandas.core.arrays.masked import BaseMaskedArray
...
File ".../pandas/core/api.py", line 16, in <module>
  from pandas.core.dtypes.missing import (
ImportError: cannot import name 'isna' from partially initialized module 'pandas.core.dtypes.missing' (most likely due to a circular import)
```

Observed in the dashboard server (fake backend, no Kafka) under the browser test suite, both serially on an idle machine and more often under CPU contention. It can fire at most once per process — whichever render triggers the first pandas import — which makes it look like an unreproducible one-off in CI.

The failed render is caught by `Plotter.plot` and turned into an error placeholder, which then permanently kills the cell via the type wedge described in the follow-up issue — so the visible symptom is a plot that never receives data.

Fix: import pandas eagerly at dashboard startup, before any render thread exists, so holoviews' lazy import always finds a fully initialized module.

### LINKED ISSUE #1193: A transient render error permanently kills a cell: bare error Text wedges the DynamicMap
When `Plotter._build_result` raises, `Plotter.plot` caches a bare `hv.Text("Error: ...")` placeholder (`plots.py`). The cell's DynamicMap, however, holds the plotter's normal container type -- `hv.Overlay` or `hv.Layout` -- and holoviews refuses mixed types:

```
File ".../holoviews/core/ndmapping.py", line 992, in _item_check
  raise AssertionError(
AssertionError: DynamicMap must only contain one type of object, not both Overlay and Text.
```

From that point every frame of the cell raises, so one transient render error (e.g. #1192) permanently kills the cell for the session's lifetime. The placeholder that was supposed to surface an error instead silently stops all updates -- the exact opposite of its intent.

Observed end-to-end in the browser test suite: `test_clicking_out_an_empty_region_creates_a_live_plot` fails with "stopped receiving data updates" whenever #1192 fires, including serially on an idle machine.

The no-data placeholder in `_build_result` already handles this correctly by routing the `hv.Text` through the same Overlay/Layout combine step as real data. The error path predates that and bypasses it.

Fix for the base plotter: wrap the error placeholder exactly like the no-data placeholder, so every frame a plotter emits has the same container type.

Latent same-class cases (not covered by that fix): `table_plotter.py` and `slicer_plotter.py` return a bare `hv.Text` as their no-data placeholder while their data frames are `hv.Table`/image elements, so a no-data-to-data transition inside one DynamicMap would wedge the same way. If their pipes only ever push once data exists, the placeholder is dead code today; either way they should be made consistent when touched.

SimonHeybrock:
Fixes #1192 and fixes #1193, the chain behind the "plot stopped receiving data updates" browser-test flake (which reproduces serially on an idle machine, so it is not a CI-load artifact).

holoviews imports pandas lazily during rendering (`masked_types` reached via `isfinite` in range computation). If that first pandas import of the process races another thread's import, the render fails on a partially initialized module. plots.py now imports pandas eagerly, before any render thread exists (#1192).

When a render fails, `Plotter.compute` cached a bare `hv.Text` error placeholder next to the `Overlay`/`Layout` frames the cell's DynamicMap normally holds; holoviews refuses mixed types, so every later frame raised and the cell was dead for the session's lifetime. The placeholder now goes through the same combine step as data frames, via an overridable `_error_placeholder`. `TablePlotter` overrides it to keep its existing bare-Text convention (its frames are `hv.Table`, the latent case noted in #1193).

## Test plan

- [x] New regression tests: the error placeholder shares the container type of data frames (overlay and layout modes), and a rendered plot shows the error frame and then recovers to data — exercising the `get_plot_frame` path where the wedge lived. All flip on the old code.
- [x] Full unit suite passes (4669 passed).
- [ ] Browser suite flake rate: observe `test_clicking_out_an_empty_region_creates_a_live_plot` over the next runs.

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -11,6 +11,12 @@ from typing import Any, ClassVar
 
 import holoviews as hv
 import numpy as np
+
+# holoviews imports pandas lazily during rendering (e.g. the masked-type check
+# in its isfinite). If that first import races another thread's pandas import,
+# the render fails on a partially initialized module (#1192). Import eagerly,
+# before any render thread exists.
+import pandas  # noqa: F401
 import scipp as sc
 from bokeh.models import TeeHead
 from holoviews.core.util import range_pad
@@ -722,9 +728,7 @@ class Plotter:
             result = self._build_result(data, resolver, **kwargs)
         except Exception as e:
             self._range_targets = {}
-            result = hv.Text(0.5, 0.5, f"Error: {e}").opts(
-                text_align='center', text_baseline='middle', responsive=True
-            )
+            result = self._error_placeholder(f"Error: {e}")
 
         # Time bounds drive the cell titlebar's freshness indicator; they are
         # kept off the plot title to avoid minting an OptionTree entry per tick.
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1196, opened 2026-08-04
Speed up the browser tests: poll instead of sleep, run two xdist workers

SimonHeybrock:
Browser-test-helper changes plus parallelization, no test semantics weakened:

`assert_updating` slept out the full 6 s update window on every call before comparing fingerprints. It now polls the fingerprint until it changes (same 6 s deadline), so a healthy session — the fake backend emits at 1 Hz — passes in about a second. Across the 15 call sites this takes the serial suite from 8:01 to 6:44 locally.

The data-fingerprint sampler recursed into any value with a numeric `.length`. Strings have one, and `'x'[0] === 'x'`, so a single string column in any rendered ColumnDataSource recursed until `RangeError: Maximum call stack size exceeded`. Serial runs never happened to fingerprint while a string column was present; under shifted timing (parallel runs) it fails deterministically. String columns now contribute via the length count only.

The tox browser env now runs with `pytest -n 2`: each test launches its own dashboard on a dynamically allocated port, so the suite parallelizes, and two workers on the 4-core runners should roughly halve the job. Whether parallel load raises the #1185 stall-flake rate can only be measured by running it here; if it does, dropping the `-n 2` is a one-line revert. Locally, combined with the #1195 fixes, `-n 4` passes 16/16 in 3:31.

## Test plan

- [x] Full browser suite passes serially (16/16, 6:44).
- [x] Full browser suite passes with `-n 4` when combined with the #1195 fixes (16/16, 3:31).
- [ ] Watch the browser job here and on the next few main runs for stall flakes ("modal opened only on attempt" warnings, update-window failures).

### DIFF
--- a/tests/helpers/browser.py
+++ b/tests/helpers/browser.py
@@ -46,7 +46,10 @@ DATA_FINGERPRINT_JS = """() => {
     for (let i = 0; i < Math.min(column.length, 64); i++) {
       const v = column[i];
       if (typeof v === 'number' && Number.isFinite(v)) checksum += v;
-      else if (v && typeof v.length === 'number') sample(v);
+      // Strings also have a numeric length, but 'x'[0] === 'x', so recursing
+      // into one never terminates. A string column contributes via `length`.
+      else if (v && typeof v !== 'string' && typeof v.length === 'number')
+        sample(v);
     }
   };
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

Issue #1198, opened 2026-08-05
Plot grid first render blocks the shared IOLoop, superlinearly in cell count


SimonHeybrock:
A plot grid's first render blocks the dashboard's shared Tornado IOLoop, and the block grows faster than the number of cells in the grid. At 24 cells -- a legal grid, the size widget allows 2-6 rows by 2-6 columns -- one session switching to the grid tab makes **every other session on that server unresponsive for 11 seconds**.

This is the residual of #1185 after PR 1197 fixed the colormap cold start that dominated it. It was previously assumed to be a first-paint latency for the rendering user only; measuring from a second, idle session shows it is server-side and shared.

## Measurements

Fake backend, dummy fixture, Detectors grid expanded to N 1x1 cells, idle 24-core machine. Two browser sessions: one renders the grid, the other sits idle on the Workflows tab and probes. The idle session's `fetch` latency is a server-side measurement -- its own main thread is idle, which its `setInterval` lateness (<= 8 ms throughout) confirms.

| cells | models in doc | server block (seen by the *idle* session) | client block (renderer's own main thread) |
|---|---|---|---|
| 3 | 267 | 0.7 s | 1.9 s |
| 6 | 489 | 0.9 s | 4.3 s |
| 12 | 937 | 2.8 s | 13.9 s |
| 24 | ~1900 | 11.1 s | 17.6 s |

The client column is the browser-side cost of instantiating the model tree. It is far larger, but it is a per-user cost and belongs to #1055, whose "switch into plot tab" target reports 1.7-1.9 s of blocked main thread with 4 plots -- consistent with the 3-cell row here. This issue is about the server column.

Server-side accounting matches the observed block. At 12 cells:

```
[passprof] tab_switch 444ms (wall) models 84->325
[passprof] pass full=True handlers=4 models=325 inner=1632.9/1508.0ms freeze_exit=59.1/55.9ms hold_exit=615.7/537.2ms (wall/cpu)
```

444 ms + 2.31 s = 2.75 s against 2.78 s measured from the idle session. `inner` is entirely `PlotGridTabs._poll_for_plot_updates`; `hold_exit` is event dispatch and serialization; `freeze_exit` is bokeh's model-graph recompute.

[...]

PR #1205, opened 2026-08-05
Build per-layer cell toolbars lazily instead of hidden

SimonHeybrock:
The per-layer toolbars in a plot cell start collapsed, but they were built up front and hidden with `visible=False`. A hidden Panel component still creates every one of its Bokeh models (our own rule in `.claude/rules/dashboard-widgets.md` says as much), so every session paid for detail most cells never show — and the poll loop then wrote time readouts into those invisible panes on every frame.

They are now built when the toggle first reveals them. Each layer's time pane is seeded from the layer's current bounds as it is built, so a toolbar revealed between frames does not read blank until the next data flush.

Measured on the dummy fixture (#1198 harness, differencing settled 6- and 12-cell censuses): **98 -> 75 models per cell**, and freeze-exit — bokeh's model-graph recompute, O(models), on every pass — **49 -> 40 ms** on a 12-cell grid. Model count also drives the first-render block and the client's instantiation cost, which is the larger half.

Tests reach the toggle through its committed `lt-tool-layer-details` hook rather than by indexing the layout; the new `tests/helpers/panel_ui.py` does that lookup for any `lt-*` button.

## Test plan

- [x] `pytest tests/dashboard` passes (new coverage: collapsed cell builds no toolbars, revealing builds them once, a rebuild preserves the revealed state)
- [x] Manual: expand a cell's layer details in a running dashboard, confirm the rows appear with their time range already filled in, collapse and expand again

### DIFF
--- a/src/ess/livedata/dashboard/widgets/cell.py
+++ b/src/ess/livedata/dashboard/widgets/cell.py
@@ -202,8 +202,8 @@ class CellWidget:
     """
     Per-session widget for a single plot cell.
 
-    Builds a Panel ``Column`` with a titlebar, a (toggleable) column of
-    per-layer toolbars, and a content area showing the composed plot or a
+    Builds a Panel ``Column`` with a titlebar, a column of per-layer toolbars
+    built on first reveal, and a content area showing the composed plot or a
     status placeholder. Owns the cell's autoscale controller and the panes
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

PR #1206, opened 2026-08-06
Style computed frames instead of the session's DynamicMap

### LINKED ISSUE #1199: Collapse the per-frame .opts() wrappers on session plots
Each layer's session-side plot carries two chained `.opts()` wrappers, and neither is the one-off it reads as. `Opts._dynamicmap_opts` does not set options on a DynamicMap — it wraps the map in a `Dynamic` whose operation calls `obj.opts(...)` again for **every frame**. We stack two:

- `PassthroughPresenter.present` (`plots.py`) applies the plotter's style opts.
- `CellWidget._compose_plot` (`cell.py`) applies the cell's hooks: `result.opts(hooks=hooks)`.

plus overlay collation for multi-layer cells, which is why the in-app chain evaluates 5 DynamicMap callbacks per `Pipe.send`.

Isolated benchmark, one image layer, `Pipe.send` through to the bokeh models:

| variant | ms/frame |
|---|---|
| two wrappers (as shipped) | 12.90 |
| one wrapper | 9.47 |
| passthrough, elements pre-styled | 9.87 |

So the second wrapper costs **27 % of every repaint**, on every layer, of every visible cell, on every frame. Steady state is where a plot grid lives (see #1198), so this is the cheapest available cut in the shared IOLoop's per-frame budget.

Proposal: apply the cell's hooks in the same `.opts()` call as the style opts, i.e. pass them into `PresenterBase.present`, so a single-layer cell — the common case — builds one wrapper instead of two. Multi-layer cells keep the overlay level; hooks that must act on the `OverlayPlot` rather than a subplot stay where they are.

Care needed: `hooks` today reach the composed result (SaveTool filename, hover suspend, autoscale controller), so per-layer application must be checked to be equivalent for single-layer cells, not assumed.

What does *not* help, measured: styling at compute time instead of per frame (9.87 vs 9.47 ms — the wrapper is not the cost, the holoviews update path is) and registering the options in the `Store` against a per-layer group (9.82 ms). Getting past that ~9.5 ms floor is a separate change (see the repaint fast-path issue).

SimonHeybrock:
## Motivation

`DefaultPresenter` applied the plotter's style opts to each session's DynamicMap. That reads as a one-off, but `Opts._dynamicmap_opts` does not set options on a DynamicMap -- it wraps the map in a `Dynamic` operation that re-applies them to **every frame**, minting a custom option id each time. The styling therefore ran per layer, per session, per frame, on the shared IOLoop, which is the resource a plot grid runs out of first (#1198).

This is the mechanism #958 was reaching for and did not get: that PR moved the styling decision out of the per-tick build (correct, and kept here), but attached the result somewhere that re-does the work per frame -- and moved it from the shared compute path onto each session's render path in the process.

## Approach

Apply the opts to the finished frame in `compute()`. The frame is computed once for all sessions and off the render path, so one typed `.opts()` call per frame now serves every session, and the session's DynamicMap goes back to being a bare passthrough.

This is the alternative to the merge-the-two-wrappers proposal in #1199: both end at one per-frame wrapper, but this route leaves the cell's hooks exactly where they are and so needs none of the hook-lifecycle machinery that proposal required (cell-scoped hooks vs layer-scoped DynamicMap). The remaining wrapper is the cell's `.opts(hooks=...)`, which cannot be hoisted -- hooks are session state.

## Measurements

Fake-backend 6-cell render harness, 12 layers at 1 Hz, single session, three alternating runs per side, pooled over the settled full passes: **11.7 / 12.4 / 11.7 ms -> 9.5 / 9.8 / 10.6 ms** of IOLoop time per layer per frame. The saving is per session, so it scales with the number of open sessions.

## Rendering equivalence

A bokeh-level model signature (every model in the rendered document, all non-default properties) over 51 plotter configurations -- both plotters, both combine modes, one and two sources, four aspect types, log and linear colour scales, error bars, and the slicer -- is identical before and after.

The new test in `plot_sizing_invariant_test.py` pins the invariant that styling rides on the frame rather than on the DynamicMap; it fails on every parametrisation without the change.

## Test plan

- [x] `tests/dashboard` (2066 passed, 4 xfailed)
- [x] Manual: open a grid with 1D, 2D, overlay and layout cells and confirm styling, sizing and colorbars are unchanged.

Closes #1199.

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -339,15 +339,16 @@ class DefaultPresenter(PresenterBase):
     def present(self, pipe: hv.streams.Pipe) -> hv.DynamicMap:
         """Create a DynamicMap that passes through pre-computed elements.
 
-        Static styling is applied once on the DynamicMap (see
-        :meth:`Plotter.style_opts`) rather than per element on every tick.
+        The elements arrive already styled from :meth:`Plotter.compute`. Styling
+        must not be applied here: ``DynamicMap.opts`` is not a one-off, it wraps
+        the map in a ``Dynamic`` operation that re-applies the options to every
+        frame, in every session, on the shared IOLoop.
         """
 
         def passthrough(data):
             return data
 
-        dmap = hv.DynamicMap(passthrough, streams=[pipe], cache_size=1)
-        return dmap.opts(*self._plotter.style_opts())
+        return hv.DynamicMap(passthrough, streams=[pipe], cache_size=1)
 
 
 class StaticPresenter(PresenterBase):
@@ -735,7 +736,7 @@ class Plotter:
         # Computed outside the try so a render failure still stamps the bounds
         # of the data that was received.
         self._time_bounds = _compute_time_bounds(data)
-        self._set_cached_state(result)
+        self._set_cached_state(result.opts(*self.style_opts()))
 
     def _build_result(
         self,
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
LGTM

PR #1207, opened 2026-08-06
Recompute the model graph only on passes that changed it

### LINKED ISSUE #1202: Skip the freeze-exit model-graph recompute on passes that changed nothing
`SessionUpdater._tick` wraps every pass in `doc.models.freeze()`, whose exit recomputes the document's model graph at O(models) — unconditionally, including on passes where no handler mutated anything.

The unconditional full pass every 5 s is the clearest case: at 2139 models (a 24-cell grid) it costs 86-232 ms of freeze-exit with `inner` at 0.4-1.5 ms, i.e. bokeh walking the whole model graph for a pass that changed nothing. That is 2-4 % duty per session, and it grows with document size, so every session sitting on a big grid pays it forever.

Fix: skip the freeze (or the recompute) when the pass produced no document mutation. The signal has to be something the pass can observe cheaply — e.g. whether any document change event was emitted while the batch was held — rather than trusting handlers to report it.

Small in absolute terms next to the repaint cost (see the fast-path issue), but it is pure waste and it is the one item on the #1198 list that is entirely inside our own code.

SimonHeybrock:
Closes #1202.

`doc.models.freeze()` recomputes the document's model graph when it exits, whether or not anything changed, at O(models). Every `SessionUpdater` pass and every tab switch paid that on the IOLoop all sessions share.

The freeze counter is now held up directly for the pass, and the recompute at the end runs only if the pass could have changed which models are reachable from the roots: Bokeh called `DocumentModelManager.invalidate` — its own signal that a property change touched model references, interceptable here because while frozen it does nothing — or the roots themselves changed, which Bokeh signals by freezing around the change rather than by invalidating. Batching is otherwise unchanged: the counter is up for the whole pass, so inner freeze/unfreeze cycles (Panel's `freeze_doc`, HoloViews' `hold_render`) stay no-ops and a mutating pass still recomputes exactly once. Unlike Bokeh's `freeze`, the pop also runs on the exception path, so a raising handler can no longer leave the document frozen for good.

The idiom was duplicated in `_BatchedTabs`, so it moved into one `batched_update()` both call.

## Measured

Fake backend, dummy fixture, synthetic N-cell grid at 1 Hz, one session; this branch against `main` in the same harness, timing each pass and the recompute it skipped.

| grid | models | full pass (5 s cadence) | steady-state data pass |
|---|---|---|---|
| 6 cells | 473 | 33 ms → 0.2 ms | −17 ms of ~150 ms |
| 24 cells | 1753 | 128–213 ms → 0.7 ms | −141 ms of ~1.1 s |

The steady-state data passes benefit too, which #1202 did not anticipate: plot updates reach the browser as `ColumnDataSource` patches and streams, which Bokeh deliberately does not invalidate on, so a pass that only pushes new data into existing glyphs recomputes nothing at all.

That the *whole* steady state is such passes was not obvious, so it was measured rather than assumed: instrumented to record the outcome of every pass, a session sitting on the dummy fixture's Detectors grid for 90 s ran **106 passes and skipped the recompute in all 106**, including the 5 s full passes. The other per-tick cell updates do not put that back: the freshness pill and the text readouts write strings into panes that already exist, and a string cannot hold a model reference. So the saving is a recompute per pass at the tick cadence (~1.2/s there), not one per 5 s — about 3–6 % of a core per session at 6 cells, 15–30 % at 24 cells, growing with document size.

## Risk

Correctness now rests on Bokeh's invalidation being complete. `invalidate` has a single call site, `Model.trigger`, which skips it only for hinted events — `ColumnsStreamed` and `ColumnsPatched`, whose payload is column data and cannot carry model references — and which otherwise counts references by recursing through lists, tuples, dicts, dataclasses and non-Model `HasProps`. Reachability therefore cannot change without either that signal or a roots change, by induction over passes: a subtree becomes reachable only through a property set on an already-reachable model, and reachable implies attached implies `trigger` invalidates.

A hole in that would break any unfrozen Bokeh app too, since unfrozen Bokeh recomputes on exactly the same signal — but the exposure differs in degree, and this is where the change actually costs something. Unfrozen, an orphaned model heals at the next invalidating change anywhere in the document, which in a live app is near-constant. Here every mutation is inside a batch, and a steady state made entirely of data passes invalidates nothing at all, so an orphan can persist for as long as the session only pushes data. Until now the unconditional recompute papered over such a hole once per pass, and it no longer does. The failure mode would be a model that is added but never attached to the document.

The residual risk is therefore less a hole in Bokeh today than a future Bokeh changing the contract. The unit tests are written to pin the mechanism rather than the outcome, so that shift fails the suite instead of the dashboard.

## Test plan

- New unit tests drive a real `Document`: added/removed models attach and detach, many mutations recompute once, a pass that mutates nothing recomputes zero times, a data-only pass (stream + patch) recomputes zero times and leaves the graph consistent, an inner Bokeh `freeze()` does not recompute mid-pass, nested batches including a root added by the inner one, and a raising pass leaves the document usable.
- Full suite green; all 16 `-m browser` tests pass (they render plots, drive modals, and cover multi-session).
- The browser suite was also run with a temporary check that, on every skipped recompute, recomputed the reachable set and compared it against `doc.models`: 212 skips, no mismatch. Probing 14 mutation styles directly against Bokeh (in-place `children` mutation, glyphs and tabs added after attach, `data_source` swap, `CustomJS.args`, `js_on_change`, `tags`, `stylesheets`, deep detached subtrees) found no invalidation hole either.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -36,7 +36,7 @@ def refresh(self):
 
 **Key principles**:
 - Widgets detect changes via version counters on shared state, not via push callbacks
-- All widget updates run inside `SessionUpdater._batched_update()` for efficient recomputation
+- All widget updates run inside `batched_update()` for efficient recomputation
 - Widget event handlers should only call methods on shared components, never rebuild directly
 
 ## Icons
--- a/src/ess/livedata/dashboard/batched_update.py
+++ b/src/ess/livedata/dashboard/batched_update.py
@@ -33,6 +33,9 @@ def batched_update() -> Iterator[None]:
     Nesting is a no-op: an inner batch is already covered by the outer one.
     """
     doc = pn.state.curdoc
+    # The freeze is the inner context so that it recomputes before the hold
+    # dispatches: a model added during the pass must be attached to the document
+    # by the time the queued events are serialized.
     with pn.io.hold(), _frozen_models(doc):
         yield
 
--- a/src/ess/livedata/dashboard/batched_update.py
+++ b/src/ess/livedata/dashboard/batched_update.py
@@ -1,0 +1,89 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+"""Batching of the Bokeh document mutations a single IOLoop pass makes."""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from contextlib import contextmanager
+from typing import TYPE_CHECKING
+
+import panel as pn
+
+if TYPE_CHECKING:
+    from bokeh.document import Document
+
+
+@contextmanager
+def batched_update() -> Iterator[None]:
+    """Batch the document events and model-graph recomputes of one pass.
+
+    ``pn.io.hold()`` batches document change events so they are dispatched to
+    the browser in one WebSocket flush, avoiding staggered rendering.
+
+    The freeze batches Bokeh's model-graph recomputation. Without it, each
+    operation that mutates the model graph (pipe.send, layout child changes)
+    triggers a full BFS traversal of every model in the document via
+    ``_pop_freeze`` -> ``recompute`` -> ``collect_models``, at O(models) cost.
+    Holding the freeze counter above zero for the whole pass makes the inner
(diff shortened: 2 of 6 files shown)


SimonHeybrock (commented):
LGTM

PR #1208, opened 2026-08-06
tests: Wait once for a modal instead of re-clicking its trigger

SimonHeybrock:
## Overview

Removes the `open_modal` re-click added as an explicit harness workaround for #1174, now that #1197 removed the stall it was covering. `open_modal` goes back to one click and one wait.

## Why now

The multi-second block a click could queue behind was the cold start of the colormap machinery, which #1197 moved to process startup. Across the browser jobs before that landed, 13 of 21 emitted a modal-recovery warning and 9 of 21 failed a modal test; across the 10 jobs after it, none did (Fisher one-sided: p=0.0017 and p=0.021). `browser-test-speedups` merged near the same boundary, but four baseline runs already contained it and still flaked, so it is not the explanation.

Headroom is now about 4x: driving `main` with the retry bypassed, 63 tests across 7 rounds passed with 30 modal opens, worst 2.3 s under full CPU saturation, against a 10 s wait.

The retry also could not answer the question it was built for. Every recovery it logged was "attempt 2 via 1 re-click", but the re-click fires before the second wait, so the warning cannot separate the re-click from the extra waiting. And absorbing the symptom into a warnings-summary line costs the loud failure that the remaining event-loop work (#1185, #1198) wants as its detector.

`Dashboard.click` keeps its retry-on-detach: a row rebuild genuinely detaches the element and Playwright raises, which is a different mechanism.

Caveat worth stating: 10 clean jobs bound the residual per-run flake rate at about 26% (95% upper), so this is "not observed", not "proven impossible". A recurrence now shows up as a red browser job.

## Test plan

- [x] `tox -e browser` equivalent (`pytest -m browser -n 2`) after the strip: 9 passed.
- [x] Seven further rounds on `main` with the retry path bypassed (3 unloaded at `-n 12`, 4 at `-n 2` with all 24 cores saturated): 63 passed, no modal open above 2.3 s.

### DIFF
--- a/.claude/rules/dashboard-widgets.md
+++ b/.claude/rules/dashboard-widgets.md
@@ -129,8 +129,7 @@ assets are served off the same IOLoop as everything else, so their `tornado.acce
 timings measure how long that loop was blocked: a `GET /static/...` taking seconds
 means the click's round-trip was queued behind the session's periodic pass, not lost
 (#1185). Sub-millisecond serves with no post-click activity at all point at a dropped
-click instead. `Dashboard.open_modal` re-clicks rather than waiting once, because only
-one of those two recovers.
+click instead.
 
 One console line is a red herring: `pageerror: Cannot read properties of undefined
 (reading 'parent_style')` fires on **every** run, passing or failing. It is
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1209, opened 2026-08-06
Unregister HoloViews' no-op nodata compositor

SimonHeybrock:
HoloViews applies its `apply_nodata` compositor to every `Image`/`Raster`/`QuadMesh`/`ImageStack` frame. It implements the `nodata` plot option: an integer sentinel value is rewritten to NaN so it draws transparent, which matters for geo rasters and datashader aggregations. It cannot fire here — no plotter offers the option, and our 2D path converts to float64 before plotting, which the operation passes through untouched.

The machinery around the operation runs regardless, and it runs *per frame per session*: the rule is applied to the session's DynamicMap, so every pushed frame is wrapped in overlays, pattern-matched, passed through the no-op and cloned back in, behind an extra DynamicMap callback hop. That is pure overhead on the per-layer repaint path that #1198 identified as the dashboard's capacity ceiling.

Measured with a 6-cell grid (12 layers) at 1 Hz, seven alternating runs per side, pooled over the settled full passes: median 10.8 -> 8.8 ms per layer per frame. One run of each side overlapped, so the honest read is 1.5-2 ms. The saving is per session, so it scales with session count — a single-session harness understates it. Both sides of that A/B predate #1206, which cut a different phase of the same pass, so the absolute per-layer figures are lower on current `main`; the delta measured here is unaffected.

`Compositor.definitions` is a global registry matched by element type, so this is all-or-nothing: `.opts(nodata=...)` is now inert in the dashboard. That is the right trade, and the reasoning is recorded next to both the removal and `Plotter._prepare_2d_image_data`, because we are likely to want display masking later (e.g. drawing masked detector regions differently). Masking belongs in `Plotter.compute`, in scipp: a frame masked there is computed once and shared by every session, rather than re-derived per session on every render — the same argument as #1206. Only the `apply_nodata` definitions are removed; the compositors that are load-bearing for element types we never create (`Distribution`, `Bivariate`, `Sankey`, `HexTiles`, `ImageStack` -> RGB) stay registered.

## Test plan

- Unit tests cover that the compositor is unregistered and that 2D data still reaches Bokeh unchanged as float64, on both the `Image` and `QuadMesh` paths.
- [x] 2D detector plots render as before in a live dashboard (colormap, color limits, autoscale).

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -633,6 +633,12 @@ class Plotter:
         calculation should properly adjust the color limits. Since zeros can never
         be included we want to adjust to the lowest positive value.
 
+        Masking values for display belongs here rather than in HoloViews. Its
+        `nodata` plot option is inert -- ``reduction`` unregisters the compositor
+        implementing it, and that registry is global, so it cannot be turned back
+        on for a single plot. A frame masked here is computed once and shared by
+        every session, instead of being re-derived per session on every render.
+
         Parameters
         ----------
         data:
--- a/src/ess/livedata/dashboard/reduction.py
+++ b/src/ess/livedata/dashboard/reduction.py
@@ -9,8 +9,9 @@ from urllib.request import urlopen
 
 import holoviews as hv
 import panel as pn
+from holoviews.core.options import Compositor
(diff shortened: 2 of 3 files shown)


SimonHeybrock (commented):
LGTM

PR #1212, opened 2026-08-06
Fix narrow layer-picker menu on multi-layer cell gear

SimonHeybrock:
## Summary
- The multi-layer cell gear's layer-picker dropdown wraps every entry across multiple lines instead of rendering on one line.
- Root cause: the `min-width: 200px` rule targeted `.bk-menu`, but Bokeh's `Dropdown` popup carries the class `bk-Menu`. The case mismatch silently dropped the rule, so the popup fell back to its intrinsic (very narrow) width.
- Fixed both occurrences (the layer-picker gear and the overlay `+` menu, which shares the same stylesheet pattern) and added a bounding-box width assertion to the existing browser test so a regression fails loudly.

## Test plan
- [x] `test_multi_layer_cell_gear_picks_the_layer_to_configure` (browser) passes with the fix and fails against the old selector (verified locally).
- [x] Manually open a multi-layer cell's gear menu and confirm entries render on one line.

### DIFF
--- a/src/ess/livedata/dashboard/widgets/plot_widgets.py
+++ b/src/ess/livedata/dashboard/widgets/plot_widgets.py
@@ -99,7 +99,7 @@ def create_overlay_add_button(
     stylesheets = create_tool_button_stylesheet(button_color, hover_color)
     stylesheets.append(
         """
-        .bk-menu {
+        .bk-Menu {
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
Works

PR #1214, opened 2026-08-06
Skip HoloViews' range computation where the dashboard owns the axes

SimonHeybrock:
Part of #1198. HoloViews recomputes each layer's axis bounds from the data on every frame, in every session — deriving extents, scanning the dimensions and writing the Bokeh ranges. For most of our layers that result is then discarded: `CellAutoscaleController` writes both numeric axes itself on every render, and ROI readback overlays are drawn in the image's coordinate space and never determine bounds at all.

Those plotters now carry HoloViews' documented `apply_ranges=False`, which skips the computation. It rides in the opts `compute()` already applies to the finished frame, so it costs no extra option application (the lesson of #1206), and being a static plot option it triggers no param update after the first render.

## Measured

6-cell grid (12 layers) at 1 Hz, one session, fake backend, four alternating runs per side, pooled over the settled data passes:

| | ms per layer per frame | range over runs |
|---|---|---|
| `main` | 8.33 | 7.58 – 8.96 |
| this branch | **6.69** | 6.55 – 6.84 |

−20 %, with no overlap between the two sides. Isolated, the saving is 1.30 ms per 1D layer and 0.44 ms per 2D layer; the grid figure is higher than that mix predicts because the ROI annotation layers benefit too.

### It scales per element, not per layer

HoloViews' range work runs per element — `_compute_group_range` per element, `get_extents` per subplot, `dimension_range` per dimension — so a layer holding several elements saves several times over. That is the common case, not a corner one: a `LinePlotter` with a multi-source selection overlays every source into a single layer, and `Overlay1DPlotter` draws one curve per ROI spectrum.

Curves in one layer, one `pipe.send`, median of three runs per point:

| curves | baseline | with opt-out | saved | % |
|---|---|---|---|---|
| 1 | 5.49 | 4.80 | 0.69 | 12.6 |
| 2 | 7.79 | 6.34 | 1.45 | 18.6 |
| 4 | 11.92 | 9.40 | 2.52 | 21.1 |
| 8 | 21.39 | 16.04 | 5.35 | 25.0 |
| 16 | 41.08 | 30.88 | 10.20 | 24.8 |

Flat at ~0.65 ms per curve across the range, so a layer with N curves saves about what N single-curve layers would. The percentage improves with element count because the fixed per-layer overhead — the send, the DynamicMap hop, the freeze exit — amortises while the range saving keeps growing.

### And with the layers sharing a cell

Elements per layer is only half of it. A cell collates its layers into a single `OverlayPlot` (`_compose_plot`), so a send on *one* layer re-runs the overlay's range machinery across every subplot in the cell — 4 `get_extents` calls per layer in the cell, per send. `_update_glyphs` stays at 1, so only the sending layer's data is pushed; it is specifically the range work that fans out.

Sending on one layer of an N-layer cell, medians of three processes:

| layers in the cell | baseline | with opt-out | saved | `get_extents` per send |
|---|---|---|---|---|
| 1 | 4.58 | 4.15 | 0.43 | 1 |
| 2 | 6.90 | 5.59 | 1.31 | 8 |
| 3 | 8.42 | 6.83 | 1.59 | 12 |
| 4 | 10.13 | 8.43 | 1.70 | 16 |

This is what reconciles the isolated numbers with the grid. The fixture's six-layer cycle is one 1-layer cell, one 3-layer cell (image plus two ROI overlays) and one 2-layer cell, which predicts `(1×0.43 + 3×1.59 + 2×1.31) / 6` = 1.30 ms per layer against the 1.64 measured; the remainder is plausibly the real data these cells carry (log color scale, hover, 64-bin histograms) versus the bench's synthetic curves.

The general shape: a layer's repaint cost is not local to that layer, so the saving scales with elements per layer *and* with the number of layers sharing the cell.

Taking ranges apart also showed that the largest single component was option lookups (0.87 ms over 34 calls per frame for a 2-curve overlay) rather than range arithmetic. `apply_ranges=False` removes about half of the total; the rest is addressable upstream by memoizing lookups for the duration of one render, tracked in #1215.

## Why the opt-out is per plotter

It defaults to off, so a plotter whose axes nobody writes — `BarsPlotter`, `TablePlotter`, the slicer (which autoscales only the color axis) — keeps HoloViews' computation. That direction fails safe: a new plotter misses the saving instead of silently rendering on Bokeh's unit range. A global switch with carve-outs would fail the other way, and for a bar chart of 0D scalars a default 0..1 range is wrong rather than a don't-care.

The condition also requires `is_overlayable`, which is not incidental: a layout-mode plotter draws one figure per source, while `RangeHandles` writes through a single figure's `x_range`/`y_range`. The cell already skips its hooks for such layers for the same reason. Without that guard, detector images in layout mode collapse onto 0..1 — a silent, viewer-visible failure that no existing test caught, since the elements and figures all still exist. The new test renders through the real `compute -> hook -> bokeh` path and asserts the axis ranges actually span the data, and it fails on all four layout parametrisations if the guard is removed.

## Test plan

- [x] `pytest tests/dashboard` — 2090 passed, 4 xfailed
- [x] New range-ownership tests fail on every layout parametrisation when the `is_overlayable` guard is removed
- [x] Detectors and Diagnostics grids render identically to `main` (checked here against the dummy fixture with the fake backend: axes, colorbars, datetime axis, table and slicer all unchanged — worth a second look on real data)
- [x] Pan/zoom with an autoscale toggle off still preserves the manual range

### DIFF
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -940,8 +940,12 @@ class Plotter:
 
         False lets :meth:`_frame_opts` hand HoloViews ``apply_ranges=False``,
         which skips computing bounds, deriving extents and writing the Bokeh
-        ranges on every frame -- a third of a 1D layer's repaint. It is safe in
-        two cases, and this property is the single place that decides:
+        ranges on every frame -- a third of a 1D layer's repaint. That work is
+        per element, so it is not a fixed cost per layer: it grows both with the
+        elements a layer draws (a multi-source selection overlays one per
+        source) and with the number of layers sharing the cell's figure, whose
+        range updates are fused (see ``widgets/cell.py``). It is safe in two
+        cases, and this property is the single place that decides:
 
         - the cell's :class:`~.cell_autoscale.CellAutoscaleController` writes
           both numeric axes itself on every render, i.e. ``AUTOSCALE_AXES``
--- a/src/ess/livedata/dashboard/plots.py
+++ b/src/ess/livedata/dashboard/plots.py
@@ -463,6 +463,12 @@ def format_time_info(bounds: TimeBounds) -> str:
     return f'{format_time_ns_local(bounds.min_end)} (Lag: {lag_s:.1f}s)'
 
 
+def _is_element_opt(opt: hv.Options) -> bool:
+    """Whether ``opt`` targets an Element rather than a container."""
+    element = getattr(hv, opt.key.split('.')[0], None)
+    return isinstance(element, type) and issubclass(element, hv.Element)
+
+
 class Plotter:
     """
     Base class for plots that support autoscaling.
@@ -473,6 +479,14 @@ class Plotter:
 
     AUTOSCALE_AXES: ClassVar[frozenset[Axis]] = frozenset()
     """Per-axis autoscale capability. Override per subclass."""
+
+    IS_ANNOTATION: ClassVar[bool] = False
+    """Whether this layer is drawn in another layer's coordinate space.
+
+    ROI overlays are annotations: they are composed on top of the layer that
+    determines the figure's bounds, so they never need bounds of their own. See
+    :attr:`applies_ranges`.
+    """
 
     _autoscale_axes_override: frozenset[Axis] | None = None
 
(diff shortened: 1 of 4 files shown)


SimonHeybrock (commented):
LGTM

Issue #1219, opened 2026-08-07
Empty-cell click can raise ValueError: PlotGrid occupancy comes from inserted widgets, not topology


SimonHeybrock:
`PlotGrid` decides which cells are free from `_occupied_cells` (`dashboard/widgets/plot_grid.py:224`), a dict populated only by `insert_widget_at` (line 433). Topology is the actual authority on what is occupied. When the two disagree, the grid renders clickable empty cells over positions that already hold a cell, and completing the wizard there raises out of the callback.

## The path

`_is_region_available` / `_is_cell_occupied` (`plot_grid.py:319-337`) consult `_occupied_cells` to decide whether a drag-selected region can take a plot. A position occupied in topology but with no widget inserted in this session reads as available, so the selection is accepted and the config modal opens. On success, `_on_plot_requested.on_success` (`plot_grid_tabs.py:501`) calls `PlotOrchestrator.add_cell`, which raises `ValueError` on overlap (`plot_orchestrator.py:704`) — and the handler catches only `KeyError`. The exception escapes instead of taking the `show_error` path its sibling `_on_reconfigure_layer` uses for exactly this exception type (`plot_grid_tabs.py:531`).

## When the two disagree

Today: a cross-session race. Another session adds a cell; my session opens the wizard on that region before my poll pass reconciles and inserts the widget. Narrow, but the race cannot be designed away — two users can be in the wizard on the same region simultaneously, so `add_cell` can always lose.

#1216 widens it. Gating hidden grids' cell rebuilds means a grid can be fully populated in topology while this session has inserted nothing into it, making the disagreement a reproducible state rather than a race.

## Fix

Two independent parts, both worth doing:

- Catch `ValueError` in `on_success` and route it to `show_error`, matching `_on_reconfigure_layer`. Required regardless — the simultaneous-wizard race has no other defence.
- Drive occupancy from topology rather than from inserted widgets, so the empty cells never appear over occupied positions in the first place. This is what stops the user being offered a region that cannot be filled; the check above only makes the refusal legible.

PR #1220, opened 2026-08-07
Gate hidden-grid cell rebuilds on surviving the reveal

### LINKED ISSUE #1216: Hidden grids rebuild cell widgets that the next reveal discards
Every session's poll pass rebuilds the cell widgets of *every* enabled grid, visible or not. `dynamic=True` keeps the Bokeh models out, so this costs no models — but the Python half is real, and in the single-viewer case the widget it produces cannot survive the reveal it is preparing for: the switch rebuilds it from scratch. The wasted work recurs on every layer version bump for as long as the grid stays hidden.

## What happens today

`PlotGridTabs._poll_for_plot_updates` (`dashboard/widgets/plot_grid_tabs.py:715`) loops over all enabled grids. `is_active` — this session's visible tab — gates layer activation and the data flush; the cell rebuild at lines 803/827/831 is ungated. So a session parked on Workflows builds and rebuilds the widgets of every plot grid in the shared topology, including grids another session created and this one never opens.

Measured on the dummy fixture expanded to 12 cells (15 cells total across the two fixture grids), page sitting on Workflows the whole time:

- page load: 15 `CellWidget`s built, 272 ms
- on each subsequent layer version bump: all 15 rebuilt, 89 ms

## Why it is like that

Deliberate, and documented at `plot_grid_tabs.py:735`: a hidden grid does "no display work at all between switches … this pass only keeps their structure reconciled". Visibility was designed to gate *data flow*, not *widget existence*, so the reconcile stays one uniform pass and a tab switch has no structural work to do. Widget construction was treated as bookkeeping; it is 15–20 ms per placeholder cell and more when the layers are live.

## The hidden build cannot survive its own reveal

What makes a layer computable is a viewer token, and a session holds one only while the grid is its visible tab. Leaving the tab releases it (`activate_layer(..., False)`); returning re-acquires it. `set_active` returns `not was_active` (`dashboard/plot_data_service.py:449`), so for a session that is the grid's only viewer every reveal is a 0→1 transition: `_refresh_layer` runs, the layer version increments, and the same poll pass rebuilds the cell. The widget built while hidden had no computed plot behind it — frame flushes skip layers without viewers — so it was a placeholder, and the reveal discards it. On a 12-cell grid this shows up as the switch materializing ~99 placeholder models (0.26 s) that the next pass throws away.

This is not a first-visit effect. The widget itself survives being hidden (`self._cells` is swept only when topology removes the cell), but the token does not, so a reveal is structurally identical the first time and the tenth. The waste is proportional to version bumps while hidden — every job restart, every plotter swap, every activation refresh caused by another session opening a grid — not to the grid count and not to how much of the instrument a session has explored. It does not taper off as a session ages.

## Where the current behaviour is right

Viewer tokens are shared across sessions (`PlotDataService.set_active`, `plot_data_service.py:410` — a set per layer). If another session already views the grid, `set_active` returns False for the second acquirer: no version bump, no rebuild, and the widget built while hidden is a real plot that survives the switch. There the eager build genuinely pre-warms and moves the HoloViews composition off the tab-switch path.

Any fix has to keep that case working; it is the multi-user case this dashboard exists for.

## Proposed change

Both cases fall out of one predicate:

> Build a hidden grid's cells iff the build will survive its reveal — iff another session already holds a viewer token on the layers.

That is exactly `PlotDataService.has_viewers` (`plot_data_service.py:451`), and it is the same condition that decides whether my reveal is a 0→1 transition. Gate `cells_to_rebuild` on it for non-active grids and leave the rest of the pass — session layers, activation, orphan sweep, `sever` — untouched.

Under the gate, a hidden-grid build is never discarded: grids nobody is watching cost nothing, grids someone is watching keep their pre-warm, and the set adapts as sessions come and go. Nothing here depends on session history, so there is no cadence or debounce to tune.

An earlier revision of this issue proposed gating on "has this session ever displayed this grid" instead. That is strictly weaker — the visited set only grows, so it stops paying once a session has opened everything, and it does not separate a build that will survive from one that will not. Worth keeping only as a fallback if the viewer predicate proves unusable in the poll pass.

## Traps for whoever picks this up

- **Stale empty cells.** `PlotGrid` derives occupancy from `_occupied_cells`, populated by `insert_widget_at` (`plot_grid.py:433`). A grid whose cells were never built shows clickable empty cells over positions topology says are occupied. Clicking one and completing the wizard calls `add_cell`, which raises `ValueError` on overlap — and `_on_plot_requested`'s `on_success` (`plot_grid_tabs.py:501`) catches only `KeyError`, so it surfaces as an uncaught exception. The hole already exists for the cross-session race (another session adds a cell before my grid reconciles); this change widens it to a reproducible state for any grid nobody is viewing. Fix the handler, or drive occupancy from topology rather than from inserted widgets.
- **Reveal ordering.** The tab switch renders the grid, and the pass requested by `_on_active_tab_changed` builds and inserts. Between them the user sees empty cells for one pass. Either build synchronously ahead of materialization or confirm the flash is acceptable.
- **`pn.state.execute` around the insert (`plot_grid_tabs.py:899`) never defers in a session context** — `state._unblocked(doc)` is true under the document lock, so it runs inline (confirmed by phase timing: the insert is inside the handler's wall time). Its comment and the `cid in self._cells` guard describe a deferral that does not happen. Whoever touches this path should decide whether the race it names is real and unprotected, or whether both can go.
- Expect no change to the reveal itself. This removes work done *before* a reveal, not the ~50 ms per layer of Bokeh model construction the reveal pays (see #1203).

## Verifying it

Use the two-session observer setup from #1198 (one session driving, one idle and probing, so the shared-IOLoop cost is isolated from the renderer's own main thread), with server-side phase timing around `CellWidget.__init__` and `PlotGrid.insert_widget_at`. What should change:

- an idle session on Workflows, with no other session viewing those grids: no `CellWidget` construction at all, at page load or on any version bump
- reveal of a 12-cell grid nobody else is viewing: no placeholder models materialized (~99 today), cells built once instead of twice — and this holds on every reveal, not only the first
- unchanged: the multi-viewer case above, and the total block a reveal imposes

Related: #1203 (measures the reveal block and concludes there is no root cause behind it), #1198 (tracker).

SimonHeybrock:
Fixes #1216. Partially addresses #1219 (the `ValueError` handler; topology-driven occupancy is left for a follow-up).

## What

Every session's poll pass rebuilt the cell widgets of every enabled grid in the shared topology, visible or not — at page load and again on every layer version bump (job restarts, plotter swaps, errors). For a grid with no viewer in any session such a build is a placeholder: frame flushes skip layers without viewer tokens, so the reveal's 0→1 activation bumps the layer version and rebuilds the cell anyway, discarding the hidden build. The cost is per session and serialized on the shared IOLoop, so N parked sessions pay it N times back-to-back on every bump.

The poll pass now skips a hidden cell's rebuild unless another session holds a viewer token on one of its layers — the one case where the build is a real plot that survives the reveal and pre-warms the tab switch. Skipped rebuilds leave the session's signature/version records stale, so the reveal pass builds the cell once; N bumps while hidden coalesce into that single build.

Consequences handled along the way:

- A revealed grid's widgets can now lag topology for one pass, so completing the plot wizard on a stale empty cell can hit an occupied position. `add_cell`'s `ValueError` was uncaught in that handler (already reachable via the cross-session wizard race, #1219); it now shows an error notification.
- The `pn.state.execute` wrapper around cell insertion described a deferral that never happens (the pass holds the document lock, so it ran inline); the insert is now direct and the dead removed-cell guard is gone.

## Test plan

Automated tests cover the gate (hidden cell not built, reveal builds it, viewer token preserves the pre-warm build, bumps while hidden defer to the reveal) and the occupied-position error path.

Manual checks:

- [ ] Reveal of a grid whose cells were deferred shows no objectionable empty-cell flash before the widgets appear
- [ ] Two-browser session: with session A viewing a grid, session B parked elsewhere still gets pre-warmed cells and a fast tab switch to that grid
- [ ] Re-measure the idle-session poll cost from #1216 (no `CellWidget` construction while parked on Workflows)

### DIFF
--- a/src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
+++ b/src/ess/livedata/dashboard/widgets/plot_grid_tabs.py
@@ -481,6 +481,25 @@ class PlotGridTabs:
                     # Grid was re-enabled or is new — create fresh tab
                     self._add_grid_tab(grid_id, grid_config)
 
+    def _add_plot(
+        self, grid_id: GridId, geometry: CellGeometry, plot_config: PlotConfig
+    ) -> None:
+        """Commit a configured plot to the orchestrator, surfacing races."""
+        try:
+            cell_id = self._orchestrator.add_cell(grid_id, geometry)
+            self._orchestrator.add_layer(cell_id, plot_config)
+        except KeyError:
+            # The grid vanished while the modal was open (removed or
+            # replaced by another session).
+            show_error('Cannot add plot: the grid was removed.')
+        except ValueError:
+            # The position is occupied in topology but rendered as an empty
+            # cell: another session placed a plot while the modal was open,
+            # or this grid's widgets were not built yet (hidden grids defer
+            # cell builds, so occupancy lags topology until the reveal pass
+            # inserts the widgets).
+            show_error('Cannot add plot: that position is already occupied.')
+
     def _on_plot_requested(self, grid_id: GridId, geometry: CellGeometry) -> None:
         """
         Handle plot request from PlotGrid.
@@ -495,18 +514,9 @@ class PlotGridTabs:
         geometry
             Cell geometry of the selected region.
         """
-
-        def on_success(plot_config: PlotConfig) -> None:
-            """Handle successful plot configuration."""
-            try:
-                cell_id = self._orchestrator.add_cell(grid_id, geometry)
-                self._orchestrator.add_layer(cell_id, plot_config)
-            except KeyError:
-                # The grid vanished while the modal was open (removed or
-                # replaced by another session).
-                show_error('Cannot add plot: the grid was removed.')
-
-        self._show_config_modal(on_success=on_success)
+        self._show_config_modal(
+            on_success=lambda cfg: self._add_plot(grid_id, geometry, cfg)
+        )
 
     def _on_reconfigure_layer(self, layer_id: LayerId) -> None:
         """
@@ -735,10 +745,15 @@ class PlotGridTabs:
         A hidden grid therefore does no display work at all between switches --
         not even on the periodic full pass. Its layers are deactivated
         (``activate_layer(..., False)``), so unless another session is viewing
-        them they do not even compute, though their buffers keep filling; this
-        pass only keeps their structure reconciled. What the user sees on
-        returning to the tab is the latest state, computed on the switch, not a
-        five-second-old rendering.
+        them they do not even compute, though their buffers keep filling. Its
+        cell widgets are rebuilt only while another session holds a viewer
+        token on them: then the plotters have computed state and the build is
+        a real plot pre-warming this session's tab switch. With no viewers the
+        build would be a placeholder that the reveal's 0->1 activation bumps
+        and discards, so it is skipped; the stale signature/version records
+        make the reveal pass build the cell once (#1216). What the user sees
+        on returning to the tab is the latest state, computed on the switch,
+        not a five-second-old rendering.
         """
         # Snapshot before reading any layer state, and record it only once the
         # pass completes (below). The ingestion thread bumps this counter while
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
LGTM. UX is also better since we do not see the empty cells for a second when first switching to a tab -- instead a blank page shows.

